Issue #51, opened 2022-04-15
Ternary operators should not be nested.RSPEC-3358


yeikel:
> Just because you can do something, doesn’t mean you should, and that’s the case with nested ternary operations. Nesting ternary operators results in the kind of code that may seem clear as day when you write it, but six months later will leave maintainers (or worse - future you) scratching their heads and cursing.
> 
> Instead, err on the side of clarity, and use another line to express the nested operation as a separate statement.

Noncompliant Code Example

```java

public String getReadableStatus(Job j) {
  return j.isRunning() ? "Running" : j.hasErrors() ? "Failed" : "Succeeded";  // Noncompliant
}

```

Compliant Solution

```java

public String getReadableStatus(Job j) {
  if (j.isRunning()) {
    return "Running";
  }
  return j.hasErrors() ? "Failed" : "Succeeded";
}

```

https://rules.sonarsource.com/java/RSPEC-3358

pstreef:
I would like to pick this up. Since the issue is a year old I guess there is not much of a hurry and I can try out creating an actual recipe.

yeikel:
> I would like to pick this up. Since the issue is a year old I guess there is not much of a hurry and I can try out creating an actual recipe.

Please go for it

@timtebeek Might be able to assign it to you, but you can get a head start

timtebeek:
Thanks for the tag @yeikel and thanks for the offer @pstreef ! Let us know if you'd need any help with this.

In the past it's worked best to first get the boundaries straight with a draft PR that only adds tests.
It's a nice exercise in adversarial thinking on what code patterns in the wild might trip up the recipe.

With recipe development [the TreeVisitingPrinter might help](https://docs.openrewrite.org/concepts-explanations/tree-visiting-printer) explain what elements are available in the source tree.
Then from there you might want to look at [JavaTemplate to generate new code constructs](https://docs.openrewrite.org/concepts-explanations/javatemplate).

There could be some tricky aspects with regards to type attribution on the new code elements with JavaTemplate though, so that might not be the absolute best fit here. If you find that's troublesome we also support creating new tree elements using constructors, which are easier to type attribute.

pstreef:
@timtebeek 

Help is definitely welcome as I'm flailing about a bit with the implementation. However, messing around is a good way to learn, and I'm already starting to get somewhere.

[...]

PR #93, opened 2023-05-07
Ternary operator should not be nested

pstreef:
draft for: https://github.com/openrewrite/rewrite-static-analysis/issues/51

Recipe to replace nested ternary statements.


src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,88 @@
...
+                "but six months later will leave maintainers (or worse - future you) scratching their heads and cursing.\n\n" +
+                "Instead, err on the side of clarity, and use another line to express the nested operation as a separate statement.";
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(5);
+    }
+
+
```
timtebeek, line 40: Would be nice to override `getTags()` here to show [RSPEC-3358](https://rules.sonarsource.com/java/RSPEC-3358).

timtebeek:
Thanks for starting yet another one! Fun to see how you're progressing. :)

It's getting a bit late for an in depth review, so I'll keep it at some general comments for now that might help you find similar recipes to copy patterns from.

I noticed you're having some excess braces / blocks inserted; while ideally these are not inserted, sometimes you can simplify your logic by inserting them unconditionally, and then cleaning up after yourself by explicitly invoking `new RemoveUnneededBlock().getVisitor().visit...(...)`. We have a few such cases for unnecessary parenthesis, but this might suit you here as well. Mostly as a way to get the tests going first, and then clean things up after.

Good seeing the tests you have already. I think it's fine to start small for instance with return statements only, and expand from there. Any tests we don't cover yet you can then annotate with `@ExpectedToFail` & `@Issue("")` referencing a new backlog item for further improvements. Benefit of only looking at `return` for now is indeed that we don't need `else` blocks yet.

pstreef:
The thought to use something to clean the braces had crossed my mind but I thought that would be considered cheating 🙈 

Tests are green now but polishing is still required. Will leave this as a draft until and revisit it later this week (or possibly later today).

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,92 @@
...
+    public String getDisplayName() {
+        return "Ternary operators should not be nested";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Just because you can do something, doesn’t mean you should, and that’s the case with nested ternary operations. " +
+                "Nesting ternary operators results in the kind of code that may seem clear as day when you write it, " +
+                "but six months later will leave maintainers (or worse - future you) scratching their heads and cursing.\n\n" +
+                "Instead, err on the side of clarity, and use another line to express the nested operation as a separate statement.";
```
timtebeek, line 34: This gave me a good chuckle. At first I thought it's not as inline with our other recipe descriptions, but it turns out it's a direct copy of the description over on: https://sonarsource.atlassian.net/browse/RSPEC-3358

Now I'm wondering if we should rewrite this to be a bit shorter and more in line, and potentially not use the description one-to-one as seen on an external source. Something like:
```suggestion
        return "Nested ternary operators can be hard to read quickly. Prefer simpler constructs for improved readability.";
```

timtebeek:
Maybe we could add a few more tests just to be sure where the choice options are not literal strings. So for instance:
1. method calls
2. or `null`,
3. or expressions `"foo" + 123`,
4. or options preceded by a comment or newline.
 
All such cases could require slightly different handling, which could negatively affect either the functioning of code, or preservation of whitespace/comments for readability.

rpau:
Thanks @pstreef for your contribution! 

I would like to add to the @timtebeek list, a new test for lambda expressions. It might happen that we should add a new block entry before a new if statement. 

```
 s.stream().map(item -> item.startsWith("s") ? "ss".equals(item) ?"a": "b": "c").collect(Collectors.toSet());
```

should be translated to 
```
s.stream().map(item -> { 
  if (item.startsWith("s") {
    if ("ss".equals(item)) {
       return "a";
    } else {
       return "b";
    }
    return "c";
 }
}).collect(Collectors.toSet());
```

timtebeek:
> a new test for lambda expressions. It might happen that we should add a new block entry before a new if statement.
> 
> ```
>  s.stream().map(item -> item.startsWith("s") ? "ss".equals(item) ?"a": "b": "c").collect(Collectors.toSet());
> ```

Interesting case as well! :) Might be good to add a test that verifies no changes are made in such cases yet; so not `@ExpectedToFail`, but rather guard against accidental changes that could result in incorrect code, as the use within a lambda implies a return. Unless we're able to cover this case easily as well of course.

pstreef:
Thanks for these notes/additions. I will take them into account.

@rpau I'm doubting if that is really a rewrite everyone would be happy with. Some people prefer to keep their streams brace-less (which is also why you might see ternaries like these pop up) and in stead of adding this in line would expect a method call.

So this would be an opinionated approach, which I guess openrewrite is full of, but I'm still doubting which way would be "best"

rpau:
@pstreef  the problem extracting a method is that we can't infer a meaningful method name, and I agree that the strategy might be opinionated.

IMO, Skipping these cases for lambda expressions might create the impression that the recipe is not fully working. As a developer, I would like at least to identify those cases and then adapt the code.. extracting the method for instance, but if no changes are made, it is hard to identify if the code is satisfying the rule or not. 

We can add a boolean parameter to support those cases or not, but as a developer, I believe that we want to identify it to create trust in our recipes.

pstreef:
@timtebeek the build here keeps failing, probably because it depends on stuff that is not yet merged. Since the dependant change seems to be "coming soon" perhaps it's best to just keep this open until that is rounded up?

pstreef:
@timtebeek the build here keeps failing, probably because it depends on stuff that is not yet merged. Since the dependant change seems to be "coming soon" perhaps it's best to just keep this open until that is rounded up?

pstreef:
@timtebeek the build here keeps failing, probably because it depends on stuff that is not yet merged. Since the dependant change seems to be "coming soon" perhaps it's best to just keep this open until that is rounded up?

timtebeek:
Yes it's a bit unfortunate this past week.. hopefully resolved shortly! Right now you're dependent on whatever was the last snapshot pushed, which is unfortunate. We don't typically have such land lived branches, but then we also don't do many major releases. Here's hoping we can continue this soon.

pstreef:
no worries, I will keep an eye out for the release then!

knutwannheden:
Looking at the tests I think it would be cool if the recipe could turn some of the ternary operators into switch statements instead.

```java
                  return "a".equals(a) ? "a" : "b".equals(b) ? "b" : "nope";
```
could then be transformed to:
```java
                  return switch (a) {
                    case "a" -> "a";
                    case "b" -> "b";
                    default -> "nope";
                  };
```
or (without Java 14 switch expressions):
```java
                  switch (a) {
                    case "a":
                      return "a";
                    case "b":
                      return "b";
                    default:
                      return "nope";
                  };
```
This would of course only be applicable when all conditions always compare the same object to either literals or final variables and only be applicable to certain types (primitives, enums, and Strings). With Java 17 switch expressions with pattern matching these limitations could be relaxed a bit. The conditions would still have to refer to the same object, but the conditions could be different and also the types wouldn't have to be restricted.

Any thoughts on this?

knutwannheden:
@pstreef Thanks for your feedback.

> But then again, wouldn't that also mean that any if that matches this should be rewritten to such a switch, so maybe a separate recipe is warranted?

I also thought a bit about this and this could indeed be an idea. Although the analysis of the `if` statements is probably harder than that of a nested ternary expression, since it isn't just a single expression anymore. But regardless of this, I also can't quite make up my mind where the lines should be drawn between these recipes.

> Note there is already a recipe that checks: `switch` statements should have at least 3 `case` clauses
> Which means that your example actually should not be a switch 😬

You are probably referring to `MinimumSwitchCases`. There the `default` case is also considered a case, so a switch with two cases and a default would not be rewritten by `MinimumSwitchCases`. But possibly there should also be an inverse recipe which rewrites a cascade of `if` statements to a `switch`, at which point we could again discuss leaving this recipe as-is.

IMHO it would be cool if we could extend this recipe to also support mapping to Java 14 switch expressions. We could then in a next step consider pulling this logic out into a separate visitor or even recipe. But I think it is also fine if we delay all this to a later commit.

pstreef:
Hi @knutwannheden 

While investigating this possibility I found that with this PR deeper nested ternaries were not unwrapped. I'm now experimenting with a recursive approach which might also be extensible for the switch approach. Although I expect determining if a switch can be used is not so easy to implement.

While I agree it would be cool, we have to also be practical and consider how often would someone actually write such a (pretty badly) nested ternary and want it to be replaced with a switch.

I'm working now on getting all the comments properly placed with this recursive approach. After which I will update the PR.

knutwannheden:
Sounds good. Let's wait with the switch part. Although personally I would probably already use it instead of two nested ternaries, as it has the big advantage of staying an expression (which is why I personally also sometimes like using nested ternaries, when switch expressions are not available 🙈).

pstreef:
hi @knutwannheden
I find the challenge intriguing and since we're still blocked by v8 I can spend a little more time to try and figure it out.
 
I agree that in most of the test cases the nested ternary can be changed into a switch. I think the complexity comes from
1. Nesting can be in both true/false part. Maybe we can conclude that if there is any nesting in the truePart it's not a "switch candidate"
2. Nested ternaries can check different things, but if they all check the same variables equality to a thing it is a switch candidate.
3. If we include pattern matching for switch we can loosen the 2. constraint. But this is very dependant on the language level/confi. I'm not sure how to add this into consideration.

do you see any other checks I should add?

knutwannheden:
@pstreef I agree, this is a challenging recipe, which is also what makes it fun to write.

> 1. Nesting can be in both true/false part. Maybe we can conclude that if there is any nesting in the truePart it's not a "switch candidate"

Agreed. Although if it turns out to make your recursive algorithm simpler, we could of course also turn this into a nested switch expression.

> 2\. Nested ternaries can check different things, but if they all check the same variables equality to a thing it is a switch candidate.

This is probably the difficult part. An equality check could when using `Object#equals(Object)` also be "reversed", using referential equality (for primitives and enums) or even use `Objects#equals(Object, Object)`.

> 3\. If we include pattern matching for switch we can loosen the 2. constraint. But this is very dependant on the language level/confi. I'm not sure how to add this into consideration.

This would be really cool, but adds more complexity and could also be left for a follow-up commit.

I can't think of any additional checks in addition to what you already identified.

pstreef:
@knutwannheden it is indeed both fun and a challenge, but not as big a challenge as finding enough time for it. I'm well underway though!

image

pstreef:
@knutwannheden I've found some more time. I have a lot of green tests and a few which we can either leave as `@ExpectedToFail` or maybe you can help (especially with the lost comment one)

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,371 @@
...
+                            .withPrefix(Space.SINGLE_SPACE)
+            );
+        }
+
+        private J.Case toCase(final J.Identifier switchVar, final J.Ternary ternary) {
+            Expression compare;
+            if (ternary.getCondition() instanceof J.MethodInvocation) {
+                J.MethodInvocation inv = ((J.MethodInvocation) ternary.getCondition());
+                if (isObjectsEquals(inv)) {
+                    doAfterVisit(new RemoveImport<>("java.util.Objects"));
```
knutwannheden, line 241:
```suggestion
                    maybeRemoveImport("java.util.Objects");
```

knutwannheden (commented):
Apart from a few minor comments, I think the code looks very good and I think this will be a very useful recipe to have!

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,371 @@
...
+                    final ExecutionContext executionContext
+            ) {
+                if (cu.getMarkers()
+                        .findFirst(JavaVersion.class)
+                        .filter(javaVersion -> javaVersion.getMajorVersion() >= 14)
+                        .isPresent()) {
+                    doAfterVisit(new UseSwitchExpressionVisitor());
+                }
+                doAfterVisit(new UseIfVisitor());
+                return super.visitCompilationUnit(cu, executionContext);
```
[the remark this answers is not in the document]
pstreef: yes, good catch!

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
pstreef: done!

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,368 @@
...
+                    result = (J.Identifier) inv.getArguments().get(0);
+                }
+                if (isObjectsEquals(inv)) {
+                    //one has to be constant, other not
+                    J first = inv.getArguments().get(0);
+                    J second = inv.getArguments().get(1);
+                    result = xorVariable(first, second);
+                }
+            }
+            if (ternary.getCondition() instanceof J.Binary) {
```
knutwannheden, line 311: Could be an `else if` AFAICT.

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,368 @@
...
+                J.Binary bin = (J.Binary) ternary.getCondition();
+                result = xorVariable(bin.getLeft(), bin.getRight());
```
knutwannheden, line 313: In the body of this `if` we need to check the `J.Binary#getOperator()` result. I think currently we only support `==` (and not `!=`), so we should make sure that this is indeed also the case. A corresponding test would also be nice.

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,368 @@
...
+            }
+            return Optional.ofNullable(result);
+
+        }
+
+        @Nullable
+        private static J.Identifier xorVariable(J first, J second) {
+            J.Identifier result = null;
+            if (isVariable(first) && isVariable(second)) {
+                return null;
```
knutwannheden, line 323: In the future we could maybe remove this limitation. For example we could exclude any variables which are constants (i.e. where the type has the `final` and `static` flags) so that we can allow rewriting ternaries like: `v == CONST_1 ? "1" : v == CONST_2 ? "2" : ""`.

knutwannheden:
@pstreef Sorry about these late review comments. I didn't look carefully enough the first time around.

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,368 @@
...
+                return false;
+            }
+            J.Identifier identifier = (J.Identifier) maybeVariable;
+            return identifier.getFieldType() != null;
+        }
+
+        private static boolean isObjectsEquals(J.MethodInvocation inv) {
+            if (inv.getSelect() instanceof J.Identifier) {
+                J.Identifier maybeObjects = (J.Identifier) inv.getSelect();
+                boolean isObjects = TypeUtils.isOfType(maybeObjects.getType(), JavaType.buildType("java.util.Objects"));
```
knutwannheden, line 348:
```suggestion
                boolean isObjects = TypeUtils.isOfClassType(maybeObjects.getType(), "java.util.Objects");
```

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
pstreef: Correct, will change!

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
pstreef: good catch, I will extend!

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
pstreef: The future is now

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,368 @@
...
+                            null
+                    )).withAfter(Space.SINGLE_SPACE))),
+                    JContainer.build(Collections.emptyList()),
+                    JRightPadded.build(ternary.getFalsePart())
+            );
+        }
+
+        private Optional<J.Identifier> findConditionIdentifier(final J.Ternary ternary) {
+            J.Identifier result = null;
+            if (ternary.getCondition() instanceof J.MethodInvocation) {
```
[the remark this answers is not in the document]
pstreef: you are correct, I pushed a test + fix!

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
pstreef: done!

knutwannheden:
I can't think of any other cases to test. Once we release and deploy it to public.moderne.io I wouldn't be surprised if we discover a few corner cases more, we didn't think of, but that is normal 😄

knutwannheden:
> I'd love your input on this failing test: doReplaceMultiLevelTernariesWithComments
> I'm missing a comment that is part of a "before" of a JLeftPadded. if I use getFalsePart it automatically extracts the element and the comment is dropped. How should I get this out?

I can take a look at this. But I think we could also integrate this PR as is and fix this problem later.

src/main/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNested.java
```diff
@@ -0,0 +1,400 @@
...
+    public String getDescription() {
+        return "Nested ternary operators can be hard to read quickly. Prefer simpler constructs for improved readability. " +
+                "If supported, this recipe will try to replace nested ternaries with switch expressions.";
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(5);
+    }
+
```
timtebeek, line 68: No need to override the default value I suppose.
```suggestion
```

src/test/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNestedTest.java
```diff
@@ -0,0 +1,1315 @@
...
+                  import java.util.Arrays;
+                  import java.util.List;
+                  import java.util.stream.Collectors;
+                  class Test {
+                    public Set<String> makeASet() {
+                        List<String> s = Arrays.asList("a","b","c","nope");
+                        return s.stream().map( /* look a lambda */ item ->
+                                //look a ternary
+                                {
+                                    if (item.startsWith("a")) {
```
timtebeek, line 426: Intuitively I'd have expected this comment on the `if`, not on the `block` Would that be hard to change to reduce friction on any PRs this generates?

src/test/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNestedTest.java
```diff
@@ -0,0 +1,1315 @@
...
+                  class Test {
+                    public String determineSomething(String a, String b) {
+                      return Objects.equals("a", a) ? "a" : Objects.equals("b", a) ? b : "nope";
+                    }
+                  }
+                  """,
+                """
+                  class Test {
+                    public String determineSomething(String a, String b) {
+                      return switch (a) {
```
timtebeek, line 807: Don't we lose null safety here?

timtebeek (approved):
Impressive work! I've added a few last minor comments, but trust your judgement in resolving those before a merge.

As a last comment: Perhaps it'd be nice to clearly divide the tests over ~3 nested classes: ifs / simple switch / switch expressions.

src/test/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNestedTest.java
pstreef: You are right, I did not know switch needs a nonNull arg.
An easy way around this is to skip `Objects.equals`, alternatively we can add a null check and early exit with the "default" case. What do you think?

timtebeek: I'd say we can skip `Objects.equals` now and revisit when Java 21 is out, as Java 18+ allows for null in switch.

src/test/java/org/openrewrite/staticanalysis/TernaryOperatorsShouldNotBeNestedTest.java
pstreef: I remember playing with this and doubting where to put it. It's not hard to move and I think I agree with you. It looks better on the if.

timtebeek (approved):

pstreef:
Regarding the nesting. I made it a bit better for the already nested `SwitchExpressionSupported`. There is no "simple" switch, so I split expression, if and no change

timtebeek:
Maybe good to also annotated one or two representative tests with `@DocumentExample`; that way they ought to show up in the docs (soon).

Issue #99, opened 2023-05-09
RemoveExtraSemicolons removes newline if semicolon is in front of statement


pstreef:
sample:
```
class Test {
    void test() {
        int a = 1;
        ;int b = 2;
    }
}
```
expected
```
class Test {
    void test() {
        int a = 1;
        int b = 2;
    }
}
```
actual
```
class Test {
    void test() {
        int a = 1; int b = 2;
    }
}
```

PR #100, opened 2023-05-09
RemoveExtraSemicolons removes newline if semicolon is in front of statement

pstreef:
Add test with @ExpectedToFail
https://github.com/openrewrite/rewrite-static-analysis/issues/99


timtebeek:
Thanks for adding the replicating test already! I expect the fix is not far off either; likely prepend the prefix from the semicolon to the statement following the semicolon, if both are on the same line(?). Not expected of you to immediately fix that; but thought to note it here for when the pipeline succeeds to briefly explore before a merge.

timtebeek (approved):

pstreef:
Something like this? 🤔

src/test/java/org/openrewrite/staticanalysis/RemoveExtraSemicolonsTest.java
```diff
@@ -165,4 +164,54 @@ void test() {
...
+                      a = 6;;//then we set a to 6
+                      if (a == 6) { //if a is 6
+                        ;a = 7;;//then if a is 6 we set a to 7
+                      }
+                      ;;
+                      ;//next we set a to 8
+                      ;a = 8;
+                      return a;
+                      ;
+                      //we are done!
```
timtebeek, line 191: Awesome!

timtebeek: Maybe we could add in block comments `/* foo; */` just to be sure,
but otherwise I really like the variations you've applied here.

timtebeek:
Just merged main into this branch; we're still dependent on two branches in other repositories:
https://github.com/openrewrite/rewrite/pull/3118 for PreCondition
https://github.com/openrewrite/rewrite-recipe-bom/pull/16 for rewrite-analysis version

src/main/java/org/openrewrite/staticanalysis/RemoveExtraSemicolons.java
```diff
@@ -52,20 +59,33 @@ public Duration getEstimatedEffortPerOccurrence() {
...
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return new JavaIsoVisitor<ExecutionContext>() {
+
             @Override
-            public J.Empty visitEmpty(J.Empty empty, ExecutionContext ctx) {
-                if (getCursor().getParentTreeCursor().getValue() instanceof J.Block) {
-                    return null;
+            public J.Block visitBlock(final J.Block block, final ExecutionContext executionContext) {
+                final Iterator<Statement> iterator = block.getStatements().iterator();
+                final List<Statement> result = new ArrayList<>(); //todo should this be arraylist?
```
timtebeek, line 66: Is there anything left to do here?

pstreef: I was not sure if maybe in this tree structure you might prefer a different type of list implementation. Hence the comment. But from what I've seen by now I think this is fine

timtebeek:
Glad to see everything passing now that 8.0 has been merged! Didn't expect anything else, but still nice to see. Would you have time for a review @joanvr

joanvr (approved):
Looks really good! Thanks for the fix and new tests! Just a minor unused import in the tests.

Issue #101, opened 2023-05-09
ExplicitInitialization removes initialization from final fields when using @Value.


pstreef:
sample:
```
import lombok.Value;
@Value
class Test {
    boolean b = false;
}
```
expected: no change
Actual:
```
import lombok.Value;
@Value
class Test {
    boolean b;
}
```
this results in the field b being added the the @RequiredArgsConstructor. Changing it's signature.

timtebeek:
That's a very specific edge case I hadn't thought of; I think we can make the recipe back off in the presence of lombok.Value, just to be safe here. That would not require Lombok support, as we only look at the presence of the annotation, not what the annotation would result in.

PR #102, opened 2023-05-09
ExplicitInitialization removes initialization from final fields when using @Value

pstreef:
Add test with @ExpectedToFail
issue: https://github.com/openrewrite/rewrite-static-analysis/issues/101


rpau (approved):

rpau:
Thanks for helping to reproduce the issue

PR #108, opened 2023-05-15
Do not remove explicit init in case lombok.Value is present on enclosing class

pstreef:
fixes: https://github.com/openrewrite/rewrite-static-analysis/issues/101
Do not remove explicit init in case lombok.Value is present on enclosing class


timtebeek (approved):
Thanks for yet another fix! Looks good to me. I expect the tests to pass as soon as we merge https://github.com/openrewrite/rewrite/pull/3118 🤞🏻

knutwannheden:
I think it looks good. Unrelated to this PR (I think) it looks like the recipe doesn't handle anonymous sublcasses. So code like this is currently not updated:
```java
              class Test {
                  Object o = new Object() {
                      boolean b = false;
                  };
              }
```

PR #114, opened 2023-05-26
added a failing test for LambdaBlockToExpression

xshen053:
## What's changed?
added a failing test for LambdaBlockToExpression

## What's your motivation?
#1 

Not sure if the name of my unit test is okay

Still get myself familiar with LST, probably will have new discovery later

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
@@ -52,7 +52,7 @@ class LambdaBlockToExpressionTest implements RewriteTest {
     }
 
     @Test
-    @Disabled
+//    @Disabled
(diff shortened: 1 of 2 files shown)


kunli2 (approved):
Thanks for contribution!

src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
```diff
@@ -48,4 +48,31 @@ class Test {
           )
         );
     }
+
+    @Test
+    void simplifyLambdaBlockToExpressionWithComments() {
```
sambsnyd, line 53: Can you add a `@Disabled` annotation and an `@Issue` annotation which contains a link to the issue description this test reproduces? Thanks!

xshen053: ok

PR #118, opened 2023-06-04
Added recipe to replace `Optional#isPresent` to `Optional#ifPresent `

kunli2:
This is a following up PR of https://github.com/openrewrite/rewrite/pull/3074
To add a recipe to replace `Optional#isPresent` with `Optional#ifPresent`.
That PR is based on rewrite 7 and is no longer needed since the recipes have been moved to repo `rewrite-static-analysis`.

I migrated the recipe to rewrite 8 and updated some code.

@srmalkan , thanks for your initial contribution! It's a useful recipe. let's start from here and address a few comments and concerns and then we can merge.


kunli2:
I did these major updates:
1. Removed `IsBlockLambdaConvertibleVisitor` and use existing visitors to support the LambdaConvertible check.
2. Removed search result marker usage in the recipe.
3. Support nested cases (however the generated variable names are in revert order, which works fine but not looks perfect, will leave this as an improvement in the future).

Issue #119, opened 2023-06-05
JavaDoc: RemoveJavaDocAuthorTag and RemoveEmptyJavaDocParameters should remove empty JavaDoc after cleanup


koppor:
```diff
 /**
- * @author jzieren
  */
```

Expected:

```diff
- /**
-  * @author jzieren
-  */
```

Similar for `RemoveEmptyJavaDocParameters` - and thus refs https://github.com/openrewrite/rewrite-static-analysis/issues/98.

Maybe both `RemoveJavaDocAuthorTag` and `RemoveEmptyJavaDocParameters` should remove empty lines and empty JavaDoc after cleanup.

timtebeek:
I've added https://github.com/openrewrite/rewrite-static-analysis/pull/120 to hopefully resolve this quickly; It only handles the above case; not also #98 which concerns trimming in addition to removing empty tags. On the PR we can discuss potential broader appliability.

koppor:
@timtebeek Tried it today with recent versions - and JavaDoc is still empty - and not removed:

```diff
 /**
  *
- * @author airhacks.com
  */
 public class Configurator {
```

koppor:
Ah, if there is no `*` around, it works:

```diff
 
-/**
- * @author adam-bien.com
- */
 public class ViewLoader {
 
```

timtebeek:
Ah yes the approach taken in https://github.com/openrewrite/rewrite-static-analysis/pull/120 does not account for additional new lines. With a new unit test and likely small code change that can probably be covered; Would you want to add those changes?

Issue #98, opened 2023-05-08
RemoveEmptyJavaDocParameters should remove "empty" lines


koppor:
```diff
     /**
      * Renders this page and returns an image representation of itself.
      *
-     * @param width
-     * @param height
+     *
      */
```

Result:

```java
    /**
     * Renders this page and returns an image representation of itself.
     *
     *
     */
```

Better would be 

```java
    /**
     * Renders this page and returns an image representation of itself.
     */
```

koppor:
Similar:

```diff
-     * @param citationGroups
-     * @param style
      *
-     */
+     **/
```

https://github.com/JabRef/jabref/blob/68b72e0a93c1083f79c5710bc5e6f9607f58e07f/src/main/java/org/jabref/logic/openoffice/style/OOProcessNumericMarkers.java#L24

timtebeek:
A similar issue was resolved through https://github.com/openrewrite/rewrite-static-analysis/pull/120
Anyone welcome to produce the same for this recipe.

alixina9:
Hello, I'd like to work on this issue if that's possible.

timtebeek:
Hi @alixina9 , thanks for your offer to help! Would you need any assistance in getting started? Feel free to open a draft pull request that replicates the issue. Once that's open I can assign the issue to you.

PR #120, opened 2023-06-05
Remove blank JavaDoc blocks after RemoveJavaDocAuthorTag

timtebeek:
## What's changed?
Check if the remaining JavaDoc lines are all blank after removing author tag.

## What's your motivation?
Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/119

Usage of PrintOutputCapture & Cursor to check for blank lines; wasn't sure if there's any other mechanism.
 
We could pull `isBlank` into `JavaDoc`, if we want to also fix this for #98, which has a slightly broader scope: trim in addition to remove when fully empty.

### Checklist


knutwannheden (approved):
LGTM 👍

Issue #22, opened 2023-03-31
Use `contentEquals` to compare StringBuilder to a String


yeikel:
![image](https://blog.jetbrains.com/wp-content/uploads/2023/01/image-9.gif)

Source https://blog.jetbrains.com/idea/2023/01/intellij-idea-2023-1-eap-2/?_gl=1*wgqxap*_ga*NDc2NTIxNzM3LjE2NzMyODIxOTY.*_ga_9J976DJZ68*MTY4MDI3NDg5Mi4xMS4xLjE2ODAyNzU2ODUuMC4wLjA.&_ga=2.15990002.513565325.1680274893-476521737.1673282196#NewJavainspections

timtebeek:
Hope you don't mind I unassigned you @bhuwanupadhyay 
I want to open this issue back up again to anyone to pick up.
Feel free to open up a draft pull request if you'd like to contribute!

rpau:
@AlekSimpson reported that this is completed

PR #124, opened 2023-06-16
new recipe that changes .equals() to .contentEquals() when comparing …

AlekSimpson:
…a string with a StringBuilder

## What's changed?
This PR adds a new recipe that finds when `.equals()` is being used to compare a string and a `StringBuilder`, `StringBuffer` or `CharSequence`.

## What's your motivation?
[#22](https://github.com/openrewrite/rewrite-static-analysis/issues/22)

I think the precondition checks should be correct but maybe they could be narrowing things down more. 

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
@@ -32,16 +32,14 @@ import java.util.stream.Stream;
 
 public class EqualsToContentEquals extends Recipe {
     private static final MethodMatcher equals_matcher = new MethodMatcher("java.lang.String equals(..)");
-    private static final List<String> TYPE_NAMES = Arrays.asList(
-            "java.lang.StringBuffer",
-            "java.lang.StringBuilder",
-            "java.lang.CharSequence"
-    );
-    @SuppressWarnings("unchecked")
-    private static final TreeVisitor<?, ExecutionContext> PRECONDITION =
-            Preconditions.or(TYPE_NAMES.stream().map(s -> new UsesType<>(s, false)).toArray(UsesType[]::new));
-    private static final List<MethodMatcher> toString_matchers = TYPE_NAMES.stream()
-            .map(obj -> new MethodMatcher(obj + " toString()")).collect(Collectors.toList());
+    private static final TreeVisitor<?, ExecutionContext> PRECONDITION = Preconditions.or(
+            new UsesType<>("java.lang.StringBuffer", false),
+            new UsesType<>("java.lang.StringBuilder", false),
+            new UsesType<>("java.lang.CharSequence", false));
+    private static final List<MethodMatcher> TOSTRING_MATCHERS = Arrays.asList(
+            new MethodMatcher("java.lang.String toString()"),
+            new MethodMatcher("java.lang.StringBuffer toString()"),
+            new MethodMatcher("java.lang.StringBuilder toString()"));
 
     @Override
     public String getDisplayName() {
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.openrewrite.staticanalysis;
 
 import org.openrewrite.ExecutionContext;
@@ -13,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class EqualsToContentEquals extends Recipe {
     private static final MethodMatcher equals_matcher = new MethodMatcher("java.lang.String equals(..)");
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+public class EqualsToContentEquals extends Recipe {
+    private static final MethodMatcher equals_matcher = new MethodMatcher("java.lang.String equals(..)");
+    private static final List<String> TYPE_NAMES = Arrays.asList(
+            "java.lang.StringBuffer",
+            "java.lang.StringBuilder",
+            "java.lang.CharSequence"
+    );
+    @SuppressWarnings("unchecked")
+    private static final TreeVisitor<?, ExecutionContext> PRECONDITION =
+            Preconditions.or(TYPE_NAMES.stream().map(s -> new UsesType<>(s, false)).toArray(UsesType[]::new));
```
timtebeek, line 42: Maybe for this limited number of types we can keep things simple and just create invoke `new UsesType` three times? Helps readability.

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final List<MethodMatcher> toString_matchers = TYPE_NAMES.stream()
+            .map(obj -> new MethodMatcher(obj + " toString()")).collect(Collectors.toList());
```
timtebeek, line 44:
```suggestion
    private static final TreeVisitor<?, ExecutionContext> PRECONDITION = Preconditions.or(
            new UsesType<>("java.lang.StringBuffer", false),
            new UsesType<>("java.lang.StringBuilder", false),
            new UsesType<>("java.lang.CharSequence", false));
    private static final List<MethodMatcher> toString_matchers = Arrays.asList(
            new MethodMatcher("java.lang.String toString()"),
            new MethodMatcher("java.lang.StringBuffer toString()"),
            new MethodMatcher("java.lang.StringBuilder toString()"));
```

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+                            JavaType.buildType("java.lang.CharSequence")
+                    );
+
+                    if (TYPES.anyMatch(type -> TypeUtils.isOfType(newArg.getType(), type))) {
+                        // strip out the toString() on the argument
+                        List<Expression> args = new ArrayList<>(1);
+                        args.add(newArg);
+                        m = m.withArguments(args);
+                        // rename the method to contentEquals
+                        methodName = m.getName().withSimpleName("contentEquals");
```
timtebeek, line 86:
```suggestion
                        return m.witName("contentEquals");
```

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+                    }
+                }
+            }
+
+            return m.withName(methodName);
```
timtebeek, line 91:
```suggestion
            return m;
```

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(PRECONDITION, new EqualsToContentEqualsVisitor());
+    }
+
+    private static class EqualsToContentEqualsVisitor extends JavaIsoVisitor<ExecutionContext> {
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            J.MethodInvocation m = super.visitMethodInvocation(mi, ctx);
+            J.Identifier methodName = m.getName();
```
timtebeek, line 63:
```suggestion
```

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+                if (toString_matchers.stream().anyMatch(matcher -> matcher.matches(argument))) {
+                    J.MethodInvocation inv = (J.MethodInvocation) argument;
+                    Expression newArg = inv.getSelect();
+                    if (inv.getSelect() == null) { return m; }
+
+                    Stream<JavaType> TYPES = Stream.of(
+                            JavaType.buildType("java.lang.StringBuilder"),
+                            JavaType.buildType("java.lang.StringBuffer"),
+                            JavaType.buildType("java.lang.CharSequence")
+                    );
```
timtebeek, line 78: We can probably also move these to a field above.

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,94 @@
...
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            J.MethodInvocation m = super.visitMethodInvocation(mi, ctx);
+            J.Identifier methodName = m.getName();
+            // create method matcher on equals(String)
+            if (equals_matcher.matches(m)) {
+                Expression argument = m.getArguments().get(0);
+
+                // checks whether the argument is a toString() method call on a StringBuffer or CharSequence
+                if (toString_matchers.stream().anyMatch(matcher -> matcher.matches(argument))) {
```
timtebeek, line 69:
```suggestion
                if (TOSTRING_MATCHERS.stream().anyMatch(matcher -> matcher.matches(argument))) {
```

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
timtebeek, line 86:
```suggestion
                        return m.withArguments(Collections.singletonList(newArg))
                                .withName(m.getName().withSimpleName("contentEquals"));
```

src/test/java/org/openrewrite/staticanalysis/EqualsToContentEqualsTest.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Test
+    public void onlyRunsOnCorrectInvocations() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              class SomeClass {
+                  boolean foo(int number, String str) {
```
timtebeek, line 66:
```suggestion
                  boolean foo(Integer number, String str) {
```

timtebeek (commented):
In terms of tests it would be good to have cases where either the first or the second select is not a string variable or constant, but for instance a method invocation, to be sure we handle those fairly predictable scenarios well.

src/test/java/org/openrewrite/staticanalysis/EqualsToContentEqualsTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion())
+          .recipe(new EqualsToContentEquals());
+    }
+
+    @Test
+    public void replaceStringBuilder() {
```
timtebeek, line 35: Try to always annotate one representative test with `@DocumentExample`, such that we get code blocks to show up in the docs, such as for instance in: https://docs.openrewrite.org/recipes/java/testing/cleanup/removetestprefix
```suggestion
    @Test
    @DocumentExample
    public void replaceStringBuilder() {
```

src/main/java/org/openrewrite/staticanalysis/EqualsToContentEquals.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2021 the original author or authors.
```
timtebeek, line 2: Might want to adjust ./gradle/licenceHeader.txt and this here to set the right year.

AlekSimpson:
I added that test for if the recipe will run correctly on selects that are a method call.

timtebeek:
Great work! I've added another polish commit in 18287b2ff407b67bebbadac81a9f717ed39c12f5; with changes such as

- Use markdown in `getDisplayName()` and `getDescription()` around code elements, for better readability in the docs
- Remove nested if statements in favor of early returns; this helps keep the code readable
- Minimize use of loops through a wildcard matcher on `java.lang.* toString()` and `TypeUtils.isAssignableTo("java.lang.CharSequence", JavaType)`; we try to avoid Streams for performance reasons (fewer object allocations), and it leads less code as well.
- Minimize `public` visibility in tests

Hope you don't mind, and I look forward to the next one!

timtebeek (approved):

Issue #44, opened 2022-04-16
`hashCode` and `toString` should not be called on `array` instances. `RSPEC-2116`


yeikel:
> While hashCode and toString are available on arrays, they are largely useless. hashCode returns the array’s "identity hash code", and toString returns nearly the same value. Neither method’s output actually reflects the array’s contents. Instead, you should pass the array to the relevant static Arrays method.

```java
    String argStr = args.toString(); // Noncompliant
    int argHash = args.hashCode(); // Noncompliant
```

```java
    String argStr = Arrays.toString(args);
    int argHash = Arrays.hashCode(args);

```

More examples with `Strings`

```java
int[] s = new int[]{1,2,3};
System.out.println(s);
Sytem.out.println(String.format("s=%s",s));

// .. any other usage where a String is expected
```

```java

int[] s = new int[]{1,2,3};
System.out.println(Arrays.toString(s));
Sytem.out.println(String.format("s=%s",Arrays.toString(s)));

// .. any other usage where a String is expected

```

https://rules.sonarsource.com/java/RSPEC-2116

timtebeek:
Some implementation notes:
- both methods are available on all objects, so any implementation is performance sensitive: limit & return early where possible
- probably best to implement `hashCode` separate from `toString`, as that's slightly easier
- `toString()` can also be implied for arguments headed into `String.format`; that's probably best handled in a separate recipe or visitor to check the arguments to `String.format()` and similar methods such as `printf`.

timtebeek:
Reopened until we also merge in the separate recipes for toString https://github.com/openrewrite/rewrite-static-analysis/pull/126

PR #126, opened 2023-06-19
Alek/remove to string calls from array instances

AlekSimpson:
## What's changed?
toString should not be called on array instances. RSPEC-2116

## What's your motivation?
[#44](https://github.com/openrewrite/rewrite-static-analysis/issues/44)

I think the recipe is pretty much finished, only two things is there is an edge case with String.format() calls I still haven't quite worked out. It gives an error when I try and run the tests on them. Second thing is that I'm not sure how to get the preconditions to check for if the code is using the array type before it runs. 

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
@@ -16,13 +16,18 @@
 package org.openrewrite.staticanalysis;
 
 import org.openrewrite.*;
-import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.search.UsesMethod;
-import org.openrewrite.java.tree.*;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 public class RemoveToStringCallsFromArrayInstances extends Recipe {
@@ -36,7 +41,7 @@ public class RemoveToStringCallsFromArrayInstances extends Recipe {
             "java.lang.StringBuilder append(Object)",
             "java.io.PrintStream format(String, Object[])"
     );
-    private  static final TreeVisitor<?, ExecutionContext> PRECONDITION = Preconditions.or(
+    private static final TreeVisitor<?, ExecutionContext> PRECONDITION = Preconditions.or(
             PATTERNS.stream().map(UsesMethod::new).toArray(UsesMethod[]::new)
     );
     private static final List<MethodMatcher> METHOD_MATCHERS = PATTERNS.stream().map(MethodMatcher::new).collect(Collectors.toList());
@@ -54,54 +59,53 @@ public class RemoveToStringCallsFromArrayInstances extends Recipe {
     @Override
     public String getDescription() {
         return "The result from `toString()` calls on arrays is largely useless. The output does not actually reflect" +
-                " the contents of the array. `Arrays.toString(array)` give the contents of the array.";
+               " the contents of the array. `Arrays.toString(array)` give the contents of the array.";
     }
 
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return Preconditions.check(PRECONDITION, new RemoveToStringFromArraysVisitor());
+        return new RemoveToStringFromArraysVisitor();
+// FIXME not sure we actually need a precondition here, but it currently doesn't cover OBJECT_TOSTRING_MATCHER and OBJECTS_TOSTRING_MATCHER
+//        return Preconditions.check(PRECONDITION, new RemoveToStringFromArraysVisitor());
     }
 
-    private static class RemoveToStringFromArraysVisitor extends JavaIsoVisitor<ExecutionContext> {
+    private static class RemoveToStringFromArraysVisitor extends JavaVisitor<ExecutionContext> {
         @Override
-        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
-            J.MethodInvocation m = super.visitMethodInvocation(mi, ctx);
-
-            if (OBJECT_TOSTRING_MATCHER.matches(m) || OBJECTS_TOSTRING_MATCHER.matches(m)) {
-                Expression select = m.getSelect();
+        public J visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            if (OBJECT_TOSTRING_MATCHER.matches(mi) || OBJECTS_TOSTRING_MATCHER.matches(mi)) {
+                Expression select = mi.getSelect();
                 assert select != null;
 
                 if (!(select.getType() instanceof JavaType.Array)) {
-                    return m;
+                    return mi;
                 }
 
                 J.MethodInvocation newInv = JavaTemplate.builder("Arrays.toString(#{anyArray(java.lang.Object)})")
                         .imports("java.util.Arrays")
                         .build()
-                        .apply(getCursor(), m.getCoordinates().replace(), select);
+                        .apply(getCursor(), mi.getCoordinates().replace(), select);
                 maybeAddImport("java.util.Arrays");
                 return newInv;
-            }else if (METHOD_MATCHERS.stream().anyMatch(matcher -> matcher.matches(m))) {
+            } else if (METHOD_MATCHERS.stream().anyMatch(matcher -> matcher.matches(mi))) {
                 // deals with edge cases where .toString() is called implicitly
-                List<Expression> arguments = m.getArguments();
+                List<Expression> arguments = mi.getArguments();
                 for (Expression arg : arguments) {
                     if (arg.getType() instanceof JavaType.Array) {
                         Cursor c = getCursor();
-                        c.putMessage("METHOD_KEY", m);
+                        c.putMessage("METHOD_KEY", mi);
                         break;
                     }
                 }
             }
 
-            return m;
+            return super.visitMethodInvocation(mi, ctx);
         }
 
         @Override
         public Expression visitExpression(Expression exp, ExecutionContext ctx) {
-            Expression e = super.visitExpression(exp, ctx);
-            Cursor c = getCursor().dropParentWhile(J.Parentheses.class::isInstance);
-            System.out.println(c.getNearestMessage("METHOD_KEY") != null);
-            if (c.getMessage("METHOD_KEY") != null) {
-                if (e.getType() instanceof JavaType.Array) {
+            Expression e = (Expression) super.visitExpression(exp, ctx);
+            if (e.getType() instanceof JavaType.Array) {
+                Cursor c = getCursor().dropParentWhile(is -> is instanceof J.Parentheses || !(is instanceof Tree));
+                if (c.getMessage("METHOD_KEY") != null) {
                     Expression newExpression = JavaTemplate.builder("Arrays.toString(#{anyArray(java.lang.Object)})")
                             .imports("java.util.Arrays")
                             .build()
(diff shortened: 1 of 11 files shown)


timtebeek:
There's some conflicts here; could you resolve those? Did you perhaps start your branch from your previous branch?

AlekSimpson:
I figured out the edge case and also the precondition check. I think this recipe is functionally complete, but maybe there could be last minute polishes I don't see.

sambsnyd (approved):
Good work! This will be a useful recipe. 👍 
After you address my feedback this looks good to check in, no need to seek another round of review from me.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,116 @@
...
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+
+import java.util.List;
+
+public class RemoveToStringCallsFromArrayInstances extends Recipe {
```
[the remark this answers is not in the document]
knutwannheden: Could be https://rules.sonarsource.com/java/RSPEC-2116

I would also like to propose that we change the name, since we are either replacing `Object#toString()` with `Arrays.toString()` and sometimes we just add the latter. So maybe something like `UseArraysToStringForArrays`?

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final MethodMatcher TO_STRING_MATCHER = new MethodMatcher("java.lang.Object toString(..)");
```
knutwannheden, line 33:
```suggestion
    private static final MethodMatcher TO_STRING_MATCHER = new MethodMatcher("java.lang.Object toString()");
```

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final MethodMatcher PRINTLN_MATCHER = new MethodMatcher("java.io.PrintStream println(..)");
```
knutwannheden, line 34: As we only want to match `print(Object)` and `println(Object)` here:
```suggestion
    private static final MethodMatcher PRINTLN_MATCHER = new MethodMatcher("java.io.PrintStream print*(java.lang.Object)");
```

Further, I think there will be quite a few more methods like this one. Some which come to mind: `Objects#toString(Object)`, `String#valueOf(Object)`, `StringBuilder#insert(int, Object)`, `StringBuilder#append(Object)`, etc.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final MethodMatcher STR_FORMAT_MATCHER = new MethodMatcher("java.lang.String format(..)");
```
knutwannheden, line 35: We should also cover `java.io.PrintStream#format(String, Object...)`.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,116 @@
...
+                Expression select = m.getArguments().get(0);
+                String builder_string = "System.out.println(Arrays.toString(#{anyArray(java.lang.String)}))";
+
+                return buildReplacement(builder_string, m, select);
+            }else if (STR_FORMAT_MATCHER.matches(m)) {
+                List<Expression> arguments = m.getArguments();
+                StringBuilder builder_string = new StringBuilder("String.format(#{any(java.lang.String)}");
+                boolean arrayExists = false;
+
+                for (int i = 0; i<arguments.size(); i++) {
```
knutwannheden, line 76: For reasons of efficiency (and simplicity in a way) I am wondering if we shouldn't instead use cursor messaging and transform this bottom-up. What I mean is that you in this method would add a message to the cursor to indicate that you have found one of the methods which need transforming. After that you would do the `super` call and then you would add an override for `visitExpression()` which in case the parent cursor (skipping over any intermediate `J.Parentheses` cursors) has the mentioned message set and the expression is array-typed would return the array wrapped in `Arrays#toString()`. Does that make sense to you? This logic would then be exactly the same for both `STR_FORMAT_MATCHER` and `PRINTLN_MATCHER` and all the other methods I mentioned.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(Preconditions.or(
+                new UsesMethod<>(TO_STRING_MATCHER),
+                new UsesMethod<>(PRINTLN_MATCHER),
+                new UsesMethod<>(STR_FORMAT_MATCHER)
+        ), new RemoveToStringFromArraysVisitor());
+    }
+
+    private static class RemoveToStringFromArraysVisitor extends JavaIsoVisitor<ExecutionContext> {
```
knutwannheden, line 55: There is one more important override missing here: `visitBinary()`. This is for the case when `+` is being used to concatenate a string with an array.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
timtebeek: String.formatted is another option, as well as the printf variants.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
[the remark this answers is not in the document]
AlekSimpson: It is https://rules.sonarsource.com/java/RSPEC-2116. My bad for not tagging it initially. That name change makes sense to me.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
AlekSimpson: Yep this makes sense. I'll get started on this change.

AlekSimpson:
Ok, so I've tried to incorporate all the feedback from this thread. I still need to add more tests for a bunch of the other methods mentioned above. However, I think I should wait to do that till I have the cursor messaging working. I've updated the branch with my implementation of the cursor messaging but it doesn't work. The message never seems to be found in the `visitExpression()` visitor and I'm not sure why.

knutwannheden:
> Ok, so I've tried to incorporate all the feedback from this thread. I still need to add more tests for a bunch of the other methods mentioned above. However, I think I should wait to do that till I have the cursor messaging working. I've updated the branch with my implementation of the cursor messaging but it doesn't work. The message never seems to be found in the `visitExpression()` visitor and I'm not sure why. @knutwannheden

@AlekSimpson You got it almost all right! The important thing here is that the `super.visitMethodInvocation()` call comes after the message is added to the cursor. Otherwise `visitExpression()` will end up getting called before the cursor message is added. I made some changes and will let you take it from here.

AlekSimpson:
All changes have been made and tests have been updated. All tests pass.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,132 @@
...
+
+        @Override
+        public J.Binary visitBinary(J.Binary binary, ExecutionContext ctx) {
+            Expression left = binary.getLeft();
+            Expression right = binary.getRight();
+
+            Cursor c = getCursor().dropParentWhile(is -> is instanceof J.Parentheses || !(is instanceof Tree));
+            if (c.getValue() instanceof J.MethodInvocation) {
+                if (PRINT_MATCHER.matches((J.MethodInvocation) c.getValue())) {
+                    if (left.getType() instanceof JavaType.Array || right.getType() instanceof JavaType.Array) {
```
knutwannheden, line 123: There is no need to constrain this to method invocations. So you can drop the two outer `if`s here and on this innermost `if` I would first also check the condition `binary.getOperator() == J.Binary.Type.Addition`.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+
+            return super.visitMethodInvocation(mi, ctx);
+        }
+
+        @Override
+        public Expression visitExpression(Expression exp, ExecutionContext ctx) {
+            Expression e = (Expression) super.visitExpression(exp, ctx);
+            if (e.getType() instanceof JavaType.Array) {
```
knutwannheden, line 101: When you add a test like `System.out.println("" + (arr))` (where `arr` is an array), you will notice that you get a runtime exception. That is because `J.Parentheses` doesn't extend `TypedTree` (and I currently can't tell you if it should or not). To fix this my suggestion is to simply add an `instanceof` check:

```suggestion
            if (e instanceof TypedTree && e.getType() instanceof JavaType.Array) {
```

knutwannheden (commented):
Almost there! I added a few more comments which should be addressed.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class RemoveToStringCallsFromArrayInstances extends Recipe {
+    private static final MethodMatcher OBJECT_TOSTRING_MATCHER = new MethodMatcher("java.lang.Object toString()");
+    private static final MethodMatcher OBJECTS_TOSTRING_MATCHER = new MethodMatcher("java.lang.Objects toString()");
```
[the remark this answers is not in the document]
AlekSimpson: The latest changes should address this now.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -98,7 +98,7 @@ public J visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
         @Override
         public Expression visitExpression(Expression exp, ExecutionContext ctx) {
             Expression e = (Expression) super.visitExpression(exp, ctx);
-            if (e.getType() instanceof JavaType.Array) {
+            if (e instanceof TypedTree && e.getType() instanceof JavaType.Array) {
```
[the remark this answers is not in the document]
knutwannheden: We could replace `e.getType() instanceof JavaType.Array` with `TypeUtils.asArray(e.getType()) != null` but I don't think it is really much better. IMHO the `instanceof` here is fine.

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -117,13 +127,8 @@ public J.Binary visitBinary(J.Binary binary, ExecutionContext ctx) {
...
             Expression right = binary.getRight();
 
-            Cursor c = getCursor().dropParentWhile(is -> is instanceof J.Parentheses || !(is instanceof Tree));
-            if (c.getValue() instanceof J.MethodInvocation) {
-                if (PRINT_MATCHER.matches((J.MethodInvocation) c.getValue())) {
-                    if (left.getType() instanceof JavaType.Array || right.getType() instanceof JavaType.Array) {
-                        getCursor().putMessage("BINARY_FOUND", binary);
-                    }
-                }
+            if (binary.getOperator() == J.Binary.Type.Addition || left.getType() instanceof JavaType.Array || right.getType() instanceof JavaType.Array) {
```
knutwannheden, line 130: The binary must be using `+`:
```suggestion
            if (binary.getOperator() == J.Binary.Type.Addition && (left.getType() instanceof JavaType.Array || right.getType() instanceof JavaType.Array)) {
```

knutwannheden (commented):
Apart from one last minor comment I think this looks good.

timtebeek (approved):

PR #128, opened 2023-06-22
Remove hashCode calls on arrays

AlekSimpson:
## What's changed?
This is related to [my other PR](https://github.com/openrewrite/rewrite-static-analysis/pull/126). @timtebeek suggested to me that I split the recipe because the `toString()` part required more checks since `toString()` is such a widely used method. This recipe basically does the same thing but on a smaller scale for `hashCode()` calls on array instances. 
 
## What's your motivation?
[#44](https://github.com/openrewrite/rewrite-static-analysis/issues/44)

I think this recipe should be finished but maybe there are other edge case methods that implicitly call `hashCode()` I'm forgetting. 

@timtebeek 

## Any additional context
[RSPEC-2116](https://rules.sonarsource.com/java/RSPEC-2116)

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.staticanalysis;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+
+public class RemoveHashCodeCallsFromArrayInstances extends Recipe {
+    private static final MethodMatcher HASHCODE_MATCHER = new MethodMatcher("java.lang.Object hashCode()");
+    @Override
+    public String getDisplayName() {
+        return "hashCode should not be called on array instances";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Removes hashCode method calls and replaces it with Arrays.hashCode(...) call.";
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static class RemoveHashCodeCallsFromArrayInstancesVisitor extends JavaIsoVisitor<ExecutionContext> {
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            J.MethodInvocation m = super.visitMethodInvocation(mi, ctx);
+
+            if (HASHCODE_MATCHER.matches(m)) {
+                String builder_string = "Arrays.hashCode(#{anyArray(java.lang.Object)})";
+                Expression select = m.getSelect();
+                assert select != null;
```
timtebeek, line 57: I don't think there's any hard guarantees these assert statements are actually executed. Perhaps an if statement would be better?

src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final MethodMatcher HASHCODE_MATCHER = new MethodMatcher("java.lang.Object hashCode()");
+    @Override
+    public String getDisplayName() {
+        return "`hashCode()` should not be called on array instances";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Removes `hashCode()` calls on arrays and replaces it with `Arrays.hashCode()` because the results from `hashCode()`" +
+                " are largely useless.";
```
timtebeek, line 40: Hadn't expected to read largely useless here, but I'll allow it off we don't have

src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
```diff
@@ -0,0 +1,75 @@
...
+                    return m;
+                }
+
+                J.MethodInvocation invocation = JavaTemplate.builder(builder_string)
+                        .imports("java.util.Arrays")
+                        .build()
+                        .apply(getCursor(), m.getCoordinates().replace(), select);
+                maybeAddImport("java.util.Arrays");
+
+                return invocation;
```
timtebeek, line 69: We can inline the variable here and it should all still work, for slightly less code to read and review

src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
AlekSimpson: Oh ok, ya IntelliJ just auto suggested using an assert statement. I'll change that.

src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
AlekSimpson: Ya there is an RSPEC tag that can be included. I'll add it and fix the description.

src/main/java/org/openrewrite/staticanalysis/RemoveHashCodeCallsFromArrayInstances.java
AlekSimpson: Oh ok, I thought the `maybeAddImport()` had to be after the builder statement for some reason.

timtebeek (approved):

timtebeek:
Merged after some small clean ups in https://github.com/openrewrite/rewrite-static-analysis/pull/128/commits/1d9871f746b842f54f1755b2a3eec635cf96ae49 ; thanks again!

Issue #58, opened 2022-04-02
RSPEC-3986: Week Year ("YYYY") should not be used for date formatting


yeikel:
Source : https://rules.sonarsource.com/java/RSPEC-3986/

timtebeek:
Hope you don't mind I unassigned you @vishalsingh2972 
I want to open this issue back up again to anyone to pick up.
Feel free to open up a draft pull request if you'd like to contribute!

PR #129, opened 2023-06-28
Replace week year (YYYY) with year (yyyy) in date formats

AlekSimpson:
## What's changed?
This PR introduces a new recipe to replace date formats that use week year (YYYY) with the year format (yyyy).

## What's your motivation?
[#58](https://github.com/openrewrite/rewrite-static-analysis/issues/58)

I got the first test to pass but the second one I found a little more tricky. I tried using a `visitMethodInvocation` visitor for it. However, the `.format()` invocation was being found instead of the `ofPattern` call. I tried getting the select of the method which did give me the format invocation but getting the visitor to account for the case when only `ofPattern` is called was tough. Do you think this is the right approach for this test? Maybe I should try using a visitor like `visitLiteral` and then checking if the literal is passed into an `ofPattern` invocation instead? 

@timtebeek 

## Any additional context
RSPEC-3986

### Checklist


src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,77 @@
...
+import org.openrewrite.marker.Markers;
+
+import java.util.Collections;
+import java.util.Set;
+
+import static org.openrewrite.Tree.randomId;
+
+public class ReplaceWeekYearWithYear extends Recipe {
+    private static final MethodMatcher DATE_FORMAT_MATCHER = new MethodMatcher("java.text.SimpleDateFormat <constructor>(..)");
+    private static final MethodMatcher OF_PATTERN_MATCHER = new MethodMatcher("java.time.format.DateTimeFormatter ofPattern(..)");
```
timtebeek, line 19: This matcher is currently unused, and the associated test is failing; what's your plan there?

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,77 @@
...
+import org.openrewrite.java.tree.*;
+import org.openrewrite.marker.Markers;
+
+import java.util.Collections;
+import java.util.Set;
+
+import static org.openrewrite.Tree.randomId;
+
+public class ReplaceWeekYearWithYear extends Recipe {
+    private static final MethodMatcher DATE_FORMAT_MATCHER = new MethodMatcher("java.text.SimpleDateFormat <constructor>(..)");
```
timtebeek, line 18: Note that this will match to three or four different constructors! You'd need to handle each of those in the recipe below.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,77 @@
...
+                            Markers.EMPTY,
+                            formatString,
+                            "\"" + formatString + "\"",
+                            null,
+                            JavaType.Primitive.String
+                    );
+
+                    System.out.println("getting here");
+
+                    return JavaTemplate.builder("new SimpleDateFormat(#{any(java.lang.String)})")
```
timtebeek, line 67: Related to the matcher comment above, you're at risk of losing any additional `DateFormatSymbols` or `Locale` argument here. Best to handle and test those as well.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,77 @@
...
+                            .contextSensitive()
```
timtebeek, line 68: I'm not entirely sure about the contextSensitive here; what made you add that?

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,77 @@
...
+                            randomId(),
+                            Space.EMPTY,
+                            Markers.EMPTY,
+                            formatString,
+                            "\"" + formatString + "\"",
+                            null,
+                            JavaType.Primitive.String
+                    );
+
+                    System.out.println("getting here");
```
timtebeek, line 65:
```suggestion
```

src/test/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYearTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.Issue;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+@SuppressWarnings("SuspiciousDateFormat")
+public class ReplaceWeekYearWithYearTest implements RewriteTest {
```
timtebeek, line 12: There's no need for public test classes or methods; best to remove those to avoid any confusion about whether the classes or methods are used elsewhere.
```suggestion
class ReplaceWeekYearWithYearTest implements RewriteTest {
```

src/test/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYearTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .recipe(new ReplaceWeekYearWithYear());
+    }
+
+    @Test
+    @DocumentExample
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/58")
+    public void changeSimpleDateFormat() {
```
timtebeek, line 22:
```suggestion
    void changeSimpleDateFormat() {
```

src/test/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYearTest.java
```diff
@@ -0,0 +1,82 @@
...
+                  String result = new SimpleDateFormat("yyyy/MM/dd").format(date);
+                }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    public void worksWithOfPatternFormatter() {
```
timtebeek, line 53:
```suggestion
    void worksWithOfPatternFormatter() {
```

timtebeek (dismissed):
Some initial feedback already. Needs a bit more work I think, but good start.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,76 @@
...
+            @Override
+            public J.NewClass visitNewClass(J.NewClass nc, ExecutionContext ctx) {
+                if (!DATE_FORMAT_MATCHER.matches(nc.getConstructorType())) {
+                    return nc;
+                }
+
+                Expression argument = nc.getArguments().get(0);
+                String formatString = argument.print(getCursor());
+
+                if (formatString.equals("\"YYYY/MM/dd\"")) {
```
timtebeek, line 49: Aren't we unnecessarily limiting the recipe to this particular format only?
I'd think it'd be just as valid when using `"dd/MM/YYYY"` or even `"dd/MM/YY"`. 🤔

AlekSimpson: Hmm yes that's a good point, I'll keep this mind.

AlekSimpson:
I added a couple more tests to test cases like "YY" being used in format patterns. I think this `visitLiteral` visitor should be able to cover all cases where a date is being formatted.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
AlekSimpson: I had it there at the time because I knew I would need it but wasn't sure yet how I would use it.

AlekSimpson:
> Added [167ef5e](https://github.com/openrewrite/rewrite-static-analysis/commit/167ef5e80c99fcfa8b12c77e64e5aea032817d90) to demonstrate an issue with the current implementation; you're replacing `YY` in _any_ string, not just those used as date format. We'd need that restriction before we can consider a merge.

I thought the preconditions would stop it from running on the wrong strings. I ran across code samples where the string pattern would often be in just a standalone string variable. Other using preconditions to make sure the Date libraries are being used, I could also maybe add some extra logic to check for a larger date pattern after finding `YY`?

timtebeek:
> I thought the preconditions would stop it from running on the wrong strings. I ran across code samples where the string pattern would often be in just a standalone string variables. Other using preconditions to make sure the Date libraries are being used, I could also maybe add some extra logic to check for a larger date pattern after finding `YY`?

The preconditions determine whether to run on a compilation unit (typically file?) I believe; not on individual Strings.

I think we'd want to use the cursor here to find if there's a wrapping method invocation around the String, and if that matches any of the methods we cover here. So start with `getCursor().find...` and have autocompletion guide you from there, also looking at other similar uses in other recipes. Does that help?

AlekSimpson:
I couldn't find any method on `getCursor()` that started with `.find...`, nothing was coming up in the autocomplete. However, I did realize that same thing could be achieved with cursor messaging by leaving a message on the methods/constructors that take in the string literals we are looking for.

timtebeek:
Noticed there's now two failing tests; are you planning to pick those up and should I wait with a review?
```
ReplaceWeekYearWithYearTest > datePatternIsVariable() FAILED
    java.lang.AssertionError: Expected recipe to complete in 1 cycle, but took 0 cycle.
        at org.openrewrite.staticanalysis.ReplaceWeekYearWithYearTest.datePatternIsVariable(ReplaceWeekYearWithYearTest.java:153)

ReplaceWeekYearWithYearTest > changeSimpleDateFormat() FAILED
    java.lang.AssertionError: Expected recipe to complete in 1 cycle, but took 0 cycle.
        at org.openrewrite.staticanalysis.ReplaceWeekYearWithYearTest.changeSimpleDateFormat(ReplaceWeekYearWithYearTest.java:39)
```

AlekSimpson:
> ngeSimpleDateFor

Yes I will be picking those up. I was kind of stuck on them yesterday but now I have a new idea that maybe will help.

timtebeek:
We do have some data flow analysis support that I've not used before myself; you can try to look for examples or ask Jonathan Leitschuh, although it's also fine to for now leave out cases where the format strings are in variables or constants, and only focus on direct use of literals going into the methods identified here.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,95 @@
...
+            return super.visitNewClass(nc, ctx);
+        }
+
+        @Override
+        public J.Literal visitLiteral(J.Literal li, ExecutionContext ctx) {
+            if (li.getValue() instanceof String) {
+                Cursor c = getCursor().dropParentWhile(is -> is instanceof J.Parentheses || !(is instanceof Tree));
+                if (c.getMessage("KEY") != null) {
+                    String value = li.getValueSource();
+                    if (value != null && value.contains("YY")) {
```
timtebeek, line 85: Why the check specifically here for at least `YY`? I'm reading [the specification](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) and it seems to allow for single `Y` as well.. 🤔 

> For parsing with the abbreviated year pattern ("y" or "yy"), SimpleDateFormat must interpret the abbreviated year relative to some century. It does this by adjusting dates to be within 80 years before and 20 years after the time the SimpleDateFormat instance is created. For example, using a pattern of "MM/dd/yy" and a SimpleDateFormat instance created on Jan 1, 1997, the string "01/11/12" would be interpreted as Jan 11, 2012 while the string "05/04/64" would be interpreted as May 4, 1964. During parsing, only strings consisting of exactly two digits, as defined by [Character.isDigit(char)](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isDigit-char-), will be parsed into the default century. Any other numeric string, such as a one digit string, a three or more digit string, or a two digit string that isn't all digits (for example, "-1"), is interpreted literally. So "01/02/3" or "01/02/003" are parsed, using the same pattern, as Jan 2, 3 AD. Likewise, "01/02/-3" is parsed as Jan 2, 4 BC.
Otherwise, calendar system specific forms are applied. For both formatting and parsing, if the number of pattern letters is 4 or more, a calendar specific [long form](https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.html#LONG) is used. Otherwise, a calendar specific [short or abbreviated form](https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.html#SHORT) is used.

timtebeek: And we might need to guard against the (quite hypothetical) use single quotes
> Text can be quoted using single quotes (') to avoid interpretation. "''" represents a single quote. All other characters are not interpreted; they're simply copied into the output string during formatting or matched against the input string during parsing.

AlekSimpson: Ok the latest commits should address these issues

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,120 @@
...
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.*;
+
+import java.util.*;
+
+public class ReplaceWeekYearWithYear extends Recipe {
+    public static final MethodMatcher SIMPLE_DATE_FORMAT_CONSTRUCTOR_MATCHER = new MethodMatcher("java.text.SimpleDateFormat <constructor>(..)");
+    public static final MethodMatcher OF_PATTERN_MATCHER = new MethodMatcher("java.time.format.DateTimeFormatter ofPattern(..)");
```
knutwannheden, line 28: Detail: I suggest declaring these constants as private. API surface area is always easy to extend, but more difficult to shrink.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,120 @@
...
+                if (c.getMessage("KEY") != null) {
+                    String value = li.getValueSource();
+
+                    if (value == null) {
+                        return li;
+                    }
+
+                    String newValue = replaceY(value);
+
+                    return li.withValueSource(newValue).withValue(newValue);
```
knutwannheden, line 90: This looks a bit wrong to me. The `valueSource` will be the value with quotes (possibly even Java 17 block quotes, for which we should have a test probably), while the `value` is without quotes. We should update it correctly here.

src/main/java/org/openrewrite/staticanalysis/ReplaceWeekYearWithYear.java
```diff
@@ -0,0 +1,120 @@
...
+                } else if (currentChar == 'Y' && !insideQuotes) {
+                    output.append('y');
+                } else if (currentChar == 'Y' && nextChar == '\'') {
+                    output.append(currentChar);
+                } else {
+                    output.append(currentChar);
+                }
+            }
+
+            return output.toString();
```
knutwannheden, line 117: I suspect that when we do the `J.Literal#withValue()` call it may end up producing a new `J.Literal` object. So I think we should here in the loop update a `boolean` variable if there was any change made and only then return this result. Otherwise the input value to this method. Does that make sense to you?

AlekSimpson: Ah I see, ya I can add a check to see if there are any changes actually being made.

AlekSimpson:
Ok I've added another test case and made those last changes.

knutwannheden:
@AlekSimpson I did a `git merge origin/main` as the changes in this PR listed a lot of changes which had previously been merged to the main branch.

knutwannheden (approved):
LGTM 👍

PR #131, opened 2023-06-29
Better type attribution for `ReplaceLambdaWithMethodReference`

knutwannheden:
Improves the type attribution for method references by "manually" instantiating the `J.MemberReference` rather than via `JavaTemplate`. This may seem strange, but the reason is that the recipe doesn't know where the type is coming from (JDK, 3rd party library, or even source path), so the template's compiler may not have the type on its classpath and cannot really produce a properly type-attributed result.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/JavaElementFactory.java
+++ b/src/main/java/org/openrewrite/staticanalysis/JavaElementFactory.java
@@ -32,13 +32,19 @@ final class JavaElementFactory {
         return newInstanceMethodReference(method, containing, type);
     }
 
-    static Expression className(JavaType.FullyQualified classType, boolean qualified) {
+    static Expression className(JavaType type, boolean qualified) {
         Expression name = null;
-        String qualifiedName = qualified ? classType.getFullyQualifiedName() : classType.getClassName();
+        String qualifiedName;
+        if (type instanceof JavaType.FullyQualified) {
+            qualifiedName = qualified ? ((JavaType.FullyQualified) type).getFullyQualifiedName() : ((JavaType.FullyQualified) type).getClassName();
+        } else {
+            qualifiedName = type.toString();
+        }
+
         Scanner scanner = new Scanner(qualifiedName.replace('$', '.')).useDelimiter("\\.");
         for (int i = 0; scanner.hasNext(); i++) {
             String part = scanner.next();
-            JavaType typeOfContaining = scanner.hasNext() ? null : classType;
+            JavaType typeOfContaining = scanner.hasNext() ? null : type;
             if (i > 0) {
                 name = new J.FieldAccess(
                         randomId(),
@@ -49,7 +55,7 @@ final class JavaElementFactory {
                         typeOfContaining
                 );
             } else {
-                name = new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, part, classType, null);
+                name = new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, part, type, null);
             }
         }
         assert name != null;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/staticanalysis/JavaElementFactory.java
```diff
@@ -0,0 +1,123 @@
...
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.tree.*;
+import org.openrewrite.marker.Markers;
+
+import java.util.Collections;
+import java.util.Scanner;
+
+import static org.openrewrite.Tree.randomId;
+
+final class JavaElementFactory {
```
knutwannheden, line 27: Please note that this class was intentionally not made public. I don't know how we intend to address this kind of problem in the long run (creating LST elements which potentially refer to types from the source set). While I think it would be nice if `JavaTemplate` could be used for this, I have my doubts if we will be able to provide this any time soon. So in the meantime I suggest that we have a factory like this one.

src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java
```diff
@@ -79,45 +77,46 @@ public J visitLambda(J.Lambda lambda, ExecutionContext executionContext) {
...
-                            .contextSensitive()
-                            .javaParser(JavaParser.fromJavaVersion().dependsOn(stub))
-                            .imports(fullyQualified == null ? "" : fullyQualified.getFullyQualifiedName()).build();
-                    return template.apply(getCursor(), l.getCoordinates().replace(), identifier.getSimpleName());
-                } else if (body instanceof J.Binary) {
+                if (body instanceof J.Binary) {
                     J.Binary binary = (J.Binary) body;
                     if (isNullCheck(binary.getLeft(), binary.getRight()) ||
                         isNullCheck(binary.getRight(), binary.getLeft())) {
-                        maybeAddImport("java.util.Objects");
```
knutwannheden, line 115: This is basically how it all got started: I wanted to use the new `ShortenFullyQualifiedTypeReferences` recipe to allow removing this line (to avoid conflicts with any other imports for types with `Objects` as the simple name). That is how I discovered the incorrect type attribution with the stub from a few lines above.

joanvr (approved):

src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java
```diff
@@ -79,45 +77,46 @@ public J visitLambda(J.Lambda lambda, ExecutionContext executionContext) {
...
                 } else if (body instanceof J.TypeCast) {
                     if (!(((J.TypeCast) body).getExpression() instanceof J.MethodInvocation)) {
                         J.ControlParentheses<TypeTree> j = ((J.TypeCast) body).getClazz();
-                        if (j != null) {
-                            @SuppressWarnings("rawtypes") J tree = ((J.ControlParentheses) j).getTree();
-                            if (tree instanceof J.Identifier &&
-                                !(j.getType() instanceof JavaType.GenericTypeVariable)) {
-                                body = tree;
-                                code = "#{}.class::cast";
+                        J tree = j.getTree();
```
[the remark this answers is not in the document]
knutwannheden: `TypeCast#getClazz()` is not annotated as `@Nullable`, so I think this should be fine.

src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java
```diff
@@ -79,45 +77,46 @@ public J visitLambda(J.Lambda lambda, ExecutionContext executionContext) {
...
+                if (body instanceof J.Binary) {
                     J.Binary binary = (J.Binary) body;
                     if (isNullCheck(binary.getLeft(), binary.getRight()) ||
                         isNullCheck(binary.getRight(), binary.getLeft())) {
-                        maybeAddImport("java.util.Objects");
-                        code = J.Binary.Type.Equal.equals(binary.getOperator()) ? "Objects::isNull" :
-                                "Objects::nonNull";
+                        doAfterVisit(new ShortenFullyQualifiedTypeReferences().getVisitor());
+                        code = J.Binary.Type.Equal.equals(binary.getOperator()) ? "java.util.Objects::isNull" :
+                                "java.util.Objects::nonNull";
```
knutwannheden, line 117: OpenRewrite doesn't provide any API to do a lookup of types. That would sometimes be useful to have.

PR #132, opened 2023-06-29
`ReplaceLambdaWithMethodReference` should not replace lambda supplier of method reference with method reference

rickie:
## What's changed?

While applying the `CommonStaticAnalysis` [recipe](https://docs.openrewrite.org/recipes/staticanalysis/commonstaticanalysis) on our internal codebase, we found a false positive that results in non-compilable code.

This PR introduces a minimal reproduction case to show the problem. Probably the test case can be simplified even further.
If anything is unclear, let me know.

## Any additional context

Depending on what the fix would be, I'm interested in contributing 😄. 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
@@ -1184,5 +1184,27 @@ class ReplaceLambdaWithMethodReferenceTest implements RewriteTest {
         );
     }
 
+    @Test
+    void orElseGet() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import java.util.Optional;
+              import java.util.function.Function;
 
+              class A {
+                public void testCase() {
+                  Function<String, Integer> function = str -> 1;
+                  Optional.of(function).orElseGet(() -> this::foo);
+                }
+
+                private Integer foo(String bar) {
+                  return 1;
+                }
+              }
+              """
+          )
+        );
+    }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
```diff
@@ -1184,5 +1184,27 @@ void m() {
...
+          //language=java
+          java(
+            """
+              import java.util.Optional;
+              import java.util.function.Function;
+
+              class A {
+                public void testCase() {
+                  Function<String, Integer> function = str -> 1;
+                  Optional.of(function).orElseGet(() -> this::foo);
```
timtebeek, line 1199: So after applying the recipe we incorrectly end up with
```diff
-Optional.of(function).orElseGet(() -> this::foo);
+Optional.of(function).orElseGet(this::foo);
```
Whereas we should not have made a change if I understand you correctly.

src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
```diff
@@ -1184,5 +1184,27 @@ void m() {
         );
     }
 
+    @Test
+    void orElseGet() {
```
timtebeek, line 1188:
```suggestion
    @Issue("https://github.com/openrewrite/rewrite-static-analysis/pull/132")
    void orElseGet() {
```

rickie:
Thanks for the fix and applying the feedback @timtebeek. 

Added a commit to update the name of the test method 😄.

timtebeek:
Thanks again for reporting this; merged in the expectation that an additional restriction before we apply code changes is safe.

PR #134, opened 2023-07-05
Context-free template for `AddSerialVersionUidToSerializable`

knutwannheden:


knutwannheden:
Integrating this will require a new release of OpenRewrite, so I keep it open for now.

timtebeek:
Would we want to enable this again as part of this change?
https://github.com/openrewrite/rewrite-static-analysis/blob/ea029b4c95a991a481da969535b4228bd0d8bea7/src/main/resources/META-INF/rewrite/common-static-analysis.yml#L22

knutwannheden:
Possibly, yes. We should do some profiling to make sure that performance looks good now. I just want to keep this PR open until we do the next OpenRewrite release, because otherwise it won't work. (Should we add a specific label for this maybe?).

timtebeek (approved):

Issue #121, opened 2023-06-09
FinalizePrivateFields must not change volatile fields


Bananeweizen:
## What is the smallest, simplest way to reproduce the problem?

```java
public final class Reproducer {

    private Reproducer() {
    }

    private static volatile String foo = "this becomes final volatile, which is invalid";

}
```

## What did you expect to see?
same code, unmodified

## What did you see instead?

```java
public final class Reproducer {

    private Reproducer() {
    }

    private static final volatile String foo = "this becomes final volatile, which is invalid";

}
```

Notice the added `final`. A field cannot be both final and volatile, only either or.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes, I'll try to create a test and fix.

timtebeek:
Thank you very much for the offer to help! I imagine you're comfortable enough by now to work on a potential fix here without further guidance? It'll start with [adding a unit test](https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/test/java/org/openrewrite/staticanalysis/FinalizePrivateFieldsTest.java), and then likely stepping through with the debugger and potentially updating this filter step. https://github.com/openrewrite/rewrite-static-analysis/blob/fd5c7dc62b25d9831d6798b2b2a5f1a65e1e1233/src/main/java/org/openrewrite/staticanalysis/FinalizePrivateFields.java#L124

MehmetTerc:
I dont understand the Issue, can you describe me the task @Bananeweizen

timtebeek:
> I dont understand the Issue, can you describe me the task

It sounds like @Bananeweizen was already planning to pick this up @MehmetTerc ; there are some [other open issues](https://github.com/orgs/openrewrite/projects/4/views/10?filterQuery=label%3A%22good+first+issue%22) if you're looking to pick up something else.

In this particular case we want to limit the execution of `FinalizePrivateFields` such that it does not update any `volatile` fields, possibly [as described here](https://github.com/openrewrite/rewrite-static-analysis/issues/121#issuecomment-1584711158).

PR #137, opened 2023-07-18
Bug Fix for #121

AlekSimpson:
## What's changed?
This PR fixes a bug having to do with the FinalizePrivateFields recipe adding the `final` flag to variables that also have the `volatile` flag which is invalid. 

## What's your motivation?
[#121](https://github.com/openrewrite/rewrite-static-analysis/issues/121)

@timtebeek 

### Checklist
- [ ] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/FinalizePrivateFields.java
+++ b/src/main/java/org/openrewrite/staticanalysis/FinalizePrivateFields.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.staticanalysis;
 
+import fj.P;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.*;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/FinalizePrivateFields.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.staticanalysis;
 
+import fj.P;
```
sambsnyd, line 18: This looks unused

AlekSimpson: Ah good catch

sambsnyd:
Thanks @AlekSimpson !

PR #142, opened 2023-07-22
Do not use String `equals` for Groovy either

timtebeek:
## What's changed?
Add a `GroovyFileChecker` as precondition to `StringLiteralEquality`.

## What's your motivation?
Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/140

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/groovy/GroovyFileChecker.java
+++ b/src/main/java/org/openrewrite/staticanalysis/groovy/GroovyFileChecker.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.staticanalysis.groovy;
+
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.groovy.tree.G;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.marker.SearchResult;
+
+/**
+ * Add a search marker if vising a Groovy file
+ */
+public class GroovyFileChecker<P> extends TreeVisitor<Tree, P> {
+    @Override
+    public @Nullable Tree visit(@Nullable Tree tree, P p) {
+        if (tree instanceof G.CompilationUnit) {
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/staticanalysis/groovy/StringLiteralEqualityTest.java
```diff
@@ -0,0 +1,46 @@
...
+class StringLiteralEqualityTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new StringLiteralEquality());
+    }
+
+    // Don't change for Groovy because in Groovy, `==` means structural equality and it's redundant to call equals().
+    @Test
+    void doNotChangeForGroovy() {
```
timtebeek, line 34: We have a similar separate class to test Kotlin; Was wondering if we shouldn't just collapse these into `@Nested` test classes in `StringLiteralEqualityTest` in package `org.openrewrite.staticanalysis`.

src/main/java/org/openrewrite/staticanalysis/groovy/GroovyFileChecker.java
```diff
@@ -0,0 +1,35 @@
...
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.groovy.tree.G;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.marker.SearchResult;
+
+/**
+ * Add a search marker if vising a Groovy file
+ */
+public class GroovyFileChecker<P> extends TreeVisitor<Tree, P> {
```
shanman190, line 27: We could probably make this a bit more generic, so that it can cover even more ground. 

```
@Value
public class IsInstanceOf<P> extends TreeVisitor<Tree, P> {
    Class<? extends Tree> treeType;

    @Nullable
    public Tree visit(@Nullable Tree tree, P p) {
        if (tree != null && treeType.isInstance(tree.getClass()) {
            return SearchResult.found(tree);
        }
        return tree;
    }
}
```

Then this may be a candidate for inclusion down in `rewrite-core`.

timtebeek: I quite like this suggestion, thanks! Especially eyeing extended support coming up for Groovy, Kotlin, Typescript, Python, ... would help to handle this once in /rewrite and not repeat ourselves per module with such checks.

timtebeek: @kunli2 Since you're a little more involved with new language development and these exclusions, where would you see this fit best? And would you be willing to take this on?

joanvr (approved):
looks good to me. Aside from the more generic approach suggested by @shanman190 , I think it actually fixes the purpose of the PR. So, maybe we can merge this one and iterate over it when we have capacity

src/main/java/org/openrewrite/staticanalysis/groovy/GroovyFileChecker.java
timtebeek: Logged as an improvement here, since I'd like to see this fix roll out before we get a response & rework through.
https://github.com/openrewrite/rewrite-static-analysis/issues/144
Thanks for the suggestion!

PR #146, opened 2023-08-02
MinimumSwitchCases should add Enum import when necessary

timtebeek:
## What's changed?
Add import for enum when not present already.

## What's your motivation?
Code otherwise will not compile.

Somehow the import is not added yet. Not sure yet why.


knutwannheden (approved):
👍

PR #153, opened 2023-08-10
RemoveUnusedLocalVariables fails inside case blocks (test)

### LINKED ISSUE #152: RemoveUnusedLocalVariables fails inside case blocks

I am using

- rewrite-java:8.2.0-SNAPSHOT

Inside the rewrite-static-analysis repo; as a test

```java
class Test {
    static void method() {
        int x = 10;
        char y = 20;
        switch (x) {
            case 10:
                byte unused;
                break;
        }
    }
}
```

results in

```java
class Test {
    static void method() {
        int x = 10;
        char y = 20;
        switch (x) {
            case 10:
                byte; // <-- mistake here
                break;
        }
    }
}
```

sfc-gh-dtran:
## What's your motivation?

Fixes #152 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedLocalVariables.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedLocalVariables.java
@@ -114,6 +114,8 @@ public class RemoveUnusedLocalVariables extends Recipe {
                     parent instanceof J.MethodDeclaration ||
                     // skip if defined in an enhanced or standard for loop, since there isn't much we can do about the semantics at that point
                     parent instanceof J.ForLoop.Control || parent instanceof J.ForEachLoop.Control ||
+                    // skip if defined in a switch case
+                    parent instanceof J.Case ||
                     // skip if defined in a try's catch clause as an Exception variable declaration
                     parent instanceof J.Try.Resource || parent instanceof J.Try.Catch || parent instanceof J.MultiCatch ||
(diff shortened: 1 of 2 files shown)


timtebeek:
Appreciate you adding this test! There's a bit of an oddity where you also remove two other tests. Would you mind reverting that?

sfc-gh-dtran:
> Appreciate you adding this test! There's a bit of an oddity where you also remove two other tests. Would you mind reverting that?

Thanks, that was weird indeed. I re-added those missing tests.

timtebeek:
Thanks! I've ran your test and briefly explored the options; the quickest thing we can do now to avoid making unwanted changes, is to simply skip unused variables defined in case statements. That's a bit unfortunate, but better than making the wrong change right now. We can then later look to also remove unused local variables inside case statements. Hope you agree!

sfc-gh-dtran:
> Thanks! I've ran your test and briefly explored the options; the quickest thing we can do now to avoid making unwanted changes, is to simply skip unused variables defined in case statements. That's a bit unfortunate, but better than making the wrong change right now. We can then later look to also remove unused local variables inside case statements. Hope you agree!

Thanks! I've actually already looked into this and it looks like the problem is with the `DeleteStatement` recipe not handling statements inside case blocks: adding a simple `visitCaseBlock` visitor fixed this; I can make a PR to `rewrite` so you can take a look at it

PR #154, opened 2023-08-18
Simplify null and empty string comparisons

timtebeek:
## What's changed?
Also simplify comparisons such as `"" == null`, `"" != null` and `"".isEmpty()` such that we can clean up the replacements inserted in https://github.com/openrewrite/rewrite-migrate-java/pull/271 when parameters are using nullable literals.

## What's your motivation?
- https://github.com/openrewrite/rewrite-templating/issues/28
- https://github.com/openrewrite/rewrite-migrate-java/pull/271

- [ ] Should we do anything more to prevent issues with Kotlin nullable?


timtebeek:
Replaced with https://github.com/openrewrite/rewrite/pull/3489; we should likely see if we can have the recipe here use the visitor in rewrite-java, such that we don't maintain these two in parallel. Might require an extensibility hook in the separate visitor to allow adding in the Kotlin method comparisons here.

PR #161, opened 2023-08-24
do not remove try block if there are resources

joanvr:
## What's changed?
Try with resources can have side effects (closing the streams) even with empty try blocks. So we cannot delete it safely.

Modified the emptyTry test to not have resources (so it's removed because it's empty), and added a new emptyTryWithResources that is actually not removed because it has resources.

@kmccarp 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EmptyBlockVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EmptyBlockVisitor.java
@@ -18,16 +18,14 @@ package org.openrewrite.staticanalysis;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.DeleteStatement;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.format.ShiftFormat;
 import org.openrewrite.java.style.EmptyBlockStyle;
-import org.openrewrite.java.tree.Expression;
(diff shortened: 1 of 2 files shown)


kmccarp (approved):

timtebeek (approved):

src/main/java/org/openrewrite/staticanalysis/EmptyBlockVisitor.java
```diff
@@ -223,6 +221,29 @@ private boolean isEmptyBlock(Statement blockNode) {
...
+                J.VariableDeclarations variableDeclarations = (J.VariableDeclarations) resource.getVariableDeclarations();
+                for (J.VariableDeclarations.NamedVariable variable : variableDeclarations.getVariables()) {
+                    // If the variable is not initialized with a new instance, it means it can come from outside the scope.
+                    if (!(variable.getInitializer() instanceof J.NewClass)) {
+                        return false;
+                    }
+                }
+            }
+        }
+        return true;
```
kmccarp, line 247: I'm not clear why this is necessary -- if resources is null or empty, that should be good enough. In the case I reported, there will be no other reference to them outside of the resources block because they're coming from an input variable to a lambda. I would rather us make fewer, safer changes, than make more changes and risk breaking.

joanvr: Without this, the existing test https://github.com/openrewrite/rewrite-static-analysis/blob/a58e9c52ad0ec5df6cd059517cd6200e034ebc0f/src/test/java/org/openrewrite/staticanalysis/EmptyBlockTest.java#L109-L134 was failing.

Actually, the code here is limiting the scope of changes, only removing the block on less scenarios than before, but keeping the one that on the tests we agreed that is valid.

Issue #162, opened 2023-08-24
Simplify lambda blocks to expressions replaces return that is used to infer which overloaded method to pick


pstreef:
When an overloaded method that takes either a consumer or Function the lambda can not always be rewritten from a block to an expression, because the return statement is implying that the Function argument is used.

<img width="774" alt="image" src="https://github.com/openrewrite/rewrite-static-analysis/assets/7645564/8a97fb78-91ee-42b9-ba6b-4ca1119fa9a6">

I've created a test that reproduces this issue, but it does not detect the compiler issue that I get when I use the output.

Note intellij also suggests this change.

kunli2:
fixed by https://github.com/openrewrite/rewrite-static-analysis/commit/3c3fdedf86e0c60782c591ac78e28346f3e4fcd5

PR #163, opened 2023-08-24
Simplify lambda blocks to expressions replaces return that is used to infer which overloaded method to pick

pstreef:
adding a test for https://github.com/openrewrite/rewrite-static-analysis/issues/162

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
@@ -15,22 +15,26 @@
  */
 package org.openrewrite.staticanalysis;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.Issue;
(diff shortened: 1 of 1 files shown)


pstreef:
Note that there are no compiler warnings or errors reported

timtebeek:
Could this be a variant of https://github.com/openrewrite/rewrite-static-analysis/issues/10 ?

timtebeek:
@kunli2 Merging this one was maybe a bit too soon; the test output here does not compile, and it should likely have been a test asserting no change was made.

kunli2:
@timtebeek , Yes, sorry about that, I found it's a folk here but not a branch, so I am not able to make a change on it here, so I planned to fix this issue right away after merging this. 
The fix is [here](https://github.com/openrewrite/rewrite-static-analysis/commit/3c3fdedf86e0c60782c591ac78e28346f3e4fcd5)

timtebeek:
Great thanks! Does this then also fix #10 ?

kunli2:
> Great thanks! Does this then also fix #10 ?

I don't think so, #10 looks like a different problem

pstreef:
@kunli2 thanks for this quick fix! 🎉

PR #167, opened 2023-09-15
Add recipe to rewrite java 11 `Stream.collect(Collectors.toUnmodifiab…

Laurens-W:
## What's changed?

Adding a recipe that rewrites Java 11 `Stream.collect(Collectors.toUnmodifiableList())` to Java 16 `Stream.toList()` with an optional toggle to include `Stream.collect(Collectors.toList())` so no harm is done initially.

## What's your motivation?

I asked on slack if there was an opposing recipe to replacestreamtolistwithcollect, we came to the conclusion that it does not exist but does seem justifiable to have.

@timtebeek @joanvr 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ReplaceCollectWithStreamToList.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ReplaceCollectWithStreamToList.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.staticanalysis;
+
+import java.time.Duration;
+import lombok.Value;
+import org.jetbrains.annotations.NotNull;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesJavaVersion;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.J.MethodInvocation;
+import org.openrewrite.java.tree.JRightPadded;
+
+@Value
+public class ReplaceCollectWithStreamToList extends Recipe {
+
+    private static final MethodMatcher STREAM_COLLECT_TO_LIST = new MethodMatcher("java.util.stream.Stream collect(java.util.stream.Collector)");
+    private static final MethodMatcher COLLECTORS_TO_LIST = new MethodMatcher("java.util.stream.Collectors toList()");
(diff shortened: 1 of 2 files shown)


timtebeek:
Hi @Laurens-W ! thanks for opening this pull request! :) Fun to see you get involved here.
I do wonder if we should perhaps target this recipe for inclusion here instead:
https://github.com/openrewrite/rewrite-migrate-java/blob/e8c30122672782861094ef366d495d7b8d35c3ec/src/main/java/org/openrewrite/java/migrate/util
There's a few more collection related recipes there, and it ties in with only upgrading when on a particular Java version or above. What do you think?

src/main/java/org/openrewrite/staticanalysis/ReplaceCollectWithStreamToList.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(1);
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(Preconditions.and(new UsesJavaVersion<>(16),
+                new UsesMethod<>(STREAM_COLLECT_TO_LIST)), new JavaVisitor<ExecutionContext>() {
```
timtebeek, line 68: Perhaps we need a `Preconditions.or()` between using either `Collectors toList` or `Collectors toUnmodifiableList`. Now we won't execute for cases that only use `toUnmodifiableList`.

timtebeek, line 68: Perhaps we can add additional preconditions for using either `COLLECTORS_TO_LIST` or `COLLECTORS_TO_UNMODIFIABLE_LIST`.

src/main/java/org/openrewrite/staticanalysis/ReplaceCollectWithStreamToList.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static final MethodMatcher STREAM_COLLECT_TO_LIST = new MethodMatcher("java.util.stream.Stream collect(java.util.stream.Collector)");
+    private static final MethodMatcher COLLECTORS_TO_LIST = new MethodMatcher("java.util.stream.Collectors toList()");
+    private static final MethodMatcher COLLECTORS_TO_UNMODIFIABLE_LIST = new MethodMatcher("java.util.stream.Collectors toUnmodifiableList()");
+
+    @Option(displayName = "Should the recipe also apply to mutable toList?",
+        description = "Also replace Java 11 `Stream.collect(Collectors.toList())` with Java 16 `Stream.toList()`.",
+        required = false)
+    @Nullable
+    Boolean includeMutable;
```
timtebeek, line 48: Not sure if we sufficiently explain the risk here; maybe we should state that there's a risk when converting from mutable to immutable collections, hence why it's an option only, and turned off by default.

src/test/java/org/openrewrite/staticanalysis/ReplaceCollectWithStreamToListTest.java
```diff
@@ -0,0 +1,188 @@
...
+              import java.util.stream.Stream;
+
+              class Example {
+                  public void test() {
+                      Stream.of().toList();
+                  }
+              }
+              """),
+            16));
+    }
```
timtebeek, line 62:
```suggestion
    @Test
    @DocumentExample
    void replacesToUnmodifiableList() {
        rewriteRun(
          version(
            //language=java
            java(
              """
              import java.util.stream.Collectors;
              import java.util.stream.Stream;

              class Example {
                  public List<String> test(Stream<String> stream) {
                      return stream.collect(Collectors.toUnmodifiableList());
                  }
              }
              """,
              """
              import java.util.stream.Stream;

              class Example {
                  public List<String> test(Stream<String> stream) {
                      return stream.toList();
                  }
              }
              """),
            16));
    }
```

src/main/java/org/openrewrite/staticanalysis/ReplaceCollectWithStreamToList.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public String getDescription() {
+        return "Replace Java 11 `Stream.collect(Collectors.toUnmodifiableList())` with Java 16 `Stream.toList()`.";
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(1);
+    }
+
```
timtebeek, line 64: Would be nice to link to the RSPEC here as well
```suggestion

    @Override
    public Set<String> getTags() {
        return new HashSet<>(Arrays.asList("RSPEC-6204"));
    }

```

PR #169, opened 2023-09-22
Do not simplify boolean return with comments

timtebeek:
## What's changed?
No longer simplify boolean returns if there are comments on the return statements.

## What's your motivation?
Comments might provide important context that can be lost when returns are simplified.


kunli2 (approved):

jbessels:
Good solution.

PR #170, opened 2023-09-23
Preserve arrays when inlining variables

timtebeek:
Before this PR we get a compiler error.
After this PR we retain the array for now.

A future PR might move the type declaration.


jbessels:
I've tested it and it works like a charm.

Issue #174, opened 2023-09-26
InstanceOfPatternMatch leads to duplicate variable name in if statement with two different instanceof


ghost:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v5.5.2
- Maven/Gradle plugin v3.11.0
- rewrite-_module_ v2.1.0

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>5.5.2</version>
    <configuration>
        <activeRecipes>
            <recipe>org.openrewrite.java.migrate.UpgradeToJava20</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>2.1.0</version>
        </dependency>
    </dependencies>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
if (test instanceof TestClass && test2 instanceof TestClass) {
    ...
}
```

## What did you expect to see?

```java
if (test instanceof TestClass t && test2 instanceof TestClass t2) {
    ...
}
```

## What did you see instead?

```java
if (test instanceof TestClass t && test2 instanceof TestClass t) {
    ...
}
```

timtebeek:
Replicated in https://github.com/openrewrite/rewrite-static-analysis/pull/175

timtebeek:
Already much improved in 
- https://github.com/openrewrite/rewrite-static-analysis/pull/175

Closing the issue as I expect any remaining cases to be rare in practice.

PR #175, opened 2023-09-27
Duplicate variable name in if statement with two different instanceof

KoTurk:
Added failing JUnit test for issue "Duplicate variable name in if statement with two different instanceof"

## What's changed?
Created a failing JUnit test for the issue.
TODO: fix the issue

## What's your motivation?
- https://github.com/openrewrite/rewrite-static-analysis/issues/174

Be kind, it's my first contribution to Open Source :-)

@timtebeek 

### Checklist


timtebeek:
As a next step it looks like we should be using VariableNameUtils when naming primitves as well; we already use it elsewhere, but not yet for primitives.
https://github.com/openrewrite/rewrite-static-analysis/blob/2fb2bfb0faf43e9f12bd1ce149e5e42ee14337a2/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java#L424-L426

timtebeek:
Stepped through this case with the debugger; the problem here seems to be that once we process the second `instanceof`, that we don't have the first `instanceof` new `string` variable in scope as we generate a name for the second one. 🤔

timtebeek:
Exploring a likely fix in
- https://github.com/openrewrite/rewrite/pull/4066

timtebeek (approved):
Thanks for adding the tests @KoTurk & exploring a fix. Was a more of a puzzle than anticipated, but this current iteration already improves things enough to warrant a merge. There's still a known case uncovered, but I'll stop here for today.

PR #183, opened 2023-10-02
Don't redeclare EnumSet as Set

### LINKED ISSUE #179: UseCollectionInterfaces breaks EnumSet iteration

I am using

- Maven/Gradle plugin v5.7.1

```java
        EnumSet<?> values = EnumSet.allOf(MyEnum.class);
        for (Enum<?> e : values) {
            // something
        }
```

no change

## What did you see instead?

```java
        Set<?> values = EnumSet.allOf(MyEnum.class);
        for (Enum<?> e : values) {
            // something
        }
```
That doesn't compile anymore. For other subclasses of Set, using the super class for the declaration doesn't change the template argument. For EnumSet it does, because EnumSet is defined with a template argument `Enum<E>` extending `Set<E>`. Therefore anx code needing then Enum typing will not compile anymore.

If some maintainer agrees, I would try to exclude EnumSet from the applicable types in this recipe.

Bananeweizen:
That changes the type of the generic, potentially breaking iterators.

Fixes #179

## What's changed?
Exclude EnumSet from the recipe.

I've not commented in the recipe but rather created a unit test to verify that EnumSet doesn't get back into the recipe on accident.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/UseCollectionInterfacesTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/UseCollectionInterfacesTest.java
@@ -688,12 +688,6 @@ class UseCollectionInterfacesTest implements RewriteTest {
     }
 
     @Test
-    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/179")
-    void enumSetNotInRecipe() {
-        assertThat(UseCollectionInterfaces.rspecRulesReplaceTypeMap).doesNotContainKey("java.util.EnumSet");
-    }
-
-    @Test
     void hashSet() {
         rewriteRun(
           //language=java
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/UseCollectionInterfacesTest.java
```diff
@@ -663,47 +664,51 @@ class Test {
...
           )
         );
     }
 
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/179")
+    void enumSetNotInRecipe() {
+        assertThat(UseCollectionInterfaces.rspecRulesReplaceTypeMap).doesNotContainKey("java.util.EnumSet");
+    }
+
```
timtebeek, line 695: Not sure we'd need this with the above test adapted already, right? Saves us from tying too strongly to implementation details (and having to maintain that) if we remove this one.
```suggestion
```

timtebeek (approved):

timtebeek:
Thanks again! Really helpful all these fixes left and right.

PR #184, opened 2023-10-02
Transform collection.size() < 1 to isEmpty()

Bananeweizen:
Not only size comparisons with zero can be optimized to isEmpty() calls. Additionally we can do (plus order changes):
* size() < 1 to isEmpty()
* size() >= 1 to !isEmpty()

## What's your motivation?
real life code

I've changed one of the existing unit tests. It was obviously meant to test `size() == 0` and `0 == size()` (to create all 8 variants of positive and negative tests).

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/IsEmptyCallOnCollections.java
+++ b/src/main/java/org/openrewrite/staticanalysis/IsEmptyCallOnCollections.java
@@ -68,14 +68,36 @@ public class IsEmptyCallOnCollections extends Recipe {
             public J visitBinary(J.Binary binary, ExecutionContext ctx) {
                 if (isZero(binary.getLeft()) || isZero(binary.getRight())) {
                     boolean zeroRight = isZero(binary.getRight());
-                    J maybeSizeCall = zeroRight ? binary.getLeft() : binary.getRight();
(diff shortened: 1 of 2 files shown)


Bananeweizen:
I know refaster (and also how horrible complicated it is to set it up in other tech stacks than what the Google people use). Even with a refaster template, this second template (comparing with 1) would be needed. My impression was that the refaster part is not yet production ready, or am I wrong there?

Stephan202:
(Sorry for going off-topic.)

> I know refaster (and also how horrible complicated it is to set it up in other tech stacks than what the Google people use).

OOTB that's very true. At Picnic we built a Refaster [compiler](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-compiler) and [runner](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-runner), such that Refaster templates are effectively exposed as a regular Error Prone check. (We should probably further improve the documentation, but all the pieces are in that repo.)

W.r.t. the rule being discussed: we have [this one](https://github.com/PicnicSupermarket/error-prone-support/blob/34fd69206466013d88bdf81a133f4b734678d89c/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java#L32-L57).

timtebeek (approved):
Much appreciate this addition, and how the test covers every aspect without much duplicated setup.

Issue #192, opened 2023-10-14
`if` statements with literal conditions and jumps in their branches aren't simplified


tkindy:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.7.2
- Maven plugin v5.8.1
- rewrite-static-analysis v1.0.8

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
  <groupId>org.openrewrite.maven</groupId>
  <artifactId>rewrite-maven-plugin</artifactId>
  <version>5.8.1</version>
  <configuration>
    <activeRecipes>
      <recipe>org.openrewrite.staticanalysis.SimplifyConstantIfBranchExecution</recipe>
    </activeRecipes>
  </configuration>
  <dependencies>
    <dependency>
      <groupId>org.openrewrite.recipe</groupId>
      <artifactId>rewrite-static-analysis</artifactId>
      <version>1.0.8</version>
    </dependency>
  </dependencies>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
class A {
  public int foo() {
    if (true) {
      return 5;
    }
    return 6;
  }
}
```

## What did you expect to see?

```java
class A {
  public int foo() {
    return 5;
  }
}
```

## What did you see instead?

```java
class A {
  public int foo() {
    if (true) {
      return 5;
    }
    return 6;
  }
}
```

## What is the full stack trace of any errors you encountered?

No errors.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes, I'm interested in contributing a fix.

tkindy:
From some digging, this seems to be the recipe's intended behavior. `if`s with jumps in the branches were ignored in https://github.com/openrewrite/rewrite-static-analysis/commit/5537be97bb56a5aafd21d360030906ef4dce2d27 as the fix for https://github.com/openrewrite/rewrite-java-security/issues/34 where the combination of `SimplifyConstantIfBranchExecution` and a different recipe led to unreachable code when the `if` had a `return` in the kept branch.

However, I think that fix is overly broad. I think the right fix would be to eliminate the unreachable code introduced in that scenario. I could try to implement such a recipe if others agree on this approach.

PR #193, opened 2023-10-14
Simplify `if`s with literal conditions even if they jump

tkindy:
## What's changed?

This PR improves the `SimplifyConstantIfBranchExecution` recipe to simplify `if`s even if they have jumps (`return`, `throw`, `break`, or `continue`) in the branch we keep. The recipe will now also remove any unreachable code introduced by simplifying these `if`s.

I added numerous unit tests to demonstrate the new behavior.

## What's your motivation?

Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/192

I considered implementing the unreachable code elimination as another recipe. However, it feels like it's tightly coupled to the `if` simplification itself; unreachable code is a compile-time error, so if this recipe is going to simplify these `if`s, I think it should do so cleanly.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
@@ -25,6 +25,11 @@ public class RemoveUnreachableCodeVisitor extends JavaVisitor<ExecutionContext> 
     }
     int firstJumpIndex = maybeFirstJumpIndex.get();
 
+    if (firstJumpIndex == statements.size() - 1) {
+      // Jump is at the end of the block, so nothing to do
+      return block;
+    }
+
     List<Statement> newStatements =
         ListUtils.flatMap(
             block.getStatements(),
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
@@ -25,7 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
-public class RemoveUnreachableCodeVisitor extends JavaVisitor<ExecutionContext> {
+class RemoveUnreachableCodeVisitor extends JavaVisitor<ExecutionContext> {
 
   @Override
   public J visitBlock(J.Block block, ExecutionContext executionContext) {
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
@@ -15,7 +15,6 @@
  */
 package org.openrewrite.staticanalysis;
 
-import lombok.AllArgsConstructor;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaVisitor;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Statement;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+public class RemoveUnreachableCodeVisitor extends JavaVisitor<ExecutionContext> {
```
timtebeek, line 28: Maybe best to start with this visitor as package private to limit the publicly exposed API. 
```suggestion
class RemoveUnreachableCodeVisitor extends JavaVisitor<ExecutionContext> {
```

tkindy: Good thinking, I'm not used to public library development. 😄

pstreef:
I really like this change. It's concise and very readable. I've added another unit test for the `throws` keyword with try catch and it behaves as expected.

pstreef (approved):

pstreef:
Thank you for the contribution! 🎉

src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    int firstJumpIndex = maybeFirstJumpIndex.get();
+
+    if (firstJumpIndex == statements.size() - 1) {
+      // Jump is at the end of the block, so nothing to do
+      return block;
+    }
+
+    List<Statement> newStatements =
+        ListUtils.flatMap(
```
knutwannheden, line 47: Note: The way `ListUtils.map()` is implemented, it allows the function to return a `null`, which indicates that the corresponding element should be removed from the list.

knutwannheden: But in this particular case I think it would be even easier with `statements.subList(0, firstJumpIndex)` (give or take a off-by-one error 😄 ).

pstreef: Good point! 
https://github.com/openrewrite/rewrite-static-analysis/pull/206

timtebeek:
Great work all around!

Probably best to run this against the default set through Moderne, just to spot check that it works well there. Well already have an impressive collection of tests here, just looking to make sure we haven't missed anything.

src/main/java/org/openrewrite/staticanalysis/RemoveUnreachableCodeVisitor.java
tkindy: Ah, good point. I initially didn't use `sublist` to avoid the potential for creating a new list if the jump was at the end of the block, and therefore the statements should remain the same (though I suppose it's possible most `List` implementations return themselves if the full list is requested with `sublist`). But once I handled that case separately with the above `if`, we know the statements are going to be trimmed at this point anyway.

tkindy:
Thanks for the feedback and merge, all! 😄

pstreef:
> Great work all around!
> 
> Probably best to run this against the default set through Moderne, just to spot check that it works well there. Well already have an impressive collection of tests here, just looking to make sure we haven't missed anything.

I ran it, all results look good!

PR #197, opened 2023-10-18
no diamond when argument to generic method

joanvr:
## What's changed?
We are trying to detect when we are calling a generic method with type inference, so we do not apply the diamond operator in those scenarios, as we cannot guarantee that the type inference will still work after applying the diamond operator.

## What's your motivation?
Fixes https://github.com/moderneinc/support-app/issues/11
We've found some examples where we are producing invalid code due to applying the diamond operator when calling a generic method, thus breaking the type inference. 

@timtebeek @knutwannheden @jkschneider @kunli2 

I was trying to detect the type inference by trying to find the method in the declaring class, since the methodType I have from the invocation has the types resolves to the inferred ones. But with the TypeUtils.findDeclaredMethod I could not find it, because it's trying to match the arguments, and because the declared method has generics, but the method invocation don't, it was not matching. I took advantage of this by returning without changes when I cannot find the method, but a better approach might be to find the method "ignoring" generics, and only return without changes if the affected parameter (the one where we want to apply the diamond operator) has generics, not any of them.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UseDiamondOperator.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UseDiamondOperator.java
@@ -25,13 +25,11 @@ import org.openrewrite.marker.Markers;
 import org.openrewrite.staticanalysis.java.JavaFileChecker;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 import static java.util.Collections.singletonList;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Nicely done!

kunli2:
Good one! LGTM.
I am also thinking maybe avoiding rewriting when a `J.NewClass` has a block body also makes sense because for these cases, the inferred types are actually from the methods in the body. but that will make some cases changeable unchanged.

kunli2 (approved):

PR #198, opened 2023-10-19
RemoveEmptyJavaDocParameters: added tests + fixed bugs

joanvr:
## What's changed?
There was a lot of missing scenarios on the tests that were causing bugs (wrong format of last lines, first lines, even index out of bounds on some very rare scenarios).
Added all the missing combinations of [param, return, throws], [regular, singleLine, firstLine], [withSpace, noSpace].
Fixed the bugs not originally covered with the new tests.

## What's your motivation?
Fixes https://github.com/moderneinc/support-app/issues/14 and hopefully all other bugs in this recipe.

@timtebeek @knutwannheden 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParameters.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParameters.java
@@ -79,6 +79,12 @@ public class RemoveEmptyJavaDocParameters extends Recipe {
                     boolean useNewBody = false;
 
                     List<Javadoc> body = javadoc.getBody();
+                    // To fix elements on the first line without space
+                    if (!body.isEmpty() && !(body.get(0) instanceof Javadoc.LineBreak)) {
+                        body = new ArrayList<>(body);
+                        body.add(0, null); // we can use null since this element is never going to be used.
+                    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParameters.java
```diff
@@ -79,6 +79,12 @@ public Javadoc visitDocComment(Javadoc.DocComment javadoc, ExecutionContext ctx)
                     boolean useNewBody = false;
 
                     List<Javadoc> body = javadoc.getBody();
+                    // To fix elements on the first line without space
+                    if (!body.isEmpty() && !(body.get(0) instanceof Javadoc.LineBreak)) {
+                        body = new ArrayList<>(body);
+                        body.add(0, null); // we can use null since this element is never going to be used.
+                    }
```
joanvr, line 86: by adding this trailing element, we can always process the elements in this loop, and we no longer require the specific element visitors, that were causing a bit of confusion and complexity to debug the recipe. Now those are gone and changed to checking boolean methods (isEmpty...)

timtebeek (approved):
Nice! I'd briefly explored this previously, but indeed found the code hard to follow. Glad that's sorted now.

PR #204, opened 2023-10-21
Manually create LST for enums in `MinimumSwitchCases`

knutwannheden:
When the types are not known in advance the `JavaTemplate` classpath can also not be configured for these types. As a consequence the corresponding elements cannot be properly type-attributed.

So that the result of `MinimumSwitchCases` is properly type-attributed when transforming a `switch` over an enum, the recipe now manually creates the LST. Then the `ShortenFullyQualifiedTypeReferences` visitor also works properly.


knutwannheden:
Going ahead and merging this PR. We can always revert again or alternatively also use the manual LST building for the `String` and primitive type cases.

PR #205, opened 2023-10-22
RenameLocalVariablesToCamelCase: Do not rename method argument names.

pstreef:
Many frameworks depend on method argument names to use for request parameter names or json field names. By renaming these without knowing how the frameworks use these we can make breaking changes. It's better to do no change.

## What's changed?
Non camel case names are kept in method argument names.

## What's your motivation?
Do no harm

`isMethodArgument()`

@joanvr @timtebeek 

no

### Checklist
- [x] I've used the IntelliJ IDEA auto-formatter on affected files (No because it adds much extra indentation to the test file. I think that's not wanted?)

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
@@ -462,20 +462,19 @@ class RenameLocalVariablesToCamelCaseTest implements RewriteTest {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/pull/205")
     @Test
-    void doNotRenameMethodArguments(){
+    void doNotRenameMethodArguments() {
         //language=java
         rewriteRun(
           java(
             """
-             @Controller
-             class MyController {
-                 @GetMapping
-                 String getHello(@RequestParam String your_name) {
-                     return "hello " + your_name;
-                 }
-             }
-             """
+              class MyController {
+                  String getHello(String your_name) {
+                      return "hello " + your_name;
+                  }
+              }
+              """
           )
         );
     }
--- a/src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
@@ -15,7 +15,9 @@
  */
 package org.openrewrite.staticanalysis;
 
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.ExpectedToFail;
 import org.openrewrite.DocumentExample;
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
```diff
@@ -458,4 +462,22 @@ void test() {
         );
     }
 
+    @Test
+    void doNotRenameMethodArguments(){
+        //language=java
+        rewriteRun(
+          java(
+            """
+             @Controller
```
joanvr, line 471: Are those anotations needed for the test? They are not imported anyway, thus giving incomplete type attribution... maybe we could remove them for simplicity.

joanvr (approved):
I'm not a big fan of removing the method parameters capability of the recipe, but for now seems the most sensitive thing to do if we cannot narrow it a bit more based on som anotations...

src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
pstreef: I added them to show give a more in-depth example of why this is added. I bet most people here recognise which annotations I mean with this.

I did not think the type attribution was worth adding spring as a dependency though.

Although perhaps I should change it to just have an `@Issue` annotation that explains the why?

knutwannheden: Note: Soon we will flag these missing type attributions in before sources as errors (unless the before type validation is configured to ignore these).

joanvr: Yes, please, reference this issue from the test too. As Knut says, we have plans to fail tests that do not have proper type attribution in the before step...

src/test/java/org/openrewrite/staticanalysis/RenameLocalVariablesToCamelCaseTest.java
```diff
@@ -114,6 +116,7 @@ public int addTen(int renameOne) {
     }
 
     @SuppressWarnings("JavadocDeclaration")
+    @Disabled
```
pstreef, line 119: @joanvr did you notice this? I've disabled this test since it fixed an issue that might be important if we ever decide to start (limited) support for renaming method args.

joanvr: Maybe we could try to rename param references in annotations too? Although I'm not sure if we can have a generic enough approach

pstreef: Maybe, but even if there is 1 annotation, it does not mean that it's always used only with that annotation. It's a slippery slope and there will likely still be cases where we could break something.

PR #207, opened 2023-10-24
Cast when needed on lambda expressions

joanvr:
## What's changed?
We need to cast the lambda expression, when we have some unbounded generics on it's body. Right now I'm trying to identify those by searching for method invocations of methods with arguments with parametrized types with generics, but this just narrows it a bit, but those generics not necessarily are unbounded without the explicit cast. 

Also, changed the `methodArgumentRequiresCast` and previous check to accept MethodCall instead of `J.MethodInvocation`, so we can also reuse it for `J.NewClass` too

## What's your motivation?
Fixes https://github.com/moderneinc/support-app/issues/17

How we can detect better when it's needed to add the cast and when it's redundant?

### Checklist


knutwannheden (approved):

knutwannheden (commented):
👍

PR #211, opened 2023-11-03
Rewriting isEmpty instead of length

jeexan2:
### Checklist


timtebeek:
As discussed on https://github.com/openrewrite/rewrite-migrate-java/issues/336#issuecomment-1791554539 we should target [UseStringIsEmpty.java](https://github.com/openrewrite/rewrite-migrate-java/blob/v2.1.1/src/main/java/org/openrewrite/java/migrate/lang/UseStringIsEmpty.java) in https://github.com/openrewrite/rewrite-migrate-java/ instead; that's a different repository to clone, so closing this PR. Thanks for wanting to get started though!

PR #212, opened 2023-11-06
No method reference on new class select

timtebeek:
As shared instance might behave different from new instances

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
@@ -1290,4 +1290,32 @@ class ReplaceLambdaWithMethodReferenceTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void newInstanceMethodCall() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              class A {
+                  int seen = 0;
+                  String lower(String s) {
+                      seen++;
+                      return s.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReferenceTest.java
```diff
@@ -1290,4 +1290,32 @@ private Integer foo(String bar) {
...
+                  }
+              }
+              """
+          ),
+          java(
+            """
+              import java.util.stream.Stream;
+              class B {
+                  void bar(Stream<String> stream) {
+                      stream.map(s -> new A().lower(s));
```
timtebeek, line 1314: This used to insert the following, but that would then share the instance across operations, inflating the `seen` counter above.
```
                      stream.map(new A()::lower);
```

knutwannheden: @timtebeek I believe the same problem applies if the select is a method invocation. Imagine having `s -> a().lower(s)` with
```java
    A a() {
        return new A();
    }
```

knutwannheden: Turned out that case was already covered.

timtebeek: Thanks for the rework! Hadn't spotted that other one, but like how you've solved it now.

PR #216, opened 2023-11-13
Remove duplicated semicolons, without removing newline

timtebeek:


src/main/java/org/openrewrite/staticanalysis/RemoveExtraSemicolons.java
```diff
@@ -69,9 +62,13 @@ public J.Block visitBlock(final J.Block block, final ExecutionContext executionC
                     if (statement instanceof J.Empty) {
                         nextNonEmptyAggregatedWithComments(statement, iterator)
                                 .ifPresent(nextLine -> {
-                                    Space updatedPrefix = nextLine.getPrefix()
-                                            .withWhitespace(statement.getPrefix().getWhitespace());
-                                    result.add(nextLine.withPrefix(updatedPrefix));
+                                    String semicolonPrefixWhitespace = statement.getPrefix().getWhitespace();
+                                    if (semicolonPrefixWhitespace.isEmpty()) {
```
knutwannheden, line 66: Only wondering if we instead should be checking for comments (in which case we probably want to leave the statement alone) and if the whitespace contains `\n` or not. Or what happens when there is `; ;` (with a space)?

timtebeek: Thanks for the suggestion, indeed there was a problem with `; ;`. Comments were already extensively tested in the unit test just above, so I expect that to be handled.

knutwannheden (approved):

PR #231, opened 2023-12-30
Simplify Ternary expressions [#203]

### LINKED ISSUE #203: Cover `? true : false`
Does OpenRewrite cover the simple rewrite of `? true : false`?

```diff
-        this.isNewItem = (item.extensionProperty().get().equals("")) ? true : false;
+        this.isNewItem = (item.extensionProperty().get().equals(""));
```

Refs https://github.com/openrewrite/rewrite-static-analysis/issues/192

ammachado:
## What's changed?
Simplify ternary expressions (closes #203)

## What's your motivation?
Explore `rewrite-templating`

No

No one

No

## Any additional context
Requires changes provided by https://github.com/openrewrite/rewrite/pull/3861

### Checklist


timtebeek:
Thanks for starting this one! I've added two new cases to the tests to check handling on negation. Without the parentheses in the after template we could end up only negating the first boolean in a binary, changing the logic. I'm not yet pleased with the left over parentheses, but think we might need to fix that in rewrite-templating first.

timtebeek:
After fixes in rewrite-templating and rewrite it appears we now need a fix in rewrite-kotlin before we can get this one through. Thanks for your patience! :)

src/main/java/org/openrewrite/staticanalysis/SimplifyTernary.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2022 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

src/test/java/org/openrewrite/staticanalysis/SimplifyTernaryTest.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright 2022 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

Issue #237, opened 2024-01-12
`ReplaceLambdaWithMethodReference` replaces `item -> item.getClass()` which leads to compile time error


timo-abele:
## What version of OpenRewrite are you using?
I am using

- Maven plugin v5.18.0
- rewrite-static-analysis v1.2.0

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>5.18.0</version>
    <configuration>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.ReplaceLambdaWithMethodReference</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-static-analysis</artifactId>
            <version>1.2.0</version>
        </dependency>
    </dependencies>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java

class Animal {}
class Cat extends Animal {}
class Dog extends Animal {}

public void groupOnGetClass() {
  List<Animal> animals = List.of(new Cat(), new Dog());
  Map<Class<? extends Animal>, List<Animal>> collect;
  collect = animals.stream().collect(Collectors.groupingBy(a -> a.getClass()));
}
```

## What did you expect to see?

```java
public void groupOnGetClass() {
  List<Animal> animals = List.of(new Cat(), new Dog());
  Map<Class<? extends Animal>, List<Animal>> collect;
  collect = animals.stream().collect(Collectors.groupingBy(Animal::getClass));
}
```

`getClass` called on the smallest common super class. (or no change at all)

## What did you see instead?

```java
public void groupOnGetClass() {
  List<Animal> animals = List.of(new Cat(), new Dog());
  Map<Class<? extends Animal>, List<Animal>> collect;
  collect = animals.stream().collect(Collectors.groupingBy(java.lang.Object::getClass));
}
```

## Notes

I know grouping on `Class` isn't great in the first place, but this is real code. 
In my case the type information wasn't used so I can simply replace `Map<Class<? extends Animal>` with `Map<Class<?>`.

## What is the full stack trace of any errors you encountered?

Compiler says
```
java: incompatible types: inference variable K has incompatible bounds
    equality constraints: java.lang.Class<? extends org.example.Main.Animal>
[...]

PR #240, opened 2024-01-15
ReplaceLambdaWithMethodReference skip parameterized method return types

timtebeek:
Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/237

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java
@@ -170,6 +170,9 @@ public class ReplaceLambdaWithMethodReference extends Recipe {
                 if (methodType != null && !isMethodReferenceAmbiguous(methodType)) {
                     if (methodType.hasFlags(Flag.Static) ||
                         methodSelectMatchesFirstLambdaParameter(method, lambda)) {
+                        if (method.getType() instanceof JavaType.Parameterized) {
+                            return l;
+                        }
                         J.MemberReference updated = newStaticMethodReference(methodType, true, lambda.getType()).withPrefix(lambda.getPrefix());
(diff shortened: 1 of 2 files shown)


knutwannheden:
For now it is probably safest if we try to exclude this type of cases. I wonder if we can narrow it down a bit, however. Maybe we can check if the parameterized type uses any generic type variables that are not declared by the method itself.

src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java
```diff
@@ -170,6 +170,9 @@ public J visitLambda(J.Lambda lambda, ExecutionContext ctx) {
                 if (methodType != null && !isMethodReferenceAmbiguous(methodType)) {
                     if (methodType.hasFlags(Flag.Static) ||
                         methodSelectMatchesFirstLambdaParameter(method, lambda)) {
+                        if (method.getType() instanceof JavaType.Parameterized) {
```
timtebeek, line 173: @knutwannheden did you mean something like this? Or should we restrict it even further?
```suggestion
                        if (method.getType() instanceof JavaType.Parameterized &&
                            ((JavaType.Parameterized) method.getType()).getTypeParameters().stream()
                                    .anyMatch(JavaType.GenericTypeVariable.class::isInstance)) {
```

knutwannheden: I was thinking something like that, yes. Otherwise we exclude a lot of cases where the recipe actually works correctly. But I can't say if this is 100% correct or not, as the type inference topic is rather complex. So if we can ensure that the new test case doesn't break and possibly some new test cases assert that it still works for other parameterized return types, then that is good enough, I think.

As an extension we could also check if all of the generic type variables are defined by the method itself. In that case I think the method reference replacement should be safe.

Issue #236, opened 2024-01-12
LambdaBlockToExpression does not work in assertThrows (or generally when no value is returned)


timo-abele:
## What version of OpenRewrite are you using?
I am using

- Maven plugin v5.18.0
- rewrite-static-analysis v1.2.0

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>5.18.0</version>
    <configuration>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.LambdaBlockToExpression</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-static-analysis</artifactId>
            <version>1.2.0</version>
        </dependency>
    </dependencies>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
import static org.junit.jupiter.api.Assertions.assertThrows;

import java.math.BigDecimal;
import org.junit.jupiter.api.Test;

public class SimpleTest {

  @Test
  void shouldFailOnDivisionByZero() {
    assertThrows(ArithmeticException.class, () -> {
      BigDecimal.ONE.divide(BigDecimal.ZERO);
    });
  }

}
```

## What did you expect to see?

```java
import static org.junit.jupiter.api.Assertions.assertThrows;

import java.math.BigDecimal;
import org.junit.jupiter.api.Test;

public class SimpleTest {

  @Test
  void shouldFailOnDivisionByZero() {
    assertThrows(ArithmeticException.class, () -> BigDecimal.ONE.divide(BigDecimal.ZERO));
  }

}
```

## What did you see instead?

no change whatsoever.

## Notes

The example above is how I encountered this bug. The examplee below are a bit more minimal but less common in my opinion.
```java
package org.example;

import javax.swing.JButton;

public class Main {

  public static void main(String[] args) {
    Runnable runHelloWorld = () -> {
      System.out.println("Hello world!");
    };
    runHelloWorld.run();

    JButton jb = new JButton();
    jb.addActionListener(event -> {
      System.out.println(7);
    });
  }
}
```

All three examples have in common that they don't return a value, so [the description in the docs](https://docs.openrewrite.org/recipes/staticanalysis/lambdablocktoexpression):

[...]

PR #241, opened 2024-01-16
feature(LambdaBlockToExpression): convert lambda with method invocation as well

timo-abele:
## What's changed?

This PR will enable the transformation of lambdas with method invocation as body.

## What's your motivation?

Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/236

1. The first test case I added doesn't work, see comment in code
2. I had to explicitly set the whitespace in the return statement (LambdaBlockToExpression.java L61) in order to have the code move up to the previous line. Is there a cleaner way to do that?

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
@@ -22,6 +22,9 @@ import org.openrewrite.java.JavaParser;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
+import java.math.BigDecimal;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.openrewrite.java.Assertions.java;
 
 class LambdaBlockToExpressionTest implements RewriteTest {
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
```diff
@@ -114,4 +117,97 @@ void doTest() {
...
+    //    at org.openrewrite.java.Assertions.assertValidTypes(Assertions.java:87)
+    //    at org.openrewrite.java.Assertions.validateTypes(Assertions.java:57)
+    //    at org.openrewrite.java.Assertions$$Lambda$438/0x00007fbb18152c28.accept(Unknown Source)
+    //    at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:304)
+    //    at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:132)
+    //    at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:127)
+    //    at org.openrewrite.staticanalysis.LambdaBlockToExpressionTest.simplifyLambdaBlockReturningVoidAsWell(LambdaBlockToExpressionTest.java:121)
+    //    at java.base/java.lang.reflect.Method.invoke(Method.java:568)
+    //    at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
+    //    at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```
timtebeek, line 139: Yes while it's possible to [specify classpath for dependencies in tests](https://docs.openrewrite.org/authoring-recipes/recipe-testing#specifying-the-classpath-for-dependencies), for this repository it might make most sense to rewrite the test to only use classes available in the JDK; I think that should be manageable here.

If you're curious this shows an example of declaring dependencies on the test classpath:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/a426e409ca3b8e04c6470e86cc09d61346e7c0f4/src/test/java/org/openrewrite/java/testing/cleanup/RemoveTestPrefixTest.java#L32-L33
But again: let's try to avoid that if we can.

src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
```diff
@@ -114,4 +117,97 @@ void doTest() {
...
+                void shouldFailOnDivisionByZero() {
+                  assertThrows(ArithmeticException.class, () -> BigDecimal.ONE.divide(BigDecimal.ZERO));
+                }
+              
+              }
+              """
+          )
+        );
+    }
+
```
timtebeek, line 182: If I understand you correctly this test isn't strictly needed, but was added to attempt to replicate the exact scenario where we you want it applied. I think we sufficiently cover that with the below test as well, so perhaps easiest to remove this one.
```suggestion
```

timtebeek (commented):
Great start at contributing here @timo-abele ! I've added a comment to explain the use of classpaths for tests, and suggestion to remove the second test. Beyond that I think we can try this out if you agree with removing that second test.

timo-abele:
There was another test that failed namely `UseLambdaForFunctionalInterfaceTest::nestedLambdaInMethodArgument` I didn't expect this but apparently UseLambdaForFunctionalInterface depends on this recipe here. Log:
```
expected: 
  "import java.util.function.Consumer;
  
  class Test {
      void bar(Consumer<Integer> c) {
      }
      void foo() {
          bar(i -> {
              bar(i2 -> {
              });
          });
      }
  }"
 but was: 
  "import java.util.function.Consumer;
  
  class Test {
      void bar(Consumer<Integer> c) {
      }
      void foo() {
          bar(i -> bar(i2 -> {
          }));
      }
  }"
```
The new result looks good to me, can I simply adapt the test?

timtebeek:
> The new result looks good to me, can I simply adapt `UseLambdaForFunctionalInterfaceTest::nestedLambdaInMethodArgument`?

Yeah sure! Agree with your observation there. Feel free to mark your PR ready for review once done.

timtebeek (approved):
Added two small changes to minimize casts and white space changes; hope you agree! Typically we leave white space changes to the specific style configured rather than setting an explicit space to use. Figured in this case folks might have chosen to add white space, rather than  `{ statement }`, so we should be mindful an keep that in.

Neat addition to this recipe! Look forward to how it will affect changes at scale. Thanks again!

timo-abele:
I agree with the casts and I don't mind the white space changes, it did feel strange adding it explicitly in the code. But the reason I did it was to be consistent with the existing behavior https://github.com/openrewrite/rewrite-static-analysis/blob/f1e53d203de026938615fee8b5175cba379dc279/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java#L38-L58 
where the white space _is_ removed (because the `expression` is extracted from the `statement` and the `statement` holding the white space is dropped). Since the `statement` in the lambda is only one line, I think it is less of a surprise to continue the line with the `->` and not to break it (by keeping the white space).

timtebeek:
You're right; in the case of `return` we implicitly drop any white space that was associated with the `return` by only returning the expression. Guess we'll have to see how this works out in practice if we should do the same for statements as well. I do quite like how only the lines that contained `{` and `}` are changed with the current approach leaving the line with the statement untouched. If you used `{ statement }` on a single line it also works as expected already. To be seen when run at scale. :)

timo-abele:
Hi @timtebeek, could you please release a new version? I'd like to try these changes out.

timtebeek:
Know that you can always try out the latest builds from [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions); we'll likely folllow up with a release next week.

Issue #242, opened 2024-01-17
org.openrewrite.staticanalysis.RemoveUnusedPrivateFields does not recognize @Data Lombok annotation


gwydionmv:
Hello!

Similar to this other open issue [here](https://github.com/openrewrite/rewrite-static-analysis/issues/8), the recipe RemoveUnusedPrivateFields is deleting private fields even though the class is annotated as @Data. 

I am using 
- OpenRewrite Maven plugin v5.19.0
- rewrite-static-analysis v1.2.0

I can help working on fixing this issue if you find it interesting too :D

timtebeek:
Hi @gwydionmv ! Thanks for pointing that out and offering to help.

I imagine the fix would be similar to what we do for explicit initialization:
https://github.com/openrewrite/rewrite-static-analysis/blob/a5dc2f7049f7ef4554180c630af75985d388ac4b/src/main/java/org/openrewrite/staticanalysis/ExplicitInitializationVisitor.java#L59-L62
Which we then test here
https://github.com/openrewrite/rewrite-static-analysis/blob/6a38549b00cad1088fc1588c4ba1f645113bb991/src/test/java/org/openrewrite/staticanalysis/ExplicitInitializationTest.java#L66-L81

timtebeek:
We would likely need similar handling in `RemoveUnusedPrivateFields`
https://github.com/openrewrite/rewrite-static-analysis/blob/e57dac0991cf3e57edf99826d975c079d0971baf/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateFields.java#L67-L68
And best to have a matching test as well in
https://github.com/openrewrite/rewrite-static-analysis/blob/23ee04cc6c9baff6677a5c7efc233f28cd244b1c/src/test/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateFieldsTest.java#L26-L30

timtebeek:
Would be much appreciated if you could provide a fix for that, which I'll gladly review shortly. We can then do the same for #8 too.

gwydionmv:
I just created the PR @timtebeek 😄  https://github.com/openrewrite/rewrite-static-analysis/pull/244

timtebeek:
Many thanks! Fixed in
- https://github.com/openrewrite/rewrite-static-analysis/pull/244

Issue #8, opened 2023-04-15
org.openrewrite.java.cleanup.RenamePrivateFieldsToCamelCase does not recognize @Data lombok annotation


blipper:
org.openrewrite.java.cleanup.RenamePrivateFieldsToCamelCase only checks if there is a private field.  It does not recognize @Data or any other lombok annotation so renaming the field will introducing a breaking change.  To preserve compatibility it should ignore private fields if the class is annotated with @Data/@Getter/@Setter

timtebeek:
@gwydionmv added a very similar change in 
- https://github.com/openrewrite/rewrite-static-analysis/pull/244

That could be replicated here to make `RenamePrivateFieldsToCamelCase` skip classes or fields with Lombok annotations.

timtebeek:
As part of that change we'd then likely want to do the same for ExplicitInitializationVisitor
https://github.com/openrewrite/rewrite-static-analysis/blob/a5dc2f7049f7ef4554180c630af75985d388ac4b/src/main/java/org/openrewrite/staticanalysis/ExplicitInitializationVisitor.java#L59-L62

gwydionmv:
@timtebeek I can take care of this too :D

blipper:
Ultimately what think is needed for all of these annotation processors is a mechanism where rewrite has access to the paired byte code generated at compilation time.  AST analysis will never be enough with annotation processors.

blipper:
Or a at a minimum if you want to keep pure AST analysis you need to pair with the delomboked source

timtebeek:
Thanks for helping think this one through! The reason I'd want to skip lombok annotated classes and fields is that this recipe also changes snake case to camel case, as seen in this test case
https://github.com/openrewrite/rewrite-static-analysis/blob/541198c276fc0e5b45c6a9a9d805e23779be2267/src/test/java/org/openrewrite/staticanalysis/RenamePrivateFieldsToCamelCaseTest.java#L45-L50

I'm not a 100% sure what getters lombok would generate before and after the rename, but I think that'd be `getD_TYPE_CONNECT()`, so best skipped.

gwydionmv:
[...]

PR #244, opened 2024-01-18
#242 Fix remove unused private fields

gwydionmv:
Adding a condition to skip applying the recipe if @lombok.Data annotation is found in the class

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateFields.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateFields.java
@@ -39,10 +39,7 @@ import java.util.stream.Collectors;
 @Value
 @EqualsAndHashCode(callSuper = true)
 public class RemoveUnusedPrivateFields extends Recipe {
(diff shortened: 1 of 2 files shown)


timtebeek:
Great work here @gwydionmv ! I've expanded your initial proposal to include four more Lombok annotations, such that we now do not remove private fields if there's any annotation from the lombok package, such that we cover additional cases such as `@Value`, `@Getter` but also `@EqualsAndHashcode` and `@ToString`, as they might subtly affect how code works. This would not match logging annotations for instance, as in those cases it's likely safe to remove unused private fields.

If you agree with this change we can do the same for some of the other cases we've identified, to further harden these recipes when lombok annotations are present. Let me know!

timtebeek (approved):

timtebeek:
Awesome; yes would be great if you'd be willing to replicate this for the other case we identified in https://github.com/openrewrite/rewrite-static-analysis/issues/8#issuecomment-1898233656
That should really help stability; and thinking of this now we should probably also skip _fields_ with Lombok annotations, not just entire classes. 🤔

PR #245, opened 2024-01-19
fix(LambdaBlockToExpression): convert lambda with method invocation in assertThrows as well

timo-abele:
## What's changed?

This PR will enable the transformation of lambdas with method invocation as body within assertThrows, a feature that unexpectedly wasn't fixed with 
- https://github.com/openrewrite/rewrite-static-analysis/pull/241.

## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/236

@timtebeek 

### Checklist


timtebeek:
Welcome back! :) I've changed the way your PR uses the recipe dependencies such that it's more in line with what we use elsewhere; also briefly stepped through with a debugger and noticed indeed that the recipe isn't triggered, and not yet clear why.

timtebeek:
It's not converted because of this block
https://github.com/openrewrite/rewrite-static-analysis/blob/f1e53d203de026938615fee8b5175cba379dc279/src/main/java/org/openrewrite/staticanalysis/LambdaBlockToExpression.java#L88-L108

Not sure how much we can improve this here without potentially breaking things elsewhere. 🤔

timtebeek:
Fun puzzle, thanks! I think it would be fixed with https://github.com/openrewrite/rewrite-static-analysis/pull/245/commits/1ac607fc4574effa2efbdeaadbe27ab747d2dcfc

timtebeek:
@knutwannheden would you agree with the isolated change in https://github.com/openrewrite/rewrite-static-analysis/pull/245/commits/1ac607fc4574effa2efbdeaadbe27ab747d2dcfc ? I know we've been careful where we convert blocks and remove type casts, but I feel this minimal change might add a lot more reductions without additional risk that I can see now.

timtebeek:
I wouldn't necessarily merge the tests we have here now, as I'd prefer to keep the tests in rewrite-static-analysis free from dependencies, but it did help bring to light why were not replacing these instances, so thanks for that @timo-abele !

timo-abele:
Hmm, discriminating by "same number of arguments" solves my use case, because it excludes the problematic case from https://github.com/openrewrite/rewrite-static-analysis/issues/162
```
void aMethod(Consumer<Integer> consumer)
void aMethod(Function<Integer,String> function)
```
whereas the heterogeneous `assertThrows` with 
```
assertThrows(Class<T> expectedType, Executable, executable)
assertThrows(Class<T> expectedType, Executable, executable, String message)
assertThrows(Class<T> expectedType, Executable, executable, Supplier<String> messageSupplier)
```
would be included. However a hypothetical
```
void bMethod(Consumer<Integer> consumer, String message)
void bMethod(Function<Integer,String> function, String message)
```
would be included again when it shouldn't be. Maye we should instead discriminate by "The parameter at the position where the lambda is should always have the same type"?

timtebeek:
That seems better indeed, and thanks for noticing that potential issue. Would you want to wire that up yourself?

timo-abele:
I've sketched something up, all tests run through but please review diligently I ended up rewriting more then I planned.

src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java
```diff
@@ -25,11 +26,12 @@
 import static org.openrewrite.java.Assertions.java;
 
 class LambdaBlockToExpressionTest implements RewriteTest {
-
     @Override
     public void defaults(RecipeSpec spec) {
         spec.recipe(new LambdaBlockToExpression())
-            .parser(JavaParser.fromJavaVersion().logCompilationWarningsAndErrors(true));
+            .parser(JavaParser.fromJavaVersion()
+              .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.10")
```
knutwannheden, line 33: Since this dependency is only used for tests, we should make sure to not add it to `src/main/resources`.

src/main/java/org/openrewrite/staticanalysis/LambdaBlockToExpression.java
```diff
@@ -79,42 +81,66 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
-                            if (++overloadingCount > 1) {
-                                return true;
-                            }
-                        }
-                    }
-                    return false;
-                })
-                .orElse(false);
-    }
+                .orElse(Collections.emptyList());
```
knutwannheden, line 112: Rather than collecting all methods into a list, it would be more efficient to loop over them.

I am also saying this, because we should actually also be checking methods in super types and even default methods on implemented interfaces. But this can quickly get difficult, as I am not convinced that we will be reliable able to tell if a method is an override of some other method (and thus not an overload) when generics are involved. But once we cross that road, a recursive method which returns as soon as it finds a conflicting overload will be more efficient than first gathering all methods up front.

knutwannheden (changes requested):

src/main/java/org/openrewrite/staticanalysis/LambdaBlockToExpression.java
timo-abele: I won't find time to finish this and have unassigned me. I trust your judgement that not collecting makes a enough of a difference to justify the loss in readability. I found the previous version very hard to get into (at first I thought it `map`s over the list elements, but it is always over the `Optional`): https://github.com/openrewrite/rewrite-static-analysis/blob/e57dac0991cf3e57edf99826d975c079d0971baf/src/main/java/org/openrewrite/staticanalysis/LambdaBlockToExpression.java#L80-L97 and so I hope the assignment to `MethodsOfType` can stay: https://github.com/openrewrite/rewrite-static-analysis/blob/758e7cfbec5e306f2e0b9487d9241ae4fe502f50/src/main/java/org/openrewrite/staticanalysis/LambdaBlockToExpression.java#L113-L118 Since the list is already there (from `getMethods`) and `Collections.emptyList()` is a constant, I think it makes no difference in performance while neatly separating the optional  stream from the loop over the elements.

knutwannheden: Yes, you are right, there is no additional list being created here. I was already thinking ahead for the use case with inherited method declarations. But let's forget about that for the moment then (a bug will probably get reported at some point).

timtebeek:
Had a quick look: the test that was added here now passes, provided that I add the junit dependency:
```java
    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/236")
    @Test
    void simplifyLambdaBlockReturningVoidAsWellAssertThrows() {
        //language=java
        rewriteRun(
          spec-> spec.parser(JavaParser.fromJavaVersion().classpath("junit")),
          java(
            """
              import static org.junit.jupiter.api.Assertions.assertThrows;

              import java.math.BigDecimal;

              public class Main {
                @Test
                public void test() {
                  assertThrows(ArithmeticException.class, () -> {
                    BigDecimal.ONE.divide(BigDecimal.ZERO);
                  });
                }
              }
              """,
            """
              import static org.junit.jupiter.api.Assertions.assertThrows;

              import java.math.BigDecimal;

              public class Main {
                @Test
                public void test() {
                  assertThrows(ArithmeticException.class, () ->
                    BigDecimal.ONE.divide(BigDecimal.ZERO));
                }
              }
              """
          )
        );
    }
```

What that I think this issue can be closed. Thanks again!

Issue #230, opened 2023-12-28
`RemoveToStringCallsFromArrayInstances` should not use `Arrays.toString` on String/PrintStream/PrintWriter.format methods


philippe-granet:
The following code:
```java
    Object[] strings = new Object[]{"foo", "bar"};
    System.out.println(String.format("%s %s", strings));
```
display `foo bar`.  
But `RemoveToStringCallsFromArrayInstances` recipe generate this code::
```java
    Object[] strings = new Object[]{"foo", "bar"};
    System.out.println(String.format("%s %s", Arrays.toString(strings)));
```
that fail with this error:
```
Exception in thread "main" java.util.MissingFormatArgumentException: Format specifier '%s'
at java.base/java.util.Formatter.format(Formatter.java:2780)
at java.base/java.util.Formatter.format(Formatter.java:2717)
	at java.base/java.lang.String.format(String.java:4150)
	at HelloWorld.main(HelloWorld.java:8)
```
I think `RemoveToStringCallsFromArrayInstances` recipe should not use `Arrays.toString` on `String/PrintStream/PrintWriter format` methods.

timtebeek:
Thanks for the clear and detailed report! That's indeed a bug that could be fixed with checking the surrounding method invocation. Is that also something you'd like to work on?

To be sure: we appreciate every report, but it can be hard to get to everything. Don't be discouraged from reporting more issues even if you don't have time to contribute! :)

knutwannheden:
Good catch! I think the issue here is that the method being called is a varargs method (e.g. `String#format(String, Object...)`) and that the recipe probably doesn't correctly handle that case.

philippe-granet:
> Thanks for the clear and detailed report! That's indeed a bug that could be fixed with checking the surrounding method invocation. Is that also something you'd like to work on?

You're welcome 😀
Sorry, I'd would love to help but I don't have time available at the moment to work on this problem

timtebeek:
The recipe has some handling to explicitly include arrays passed into one of various print/format methods:
https://github.com/openrewrite/rewrite-static-analysis/blob/30b963b9c36f3f1a4c580a0fcc1b36405b9a7e29/src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java#L76-L84

It seems as though that wasn't optimal though, as it produced additional perhaps unwanted changes. I've opened a PR with an alternative but would like your input on potential additional cases to cover:
[...]

PR #246, opened 2024-01-19
Do not adopt Arrays.toString for varargs arguments

timtebeek:
Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/230

## What's changed
In this proposal we look at the method parameter type, and if that is also an array we do not produce a change.

Not sure I actually like this change yet; It seems before we were too eager to make changes, even where we changed to logic. The recipe was intended not to show the result of `arr.toString()`, that that wasn't what happened on some of the previous cases. Take for example

```java
jshell> String[] args = new String[] {"a","b","c"}
args ==> String[3] { "a", "b", "c" }

jshell> String.format("s=%s", args)
|  Warning:
|  non-varargs call of varargs method with inexact argument type for last parameter;
|    cast to java.lang.Object for a varargs call
|    cast to java.lang.Object[] for a non-varargs call and to suppress this warning
|  String.format("s=%s", args)
|                        ^--^
$6 ==> "s=a"
```
> [!NOTE]
> How we print `s=a`, not `s=[Ljava.lang.String;@73a8dfcc` as would be the case for `args.toString()`

That means that the recipe might inadvertently in the old situation changes the above to start producing
`s=[a, b, c]` for `String.format("s=%s", Arrays.toString(args))`

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
@@ -19,7 +19,10 @@ import org.openrewrite.*;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.MethodMatcher;
-import org.openrewrite.java.tree.*;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypedTree;
 
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


philippe-granet:
> @philippe-granet do these changes align with what you expected after logging
> 
> * [`RemoveToStringCallsFromArrayInstances` should not use `Arrays.toString` on String/PrintStream/PrintWriter.format methods #230](https://github.com/openrewrite/rewrite-static-analysis/issues/230) ?

@timtebeek yes it's OK for me, thanks!

src/main/java/org/openrewrite/staticanalysis/RemoveToStringCallsFromArrayInstances.java
```diff
@@ -75,9 +78,17 @@ public J visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
                 return buildReplacement(select, mi);
             } else if (METHOD_MATCHERS.stream().anyMatch(matcher -> matcher.matches(mi))) {
                 // deals with edge cases where .toString() is called implicitly
+                JavaType.Method methodType = mi.getMethodType();
+                if (methodType == null) {
+                    return mi;
+                }
+                List<JavaType> parameterTypes = methodType.getParameterTypes();
```
knutwannheden, line 85: An alternative would be to rely on `Flag.Varargs` which when set on the `JavaType.Method` means that it is a varargs method.

timtebeek: Didn't know that flag, thanks! But then that only applies to the entire method right? Yet we'd still want to do replacements on String.format and friends when we can. Say String.format gets called with 6 arguments, one of them is an array. The current implementation would then still wrap that one array in `Arrays.toString`. Or how would you apply that flag here?

knutwannheden: Not sure. I would probably just use the flag to check what the recipe should do with the arguments corresponding to the (last) varargs parameter, if any. Probably no huge improvement...

PR #250, opened 2024-01-26
`MinimumSwitchCases`: Case with multiple constants

knutwannheden:


src/test/java/org/openrewrite/staticanalysis/MinimumSwitchCasesTest.java
```diff
@@ -621,22 +621,30 @@ class Test {
                   int variable;
                   void test() {
                       switch (variable) {
-                        case 0 -> doSomething();
-                        default -> doSomethingElse();
+                        case 0, 1: doSomething(); break;
+                        default: doSomethingElse(); break;
```
timtebeek, line 625: I see the same case covered by two tests now; should this have remained a switch expression?
```suggestion
                        case 0, 1 -> doSomething();
                        default -> doSomethingElse();
```

timtebeek (approved):

PR #251, opened 2024-01-28
Add URLEqualHashCode Recipe

JLLeitschuh:
Based upon, but not completely fixing this: https://errorprone.info/bugpattern/URLEqualsHashCode

Signed-off-by: Jonathan Leitschuh <Jonathan.Leitschuh@gmail.com>

## What's changed?

Adds a recipe `URLEqualsHashCode`.

## What's your motivation?

I was curious about trying out the new template recipe generation. This seemed like a fun way to try it out.

Are there any use cases/tests that I likely missed?

Thought about writing the recipe not using the template engine, but this seemed like a good use case.

Also, the reason I'm using `URI#toString` to create a URI, instead of `URL#toURI` is because `URL#toURI` throws a checked exception

### Checklist


knutwannheden:
Interesting. I suppose there are also legitimate uses (like file URLs). Otherwise any HashSet or HashMap would also be problematic. In this case the URLs are known to be for files: https://github.com/jetty/jetty.project/blob/5d9679adf861bc589fdb213758a88f16436f6286/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/MetaInfConfiguration.java#L660

JLLeitschuh:
Interesting insight. Not sure if there's a great way to detect that. In general, this could be an optionally applied recipe.

Dealing with collections could be super complicated, especially when the collection is returned/escapes the scope of the method.

JLLeitschuh:
Is this good to merge?

src/main/java/org/openrewrite/staticanalysis/URLEqualsHashCode.java
```diff
@@ -0,0 +1,69 @@
...
+import java.net.URI;
+import java.net.URL;
+
+@SuppressWarnings("UrlHashCode")
+@RecipeDescriptor(
+        name = "URL Equals and Hash Code",
+        description = "Uses of equals() and hashCode() cause java.net.URL to make blocking internet connections. " +
+                      "Instead, use java.net.URI.",
+        tags = {"errorprone"}
+)
```
knutwannheden, line 31: I think we should add some Markdown formatting here. Also, I don't think the `errorprone` tag is very useful. Does Errorprone have a check of this type?

JLLeitschuh: https://errorprone.info/bugpattern/URLEqualsHashCode

I'll add the markdown formatting. Good point. 😄

src/test/java/org/openrewrite/staticanalysis/URLEqualsHashCodeTest.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+package org.openrewrite.staticanalysis;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+public class URLEqualsHashCodeTest implements RewriteTest {
```
timtebeek, line 24:
```suggestion
class URLEqualsHashCodeTest implements RewriteTest {
    @DocumentExample
          java(
                """
```

src/test/java/org/openrewrite/staticanalysis/URLEqualsHashCodeTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new URLEqualsHashCodeRecipes());
+    }
+
+    @Test
```
timtebeek, line 31:
```suggestion
    @DocumentExample
    @Test
```

src/test/java/org/openrewrite/staticanalysis/URLEqualsHashCodeTest.java
```diff
@@ -0,0 +1,88 @@
...
+    void urlHashCode() {
+        rewriteRun(
+          java("""
```
timtebeek, line 34:
```suggestion
          java(
                  """
```

src/test/java/org/openrewrite/staticanalysis/URLEqualsHashCodeTest.java
```diff
@@ -0,0 +1,93 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.staticanalysis;
+
+import org.junit.jupiter.api.Test;
+
```
timtebeek, line 19:
```suggestion
import org.junit.jupiter.api.Test;
```

PR #257, opened 2024-02-11
Do not remove type casts from lambdas or method references

timtebeek:
## What's changed?
Removed the `&& castType instanceof JavaType.Parameterized` to instead not remove any type casts from lambdas or method references, given the example in the test, and our lean to do no harm.

## What's your motivation?
In response to a reported issue.


knutwannheden (approved):
@timtebeek I think we should really be checking if the required type is a SAM type (functional interface) or not. If it isn't, then a cast is required. But we can do that as a second step.

PR #258, opened 2024-02-17
Add `BufferedWriterCreation` recipe

JLLeitschuh:
- BufferedWriterCreation
 - SimplifyConstantTernaryExecution

Signed-off-by: Jonathan Leitschuh <Jonathan.Leitschuh@gmail.com>

## What's your motivation?

More recipes, more better?

There seems to be a bug in `autoFormat` where the `J.Literal` isn't having it's prefix padding removed.

 - @knutwannheden

### Checklist


JLLeitschuh:
This PR is currently failing. It seems that `autoFormat` isn't working correctly in the `SimplifyConstantTernaryExecution`, and I have no idea what I'm doing wrong. Any insights here would be greatly appreciated.

JLLeitschuh:
Also, waiting on https://github.com/openrewrite/rewrite/pull/4015 to land too for this PR

src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class SimplifyConstantTernaryExecution extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Simplify constant ternary branch execution";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Checks for ternary expressions that are always `true` or `false` and simplifies them.";
```
knutwannheden, line 37: Can we add this to the SimplifyTernary Refaster recipe?

src/main/java/org/openrewrite/staticanalysis/BufferedWriterCreation.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public Set<String> getTags() {
+        return Collections.singleton("CWE-755");
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        JavaVisitor<ExecutionContext> javaVisitor = new AbstractRefasterJavaVisitor() {
```
knutwannheden, line 57: This is interesting. I had never thought about using this utility class in handwritten recipes.

src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        JavaVisitor<ExecutionContext> v = new JavaVisitor<ExecutionContext>() {
+            @Override
+            public J visitTernary(J.Ternary ternary, ExecutionContext executionContext) {
+                J.Ternary t = (J.Ternary) super.visitTernary(ternary, executionContext);
+                Expression condition =
+                        SimplifyConstantIfBranchExecution.cleanupBooleanExpression(
+                                t.getCondition(),
```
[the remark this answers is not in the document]
JLLeitschuh: This is definitely wrong @timtebeek

src/main/java/org/openrewrite/staticanalysis/BufferedWriterCreation.java
JLLeitschuh: When you have multiple before and afters, and you want to run other visitors afterwards, it makes sense

src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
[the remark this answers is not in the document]
timtebeek: Well that's disappointing; thanks for calling it out!
The underlying patch file looks ok-ish to me at first glance
```diff
diff --git a/src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java b/src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
index d98e472..41983c2 100644
--- a/src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
+++ b/src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
@@ -46,18 +46,18 @@ public class SimplifyConstantTernaryExecution extends Recipe {
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         JavaVisitor<ExecutionContext> v = new JavaVisitor<ExecutionContext>() {
             @Override
-            public J visitTernary(J.Ternary ternary, ExecutionContext executionContext) {
-                J.Ternary t = (J.Ternary) super.visitTernary(ternary, executionContext);
+            public J visitTernary(J.Ternary ternary, ExecutionContext ctx) {
+                J.Ternary t = (J.Ternary) super.visitTernary(ternary, ctx);
                 Expression condition =
                         SimplifyConstantIfBranchExecution.cleanupBooleanExpression(
                                 t.getCondition(),
                                 getCursor(),
-                                executionContext
+                                ctx
                         );
                 if (J.Literal.isLiteralValue(condition, true)) {
-                    return autoFormat(t.getTruePart(), executionContext);
+                    return autoFormat(t.getTruePart(), ctx);
                 } else if (J.Literal.isLiteralValue(condition, false)) {
-                    return autoFormat(t.getFalsePart(), executionContext);
+                    return autoFormat(t.getFalsePart(), ctx);
                 }
                 return t;
             }
```
So then it must be https://github.com/googleapis/code-suggester that's somehow thrown off. Unless you spot anything else in the diff above.

src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
JLLeitschuh: Not if I want to include the call to `cleanupBooleanExpression`

src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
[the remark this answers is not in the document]
JLLeitschuh: Looks otherwise fine. Sounds like a fun bug to fix

src/test/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecutionTest.java
```diff
@@ -0,0 +1,79 @@
...
+              class Test {
+                  void test() {
+                      System.out.println((true) ? "foo" : "bar");
+                  }
+              }
+              """,
+            """
+              class Test {
+                  void test() {
+                      System.out.println("foo");
```
JLLeitschuh, line 72: @knutwannheden this test is failing and is currently generating a change that looks like this:

```java
System.out.println( "foo");
```
instead of the expected:
```java
System.out.println("foo");
```

Same issue with the test above. Not sure what's going on here. Is this something broken with the `AutoFormat` visitor?

knutwannheden: Could be. I can investigate.

knutwannheden: The "problem" here is that what gets auto-formatted is the true or false branch of the ternary only. In an of itself it isn't incorrectly formatted and removing the whitespace prefix would require checking what it is contained in.

While it might be possible to change how auto-formatting works (although that currently seems quite difficult, since our cursor doesn't know which "feature" of the containing element is currently being visited), I think we should here instead just format the containing element. I've updated the PR accordingly.

src/main/java/org/openrewrite/staticanalysis/SimplifyConstantTernaryExecution.java
knutwannheden: I think if we were to make sure the existing `SimplifyBooleanExpressionVisitor` can handle this case, then we should be fine, as we have the dedicated `SIMPLIFY_BOOLEANS` embedding option for that.

timtebeek: Picked up in
- https://github.com/openrewrite/rewrite/pull/4617

timtebeek:
Thanks both! Took me a little while to get to, but I've pushed up various improvements to rewrite, rewrite-templating, and rewrite-static-analysis to narrow down the changes here to just a regular Refaster recipe, such that other recipes can also benefit from those improvements:

- https://github.com/openrewrite/rewrite/pull/4617
- https://github.com/openrewrite/rewrite-templating/pull/114

timtebeek (approved):

timtebeek:
Also added to Java security best practices: https://github.com/openrewrite/rewrite-java-security/commit/9d9c9ffff82ffc680059ae8282e02e79c1694dc2
Was doubting if this should be added into the OWASP top ten as well; leave that up to you @JLLeitschuh , and thanks again!

Issue #260, opened 2024-02-19
FixStringFormatExpressions wrongly replaces `\\n`


Bananeweizen:
## What is the smallest, simplest way to reproduce the problem?

```java
class A {
    void foo(String bar) {
       return """
                  \\n
                  """.formatted(bar);
    }
}
```

## What did you expect to see?
the code remaining as is.

## What did you see instead?

```java
class A {
    void foo(String bar) {
        return """
                   \%n
                   """.formatted(bar);
    }
}
```
That's an invalid escape sequence. Root cause of the bug is the recipe replacing `\n` without ever checking if that's part of `\\n`.

Be aware that even though the test _looks_ like it contains the double backslash in the test code, it doesn't. The text block contains double backslash, but that only generates the Java test input, which contains a single backslash.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes. But too tired to look into that now.

timtebeek:
Thanks for the report @Bananeweizen ! I've pushed a quick attempt up at
- https://github.com/openrewrite/rewrite-static-analysis/pull/261

Could use a review as there's a ton of slashes involved. 😅

PR #261, opened 2024-02-20
Do not escape newline in FixStringFormatExpressions

timtebeek:
Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/260
- https://github.com/openrewrite/rewrite-static-analysis/issues/260


timtebeek:
I've updated the test to show with 1, 2, 3 and 4 backslashes; I don't see the issue you've mentioned with the results being different between odd and even. Feel free to clarify with an updated test if I missed anything, and thanks again!

Bananeweizen:
@timtebeek The new test is fine and nothing is broken. What I meant is: In your new test example it would be okay to also apply the recipe in line 124 of the test, leading to every second line of the progression being updated. As long as the literal can be split into nonbackslash + `\\` + ... + `\\` + `\n`, replacing the newline is okay.
But really, we don't need to improve the code, this is quite esoteric. :)

timtebeek:
ah cheers, yes that makes sense, but indeed I'd hope no one is relying on us to clean that up 😅

PR #265, opened 2024-03-01
`InstanceOfPatternMatch`: handle same type for multiple operands

rickie:
## What's changed?

This introduces a failing unit test for the `InstanceOfPatternmatch` recipe. 
We are giving the same name to the variable while we have an expression with two operands that check the same type.

## What's your motivation?

The code in the "after" template is incorrect as it gives the following error: 
```
Variable 's' is already defined in the scope
```

## Any additional context

I'm curious how we can fix this. It doesn't seem trivial, so probably I don't have time right now to create a fix myself.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/InstanceOfPatternMatchTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/InstanceOfPatternMatchTest.java
@@ -791,5 +791,32 @@ class InstanceOfPatternMatchTest implements RewriteTest {
               )
             );
         }
+
+        @Test
+        void multipleCastsInDifferentOperands() {
+            rewriteRun(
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for the report! I _think_ this is a variant of this case which contains some earlier attempts
- https://github.com/openrewrite/rewrite-static-analysis/pull/175

Definitely something we'd want to see resolved.

timtebeek:
Exploring a likely fix in
- https://github.com/openrewrite/rewrite/pull/4066

timtebeek:
I've merged this test into #175 ; let's continue there.

rickie:
Thanks Tim 😄! 

I should've noticed the other issue before opening this one...

PR #268, opened 2024-03-04
Skip RenamePrivateFieldsToCamelCase when using lombok

timtebeek:
Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/267

Mirrors what we do in `RemoveUnusedPrivateFields`
https://github.com/openrewrite/rewrite-static-analysis/blob/106421ad4d1db3fe062c8743d042627e07598a8c/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateFields.java#L42

@blipper


blipper:
Weird.

It seems to have lost my comments

timtebeek:
Thanks for the suggestion! Think that would make a worthwhile addition, but due to time constraints I'm merging as is for now, such that we do no harm in cases where we should not make a change. We can restore more selective changes in a followup PR both here in `RenamePrivateFieldsToCamelCase` and in `RemoveUnusedPrivateFields`.

Issue #278, opened 2024-03-18
NullPointerException in CombineSemanticallyEqualCatchBlocks recipe


Marklinzi:
## What version of OpenRewrite are you using?

I am using

- Maven plugin v5.25.0
- rewrite-static-analysis  v1.4.0

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a multi module project.
```xml
<plugin>
  <groupId>org.openrewrite.maven</groupId>
  <artifactId>rewrite-maven-plugin</artifactId>
  <version>5.25.0</version>
  <configuration>
    <activeRecipes>
      <recipe>org.openrewrite.staticanalysis.CombineSemanticallyEqualCatchBlocks</recipe>
    </activeRecipes>
    <dependencies>
        <dependency>
          <groupId>org.openrewrite.recipe</groupId>
          <artifactId>rewrite-static-analysis</artifactId>
          <version>1.4.0</version>
        </dependency>
      </dependencies>
  </configuration>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
class Test {
    void method() {
        try {
        } catch (A ex) {
            String s = "foo";
        } catch (B ex) {
            String s = "foo";
        }
    }
}
```

## What did you expect to see?

```java
class Test {
    void method() {
        try {
        } catch (A | B ex) {
            String s = "foo";
        }
    }
}
```

## What did you see instead?

A NullPointerException.

## What is the full stack trace of any errors you encountered?

```
org.openrewrite.internal.RecipeRunException: java.lang.NullPointerException: Cannot invoke "org.openrewrite.java.tree.Space.getComments()" because "space1" is null
	at org.openrewrite.TreeVisitor.visit(TreeVisitor.java:334)
	at org.openrewrite.staticanalysis.CombineSemanticallyEqualCatchBlocks$CombineSemanticallyEqualCatchBlocksVisitor$CommentVisitor.visitBlock(CombineSemanticallyEqualCatchBlocks.java:560)
	at org.openrewrite.staticanalysis.CombineSemanticallyEqualCatchBlocks$CombineSemanticallyEqualCatchBlocksVisitor$CommentVisitor.visitBlock(CombineSemanticallyEqualCatchBlocks.java:292)
	at org.openrewrite.java.tree.J$Block.acceptJava(J.java:838)
	at org.openrewrite.java.tree.J.accept(J.java:59)
[...]

PR #280, opened 2024-03-18
Combine semantically equal catch blocks with same var declarations

Marklinzi:
## What's changed?

Fix NPE in CombineSemanticallyEqualCatchBlocks occurring when semantically equal catch blocks contain a variable declaration.

## What's your motivation?

Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/278.

No.

### Checklist


src/main/java/org/openrewrite/staticanalysis/CombineSemanticallyEqualCatchBlocks.java
```diff
@@ -1692,8 +1692,7 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations m
                             nullMissMatch(multiVariable.getTypeExpression(), compareTo.getTypeExpression()) ||
                             multiVariable.getVariables().size() != compareTo.getVariables().size() ||
                             multiVariable.getLeadingAnnotations().size() != compareTo.getLeadingAnnotations().size() ||
-                            doesNotContainSameComments(multiVariable.getPrefix(), compareTo.getPrefix()) ||
-                            doesNotContainSameComments(multiVariable.getVarargs(), compareTo.getVarargs())) {
+                            doesNotContainSameComments(multiVariable.getPrefix(), compareTo.getPrefix())) {
```
timtebeek, line 1695: Reading the change set and the recipe I'm wondering what made you make this change here as opposed to a nullsafe check in the `doesNotContainSameComments`. Any context towards that? At first glance it seems we now convert fewer cases even if they could be equal; was that intended or a misreading on my part?

timtebeek (commented):
Thanks a lot for covering this case with a test and working out what makes that specific case pass. A quick question as you probably have more context than I do, as I wonder about what led to this specific code change as compared to alternatives.

src/main/java/org/openrewrite/staticanalysis/CombineSemanticallyEqualCatchBlocks.java
Marklinzi: Hi, thanks for the review.
I have now reverted the change and made a nullsafe check in `doesNotContainSameComments`.

To be honest, I thought that the check for varargs comments is unnecessary here as we are comparing catch blocks, and I didn't think about features like anonymous classes (or similar constructions) where varargs declarations could occur.
That said, I am still not sure if there is a case where `multiVariable.getVarargs()` is not null, as the `CommentVisitor` is only used when two catch blocks are semantically equal as defined by `SemanticallyEqual.areEqual(...)` and for two anonymous class definitions, this seems to never be the case.

For example, the following two catch blocks could be combined:
``` JAVA
class Test {
    void method() {
        try {
        } catch (A ex) {
            String s = "foo";
            
            new Object() {
              public void test(String... test) {
              }
            };
        } catch (B ex) {
            String s = "foo";
            
            new Object() {
              public void test(String... test) {
              }
            };
        }
    }
}
```

However, this is currently not done.

timtebeek:
Thanks a lot for the detailed look and response to my question. Good to have this case fixed. We expect to follow up with a release early next week. Until then you can try it out from [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions)

PR #287, opened 2024-04-26
Avoid parenthesis only cleanup in SimplifyConstantIfBranchExecution

### LINKED ISSUE #286: `SimplifyConstantIfBranchExecution` unconditionally removes braces

```java
    @Test
    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/286")
    void doNotChangeParenthesisOnly() {
        rewriteRun(
          //language=java
          java(
            """
              public class A {
                  public void test() {
                      boolean b = true;
                      if (!(b)) {
                          System.out.println("hello");
                      }
                  }
              }
              """
          )
        );
    }
```

No change.

## What did you see instead?
Braces removed around `b`. `SimplifyConstantIfBranchExecution` unconditionally calls further visitors to optimize away also braces, unneeded block statements etc. I can understand that this makes sense for the case that it actually can simplify an if-condition before or after calling those other visitors. But if it cannot do any optimization on its own, then returning the result of brace removal is wrong. I have disabled the recipe for removing braces.

```java
    expected: "public class A {
        public void test() {
            boolean b = true;
            if (!(b)) {
                System.out.println("hello");
            }
        }
    }"
     but was: "public class A {
        public void test() {
            boolean b = true;
            if (!b) {
                System.out.println("hello");
            }
        }
    }"
```

Maybe. I'm not sure of a good way to differentiate the two cases yet.

Bananeweizen:
## What's changed?
Recognize when the recipe cannot optimize to a constant condition and return the if-condition as before all attempted optimizations.

## What's your motivation?
Fixes #286.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/SimplifyConstantIfBranchExecution.java
+++ b/src/main/java/org/openrewrite/staticanalysis/SimplifyConstantIfBranchExecution.java
@@ -87,6 +87,7 @@ public class SimplifyConstantIfBranchExecution extends Recipe {
         @Override
         public J visitIf(J.If if_, ExecutionContext ctx) {
             J.If if__ = (J.If) super.visitIf(if_, ctx);
+            J.If ifBeforeCleanup = if__;
 
             J.ControlParentheses<Expression> cp = cleanupBooleanExpression(if__.getIfCondition(), ctx);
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Nice and clean solution, thanks!

PR #292, opened 2024-05-10
Adjusts MethdNameCasing name and description

mike-solomon:
To hopefully be clearer as to what it does. Based on feedback from the following issue: https://github.com/openrewrite/rewrite/issues/4180

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
@@ -62,8 +62,8 @@ public class MethodNameCasing extends ScanningRecipe<List<MethodNameCasing.Metho
     @Override
     public String getDescription() {
         return "Fixes method names that do not follow standard naming conventions. " +
-                "For example, `private String getFoo_bar()` would be adjusted to `private String getFooBar()` " +
-                "and `public int DoSomething()` would be adjusted to `private int doSomething()`.";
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
```diff
@@ -56,12 +56,14 @@ public class MethodNameCasing extends ScanningRecipe<List<MethodNameCasing.Metho
...
-        return "Method name casing";
+        return "Standardize method name casing";
     }
 
     @Override
     public String getDescription() {
-        return "Method names should comply with a naming convention.";
+        return "Fixes method names that do not follow standard naming conventions. " +
+                "For example, `private String getFoo_bar()` would be adjusted to `private String getFooBar()` " +
+                "and `public int DoSomething()` would be adjusted to `private int doSomething()`.";
```
timtebeek, line 66: The second line was using a mix of `public` and `private`, so to prevent confusion I'd thought to just strip modifiers.
```suggestion
                "For example, `String getFoo_bar()` would be adjusted to `String getFooBar()` " +
                "and `int DoSomething()` would be adjusted to `int doSomething()`.";
```

mike-solomon: Whoops good catch - thanks.

timtebeek (approved):

PR #293, opened 2024-05-13
`RemoveUnusedPrivateMethods` should ignore class annotated with `@SupressWarning("unused")`

### LINKED ISSUE #294: org.openrewrite.staticanalysis.RemoveUnusedPrivateMethods ignores class level @SuppressWarnings("unused")
6.6.3

```java
@SuppressWarnings("unused")
class A {
    @SuppressWarnings("unused")
    private static void foo() {
    }

    private void bar() {
    }

}
```

```java
@SuppressWarnings("unused")
class A {
    @SuppressWarnings("unused")
    private static void foo() {
    }

    private void bar() {
    }

}

```

## What did you see instead?

```java
@SuppressWarnings("unused")
class A {
    @SuppressWarnings("unused")
    private static void foo() {
    }

}
```

The looks like that https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java#L62 doesn't look at the class level annotation and is also just assuming that any class annotation means don't do anything.  IT should look for     @SuppressWarnings("unused") at the class level.

```
stacktrace output here
```

Dinozavvvr:
Fix #294
- #294

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java
@@ -15,8 +15,6 @@
  */
 package org.openrewrite.staticanalysis;
 
-import java.util.*;
-
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
@@ -28,6 +26,9 @@ import org.openrewrite.java.service.AnnotationService;
 import org.openrewrite.java.tree.*;
 
 import java.time.Duration;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 
 public class RemoveUnusedPrivateMethods extends Recipe {
 
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java
@@ -15,8 +15,8 @@
  */
 package org.openrewrite.staticanalysis;
 
-import java.util.HashMap;
-import java.util.Map;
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot for the start @Dinozavvvr ! As indicated in the issue we'll likely want to add an override of `visitClassDeclaration`, that uses `FindAnnotations` to [find the suppress annotation on the class](https://github.com/openrewrite/rewrite/blob/9f634a3bf3385dd127e7cc15d54836b1a25ae578/rewrite-java/src/main/java/org/openrewrite/java/search/FindAnnotations.java#L38), inspects the value, and if that contains `unused` then do not call `super.visitClassDeclaration`, but return the original argument classDeclaration. That'll be in between these lines in the recipe itself.
https://github.com/openrewrite/rewrite-static-analysis/blob/b7b1bd1fd82c5e713d7ea5777375c6daf6eb9067/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java#L55-L57

Dinozavvvr:
@timtebeek thank you a lot for your suggestion!

Dinozavvvr:
@timtebeek I found that solution based on visiting J.ClassDeclaration ignores nested classes, so I moved this check to visitMethodDeclaration. Is there another better way to do so or maybe I'm wrong please correct

Dinozavvvr:
Did it

timtebeek:
Great to see you've worked this out @Dinozavvvr ! I've added a small polishing commit to better fit in with our code conventions. Seems like a handy pattern to limit recipe execution by looking for suppressed warnings. We might just want to apply that in a few more cases, or pull this up to a higher level even. Neat stuff!

timtebeek (approved):

Issue #289, opened 2024-05-09
DefaultComesLast recipe creates duplicate case 


delanym:
If default comes before a case and simply continues into that case
```java
      default:
      case DERIVATION_MODE:
        System.out.println("default");
        break;
```
refactoring with `org.openrewrite.staticanalysis.DefaultComesLast` from `org.openrewrite.recipe:rewrite-static-analysis:1.6.0` results in a duplicate case and code duplication
```java
      case DERIVATION_MODE:
        System.out.println("default");
        break;
      case DERIVATION_MODE:
      default:
        System.out.println("default");
```

delanym:
I can't promise anything - we've just migrated JDK from 11 to 21 so my hands are full. I guess its always like that though.
What I feel compelled to do is write the ticket.

nbenjamin:
Hi @timtebeek , I have created a PR to address this issue and added the testcase. could you please review this

PR #295, opened 2024-05-19
fix(DefaultComesLast)[#289]: duplicate case fix

nbenjamin:
## What's changed?
Fix for DefaultComesLast recipe creates duplicate case
- #289

## What's your motivation?
Bug fix


src/test/java/org/openrewrite/staticanalysis/DefaultComesLastTest.java
```diff
@@ -264,4 +264,39 @@ boolean foo() {
...
+              """,
+            """
+              class Test {
+                  int n;
+                  boolean foo() {
+                      switch (n) {
+                          case DERIVATION_MODE:
+                            System.out.println("default");
+                            break;
+                          default:
```
timtebeek, line 294: Perhaps it's too early for me, but shouldn't the behavior preserving change due to fall through be this instead?
```suggestion
                          case DERIVATION_MODE:
                          default:
                            System.out.println("default");
                            break;
```

timtebeek (commented):
Thanks a lot for taking this on! I have a slight worry about the test case, and with that the logic that we have now. Would you agree we need to assert a different outcome @delanym ?
- #289

src/test/java/org/openrewrite/staticanalysis/DefaultComesLastTest.java
nbenjamin: So you are suggesting the statements and break from case should move to default?, I am trying to understand the actual response of this. The statements & break belongs to case should be always part of that case right? It would be great if you can provide use/test case of this

delanym: @nbenjamin whatever the outcome, the recipe should not change the behaviour. In your test the default case no longer prints to stdout. The DERIVATION_MODE case should rather fall through to the default case, which holds the action.

timtebeek: Indeed what @delanym also says: we should not change the behavior; only put the default case last. In this case the test you had added was asserting a behavior change; I've just pushed up my suggestion that asserts what I think should be the right replacement here, and we'd need to get the recipe implementation to line up.

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -75,7 +75,7 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
                     aCase.getStatements().isEmpty() && !foundNonEmptyCase) {
                     casesGroupedWithDefault.add(aCase);
                 } else {
-                    if (defaultCase != null && defaultCase.getStatements().isEmpty() && !foundNonEmptyCase) {
+                    if (defaultCase != null && defaultCase.getStatements().isEmpty() && foundNonEmptyCase) {
```
punkratz312, line 78:
```suggestion
                    if (nonNull(defaultCase) && isEmpty(defaultCase.getStatements()) && foundNonEmptyCase) {
```
https://stackoverflow.com/questions/28140193/purpose-of-objects-isnull-objects-nonnull
https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isEmpty(java.util.Collection)

punkratz312 (commented):
fell free to leverage operator first speaking api

Issue #297, opened 2024-05-30
`InstanceOfPatternMatch` recipe produces compiler error with generic types without type parameters


BoykoAlex:
Consider code snippet below:
```java
                      @SuppressWarnings("unchecked")
                      public static List<Map<String, Object>> applyRoutesType(Object routes) {
                          if (routes instanceof List) {
                              List<Object> routesList = (List<Object>) routes;
                              if (routesList.isEmpty()) {
                                  return Collections.emptyList();
                              }
                              if (routesList.stream()
                                            .anyMatch(route -> !(route instanceof Map))) {
                                  return Collections.emptyList();
                              }
                              return routesList.stream()
                                               .map(route -> (Map<String, Object>) route)
                                               .collect(Collectors.toList());
                          }
                          return Collections.emptyList();
                      }

```

It is transformed by the `InstanceOfPatternMatch` into:
```java
                      @SuppressWarnings("unchecked")
                      public static List<Map<String, Object>> applyRoutesType(Object routes) {
                          if (routes instanceof List routesList) {
                              if (routesList.isEmpty()) {
                                  return Collections.emptyList();
                              }
                              if (routesList.stream()
                                            .anyMatch(route -> !(route instanceof Map))) {
                                  return Collections.emptyList();
                              }
                              return routesList.stream()
                                               .map(route -> (Map<String, Object>) route)
                                               .collect(Collectors.toList());
                          }
                          return Collections.emptyList();
                      }

```

The transformed snippet show compiler error:
```
incompatible types: java.lang.Object cannot be converted to java.util.List<java.util.Map<java.lang.String,java.lang.Object>>
```
at the return statement stream.

What seems to work is when instead of `if (routes instanceof List routesList)` we have `if (routes instanceof List<?> routesList)`

PR #298, opened 2024-05-30
InstanceOfPatternMatch: Generic type without parameters

BoykoAlex:
Unit test and a possible straight-forward fix for
- #297

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
@@ -224,6 +224,21 @@ public class InstanceOfPatternMatch extends Recipe {
                     name,
                     type,
                     null));
+            JavaType.FullyQualified fqType = TypeUtils.asFullyQualified(type);
+            if (fqType != null && !fqType.getTypeParameters().isEmpty() && !(instanceOf.getClazz() instanceof J.ParameterizedType)) {
+                TypedTree oldTypeTree = (TypedTree) instanceOf.getClazz();
+
+                J.ParameterizedType newTypeTree = new J.ParameterizedType(
+                        randomId(),
+                        oldTypeTree.getPrefix(),
+                        Markers.EMPTY,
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
```diff
@@ -224,6 +224,21 @@ public J.InstanceOf processInstanceOf(J.InstanceOf instanceOf, Cursor cursor) {
...
+                TypedTree oldTypeTree = (TypedTree) instanceOf.getClazz();
+
+                J.ParameterizedType newTypeTree = new J.ParameterizedType(
+                        randomId(),
+                        oldTypeTree.getPrefix(),
+                        Markers.EMPTY,
+                        oldTypeTree.withPrefix(Space.EMPTY),
+                        null,
+                        oldTypeTree.getType()
+                ).withTypeParameters(fqType.getTypeParameters().stream().map(p -> new J.Wildcard(randomId(), Space.EMPTY, Markers.EMPTY, null, null)).collect(Collectors.toList())
```
timtebeek, line 238: Note to self : look at listutils to conform with code elsewhere.

timtebeek:
Thanks a lot for the thorough fix here! I'm still at Spring IO, but will merge this when not reviewing on mobile. :)

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
BoykoAlex: `ListUtil.map()` is for the same type lists, i.e. mapping of `List<J>` -> `List<J>` in my case it is `List<JavaType>` -> `List<J>`. I felt it is okay... the list can't be long... I only care about its size. If there are 5 type parameters i wanted to produce `<?,?,?,?,?>. I can rerwrite this piece to conform to what i mentioned above better rather than have `map`

timtebeek:
Thanks a lot for the immediate fix @BoykoAlex !

PR #190, opened 2023-10-13
RemoveEmptyJavaDocParameters should remove "empty" lines #98

alixina9:
## What's changed?

I added fixes for issue #98  regarding removing trailing empty lines and other fixes . Also updated tests and added new ones with the expected cases of what I thought made more sense.

 ## What's your motivation? 

Issue #98 

 ## Anyone you would like to review specifically? 

@timtebeek maybe 🙂 

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #300, opened 2024-06-04
Remove empty lines after rewrite

### LINKED ISSUE #119: JavaDoc: RemoveJavaDocAuthorTag and RemoveEmptyJavaDocParameters should remove empty JavaDoc after cleanup

koppor:
- Fixes #119 
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/98

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParametersTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParametersTest.java
@@ -168,6 +168,38 @@ class RemoveEmptyJavaDocParametersTest implements RewriteTest {
         );
     }
 
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/98")
+    void removeTrailingEmptyLines() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+               class Test {
+                   /**
+                    * Text text text
+                    *
+                    * @param arg0
+                    * @param arg1
+                    *
+                    */
+                    void method(int arg0, int arg1) {
+                    }
+                }
+                """,
+            """
+               class Test {
+                   /**
+                    * Text text text
+                    */
+                    void method(int arg0, int arg1) {
+                    }
+                }
+                """
+          )
+        );
+    }
+
     @Nested
     class NoSpace {
         @Test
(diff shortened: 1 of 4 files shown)


koppor:
I left all commits, because there will be a squash-merge, won't it?

koppor:
I did not see https://github.com/openrewrite/rewrite-static-analysis/pull/190 while implementing. Think, my code is shorter then the one suggested at https://github.com/openrewrite/rewrite-static-analysis/pull/190.

src/test/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParametersTest.java
```diff
@@ -357,14 +404,6 @@ void singleLineReturnNoSpace() {
...
                 """
                   class Test {
-                      /**@return*/
-                      int method() {
-                      }
-                  }
-                  """,
-                """
-                  class Test {
-                      /***/
```
timtebeek, line 367: With these lines removed the test no longer verifies anything; is this a case that was hard to cover? Should the test be removed instead?

src/test/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParametersTest.java
```diff
@@ -380,14 +419,6 @@ void singleLineThrowsNoSpace() {
...
                 """
                   class Test {
-                      /**@throws*/
-                      void method() throws IllegalStateException {
-                      }
-                  }
-                  """,
-                """
-                  class Test {
-                      /***/
```
timtebeek, line 390: Similar to above: we no longer seem to assert anything meaningful here; especially not in relation to the test name. Should the test be restored or removed?

timtebeek (commented):
Thanks a lot for taking this on! #190 had indeed lingered as there were a lot of formatting changes, making it a challenge to review. That earlier PR did have explicit tests for `@throws` and `@return`, which seem to have been removed here. Should any of that make it in here?

I'm inclined to proceed with this PR, as I like the minimal approach you took at making changes, such that it's easier to spot what goes into removing empty lines. Let's get the comments resolved and get this fix in.

src/test/java/org/openrewrite/staticanalysis/RemoveEmptyJavaDocParametersTest.java
koppor: TBH, I totally forgot, why I removed these. Maybe by accident?

Issue #301, opened 2024-06-11
`UseStringReplace` should not rewrite when special replacement string


protocol7:
`\` and `$` in the replacement string of `String.replaceAll()` has special meaning (https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/util/regex/Matcher.html#replaceAll(java.lang.String)) and should not be rewritten.

## What version of OpenRewrite are you using?
- org.openrewrite:rewrite-core:jar:8.27.4
- org.openrewrite.recipe:rewrite-static-analysis:jar:1.9.0

## What is the smallest, simplest way to reproduce the problem?
```
import static org.openrewrite.java.Assertions.java;

import org.junit.jupiter.api.Test;
import org.openrewrite.java.JavaParser;
import org.openrewrite.staticanalysis.UseStringReplace;
import org.openrewrite.test.RecipeSpec;
import org.openrewrite.test.RewriteTest;

public class KnownBugsTest implements RewriteTest {

  @Override
  public void defaults(RecipeSpec spec) {
    spec.parser(JavaParser.fromJavaVersion());
  }

  @Test
  public void replaceAllBackslashReplacement() {
    rewriteRun(
        spec -> spec.recipe(new UseStringReplace()),
        java(
            // replacement strings with \ escapes output characters and should not be rewritten
            """
            package com.helloworld;

            class Hello {
              public String hello() {
                return "abc".replaceAll("b", "\\\\\\\\");
              }
            }
            """));
  }

  @Test
  public void replaceAllReferenceReplacement() {
    rewriteRun(
        spec -> spec.recipe(new UseStringReplace()),
        java(
            // replacement strings with $ references capture groups and should not be rewritten
            """
            package com.helloworld;

            class Hello {
              public String hello() {
                return "abc".replaceAll("b", "$0");
              }
            }
            """));
  }
}

```

timtebeek:
Thanks for the clear runnable bug report @protocol7 ! Should fit right into [UseStringReplaceTest.java](https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/test/java/org/openrewrite/staticanalysis/UseStringReplaceTest.java). Would you be open to creating a draft pull request using just these failing tests?

Looks like we could then next add a new conditional that checks there's no special characters in the replacement string, and if so skip making changes and `return invocation` early.
[...]

PR #306, opened 2024-06-29
Do not rewrite String.replaceAll with special chars in replacement string

protocol7:
If the replacement string of String.replaceAll contains $ or \, we should not rewrite it as these indicate special replacements: https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/util/regex/Matcher.html#replaceAll(java.lang.String)

- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/301


timtebeek (approved):
Thanks for taking this on @protocol7 ! I've expanded on your work just a bit to also cover the case where the second argument isn't a literal, as that variable/field/returnValue might then contain special characters. Hope you agree!

PR #310, opened 2024-07-04
Don't use lambdas for methods with type parameters.

### LINKED ISSUE #309: UseLambdaForFunctionalInterface should exclude methods with type paramters
We are using OpenRewrite v8.28.1 and rewrite-static-analysis v1.10.1.

In our codebase, we have an interface with a single method using a type parameter. Something like this:
```java
public interface I {
  <T> List<T> call();
}
```

Which is then implemented in other places using an anonymous class like so:
```java
final I i = new I() {
  @Override
  public <T> List<T> call() {...}
};
```

The `UseLambdaForFunctionalInterface` recipe replaces such inline definitions of `I` with a lambda (`final I i = () -> ...;`). That will not compile because of the type argument being lost. I would expect the recipe to bail out and not introduce a lambda in this case.

I can reproduce it using this testcase:
```java
  @Test
  void dontUseLambdaForMethodWithTypeParameter() {
    rewriteRun(
        spec -> spec.recipe(new UseLambdaForFunctionalInterface()),
        java(
            // can't transform to lambda because of the type argument of I#call()
            """
            package com.helloworld;

            import java.util.List;

            public interface I {
              <T> List<T> call();
            }

            class Hello {
              public void hello() {
                final I i = new I() {
                  @Override
                  public <T> List<T> call() {
                    return null;
                  }
                };
                final List<String> list = i.call();
              }
            }
            """));
  }
}
```

mbruggmann:
For a method that has a type parameter, we can't replace anonymous class declarations with a lambda. For example:
```java
public interface I {
  <T> List<T> call();
}

// Can't be replaced by lambda
final I i = new I() {
  @Override
  public <T> List<T> call() { ... }
};
```

The proposed fix is to check the method in question for type parameters, and if there are any, return early from `UseLambdaForFunctionalInterface` without making any changes.

Fixes #309

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UseLambdaForFunctionalInterface.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UseLambdaForFunctionalInterface.java
@@ -96,6 +96,11 @@ public class UseLambdaForFunctionalInterface extends Recipe {
                         StringBuilder templateBuilder = new StringBuilder();
                         J.MethodDeclaration methodDeclaration = (J.MethodDeclaration) n.getBody().getStatements().get(0);
 
+                        // If the functional interface method has type parameters, we can't replace it with a lambda.
+                        if (methodDeclaration.getTypeParameters() != null && !methodDeclaration.getTypeParameters().isEmpty()) {
+                            return n;
+                        }
+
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks a lot for exploring, finding and contributing the fix!

PR #337, opened 2024-09-13
Instanceof pattern: exclusion for `Stream.collect(..)`

BoykoAlex:
Support `instanceof` with types where type parameters shouldn't be present in the `instanceof`.

For example:
```java
if (o instanceof List) {
  map.put("key", String.join(",", (List) o));
}
```
currently produces a compiler error when refactored in:
```java
if (o instanceof List<?> list) {
  map.put("key", String.join(",", list));
}
```
The compiler error:
```
The method join(CharSequence, CharSequence...) in the type String is not applicable for the arguments (String, List<capture#1-of ?>)
```

Instead it should become:
```java
if (o instanceof List list) {
  map.put("key", String.join(",", list));
}
```
(i.e. without the wildcard)

The fix is if the type has type parameters and the following type casts statements cast to type with type parameters then wildcard type parameter(s) will be added. Otherwise no type parameters will be added to the type in the `instanceof`

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
@@ -21,6 +21,7 @@ import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the work done here @BoykoAlex ! Let me know when you're done by marking the PR as ready for review. :)

BoykoAlex:
Abandoned in favour of #338

PR #338, opened 2024-09-18
InstanceOf: no action if type casts with specific type params used

BoykoAlex:
If type casts to types with defined type parameters are used, i.e. `List<Object>`, `List<String>` it is difficult to determine if possible at all if `instanceof List list` or `instanceof List<?> list` isappropriate. Soem cases where type cast without type params is used and with it is impossible to determine. Therefore, I think it is best not to touch such `instanceof` <-> `type cats` patterns.


sambsnyd:
Thanks @BoykoAlex !

PR #340, opened 2024-09-19
Support fluent chains in RemoveMethodCallVisitor

### LINKED ISSUE #339: RemoveMethodCallVisitor doesn't work in fluent APIs
I am using
- OpenRewrite 8.34.4
- rewrite-static-analysis 1.15.0

The RemoveMethodCallVisitor does not seem to work in fluent APIs.
I can reproduce it using this testcase:
```java
  @Test
  void removeMethodCallFromFluentChain() {
    final MethodMatcher methodMatcher = new MethodMatcher("java.lang.StringBuilder append(..)");
    final TreeVisitor<?, ExecutionContext> visitor =
        new RemoveMethodCallVisitor<>(methodMatcher, (i, e) -> true);
    rewriteRun(
        spec -> spec.recipe(RewriteTest.toRecipe(() -> visitor)),
        // language=java
        java(
            """
            class Main {
                void hello() {
                    final String s = new StringBuilder("hello")
                            .append("world")
                            .delete(1, 2)
                            .toString();
                }
            }
            """,
            """
            class Main {
                void hello() {
                    final String s = new StringBuilder("hello")
                            .delete(1, 2)
                            .toString();
                }
            }
            """));
  }
```

RemoveMethodCallVisitor should work even in fluent chains.

## What did you see instead?
The visitor does not find any invocations except the last one in the fluent chain.

Potentially yes, but this might be an easy fix for someone familiar with the codebase.

mbruggmann:
## What's changed?
`RemoveMethodCallVisitor` can now remove method invocations in a fluent chain, assuming that the select is also a method invocation and of the same return type. In particular, this works for the builder pattern.

## What's your motivation?
- Fixes #339

Is partial support for this use-case worth the added complexity?

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveMethodCallVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveMethodCallVisitor.java
@@ -48,33 +48,55 @@ public class RemoveMethodCallVisitor<P> extends JavaIsoVisitor<P> {
     @SuppressWarnings("NullableProblems")
     @Override
     public J.@Nullable NewClass visitNewClass(J.NewClass newClass, P p) {
-        return visitMethodCall(newClass, () -> super.visitNewClass(newClass, p));
+        Supplier<J.NewClass> visitSuper = () -> super.visitNewClass(newClass, p);
+        if (!methodMatcher.matches(newClass)) {
+            return visitSuper.get();
+        }
+        J.Block parentBlock = getCursor().firstEnclosing(J.Block.class);
+        //noinspection SuspiciousMethodCalls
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the measured improvement here! Indeed would say that this is worthwhile to add into the visitor. Could you let me know when you're ready for review?

mbruggmann:
@timtebeek Cool. Feel free to have a look, I was just waiting for the tests to go through in CI.

timtebeek (approved):
Thanks for the controlled extension to what the visitor can cover! I've added some small polishing commits, but other than that good to go.

PR #341, opened 2024-09-21
Add WrapOption to OperatorWrap to change detected style easily after running once

timtebeek:
## What's changed?
Add WrapOption wrapOption to OperatorWrap, and change from a static visitor to an inline visitor. The actual code changes are a lot smaller when ignoring whitespace.

## What's your motivation?
Projects often rely on detection, or the default which matches CheckStyle. Yet when you want to change or enforce a new standard, it's cumbersome to have to configure a style for every project when you just want to change what's used once, and henceforth rely on detected style again.

* Continue to ask folks to configure an explicit style.
* Accept the CheckStyle default of operators on next line.
* As discussed on https://github.com/openrewrite/rewrite/pull/4509#pullrequestreview-2319707088 we could also make it easier to provide one-off override of detected style, to prevent proliferation of one-off-override-flags.

## Any additional context
Discovered that we had a duplicate still in openrewrite/rewrite:
- https://github.com/openrewrite/rewrite/pull/4509

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/OperatorWrap.java
+++ b/src/main/java/org/openrewrite/staticanalysis/OperatorWrap.java
@@ -15,6 +15,8 @@
  */
 package org.openrewrite.staticanalysis;
 
+import lombok.EqualsAndHashCode;
+import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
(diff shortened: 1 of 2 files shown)


timtebeek:
With a quick local run this indeed does what I'd hoped it would do on openrewrite/rewrite
image

Perfect then for inclusion into our [OpenRewrite best practices](https://github.com/openrewrite/rewrite-recommendations/blob/010e7f97c9922ca0a4ccea1efee0b4abaaac9145/src/main/resources/META-INF/rewrite/openrewrite.yml#L17-L21), for enforcement on PRs.

knutwannheden (approved):

PR #348, opened 2024-10-02
Fix: `CatchClauseOnlyRethrows` does not handle multi catch correctly

nielsdebruin:
## What's changed?
The CatchClauseOnlyRethrows will now correctly handle multicatch blocks that use the short hand syntax `catch(Foo | Bar exception) {...}`.

## What's your motivation?

In issue https://github.com/openrewrite/rewrite/issues/4235, a bug was demonstrated that caused a `catch` statement to be incorrectly removed as a wider exception is specified later on. I found that the reason for this behavior is the incorrect handling of `catch(Foo | Bar exception)` when checking for a wider exception type. The current code only performs a direct comparison between the parameter type of the current `catch` statement and that of the next one. This works fine if the statement is of the form `catch(Foo exception)`. However, in the case where the shorthand `catch(Foo | Bar exception)` is used, direct comparison of their parameter type no longer works. I added addition logic that will also correctly handle the check for`JavaType.MultiCatch` types.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
+++ b/src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
@@ -77,7 +77,10 @@ public class CatchClauseOnlyRethrows extends Recipe {
                         // keep this one
                         for (int j = i + 1; j < tryable.getCatches().size(); j++) {
                             J.Try.Catch next = tryable.getCatches().get(j);
-                            if (!onlyRethrows(next) && hasWiderExceptionType(aCatch, next)) {
+                            if (hasWiderExceptionType(aCatch, next)) {
+                                if (onlyRethrows(next)) {
+                                    return null;
+                                }
                                 return aCatch;
                             }
                         }
--- a/src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
+++ b/src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
@@ -90,7 +90,12 @@ public class CatchClauseOnlyRethrows extends Recipe {
             private boolean hasWiderExceptionType(J.Try.Catch aCatch, J.Try.Catch next) {
                 if (next.getParameter().getType() instanceof JavaType.MultiCatch) {
                     JavaType.MultiCatch multiCatch = (JavaType.MultiCatch) next.getParameter().getType();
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
```diff
@@ -88,6 +87,14 @@ public J.Try visitTry(J.Try tryable, ExecutionContext ctx) {
                 }));
             }
 
+            private boolean hasWiderExceptionType(J.Try.Catch aCatch, J.Try.Catch next) {
+                if (next.getParameter().getType() instanceof JavaType.MultiCatch) {
+                    JavaType.MultiCatch multiCatch = (JavaType.MultiCatch) next.getParameter().getType();
+                    return multiCatch.getThrowableTypes().stream().anyMatch(alt -> TypeUtils.isAssignableTo(alt, aCatch.getParameter().getType()));
```
Laurens-W, line 93: We may want to use a foreach loop here as this recipe potentially gets ran a lot

knutwannheden:
While there are still other improvements that are possible, I think they are out of scope for this PR. But I do get the impression that there is some sort of regression. The old implementation would have removed the first two catch blocks here (or also just the first one if you remove the middle one), whereas the new implementation doesn't seem to do that:
```java
              class A {
                  void foo() throws IOException {
                      try {
                          new FileReader("").read();
                      } catch (FileNotFoundException e) {
                          throw e;
                      } catch (IOException e) {
                          throw e;
                      } catch (Exception | Throwable e) {
                          throw e;
                      }
                  }
              }
```

nielsdebruin:
Hi @knutwannheden, thank you for the review! You are right. I updated the code and modified the order of the if statement to remedy the regression you mentioned. The modification, will now cause the first catch for `FileNotFoundException` in your example to be correctly deleted. However, it does preserve the `IOException`, which could be deleted. The cause behind this is that the `onlyRethrows` method is not able to handle a multicatch, and I am not sure how to fix this...I believe this PR will thus prevent code that should not be removed from being removed (as was the original issue). However, it might not remove all the code that can be removed, for that `onlyRethrows` need to be correctly modified (which the current implementation does not support either). As you mentioned perhaps a second issue/pr should be created for this.

knutwannheden:
@nielsdebruin Thanks a lot for the improvement!

PR #361, opened 2024-10-12
Revise inline variable

punkratz312:
# Inline Code for SRP and Clean Code Principles

This document outlines the **Single Responsibility Principle (SRP)** and clean code principles, emphasizing a compact writing style.

### Extended Guideline

In addition to existing guidelines, we propose another rule: **Inline single-use variables**. When a variable is declared but only used once, it adds unnecessary clutter. Therefore, we recommend inlining such variables to promote cleaner and more readable code. The goal is to streamline the implementation to be direct and simple.

### Code Examples

Here are some examples demonstrating these principles:

```javascript
const calculateArea = () => {
    const length = 1; 
    const width = 2; 
    const area = length + width; 
    return area; 
};

const calculateAreaWithParams = () => {
    const length = 1; 
    const width = 2; 
    return length + width; 
};

const calculateAreaInline = () => {
    return 1 + 2; 
};

// Improved version with inlined single-use variables
const calculateAreaImproved = () => {
    return 1 + 2; // Length and width inlined
};

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InlineVariable.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InlineVariable.java
@@ -34,6 +34,7 @@ import static java.time.Duration.ofMinutes;
 import static java.util.Collections.singleton;
 import static java.util.Objects.requireNonNull;
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
+import static org.apache.commons.lang3.ObjectUtils.isEmpty;
 import static org.openrewrite.internal.ListUtils.concatAll;
 import static org.openrewrite.internal.ListUtils.map;
 
@@ -80,13 +81,14 @@ public class InlineVariable extends Recipe {
                         (VariableDeclarations) statements.get(statements.size() - 2)) : null;
     }
 
-    private static @Nullable Block getVariableDeclarationsBlock(List<Statement> statements, Block bl,
-                                                                final VariableDeclarations varDec) {
+    private static @Nullable Block getVariableDeclarationsBlock(List<Statement> statements,
+                                                                Block bl,
+                                                                VariableDeclarations varDec) {
         NamedVariable identDefinition = varDec.getVariables().get(0);
-        if (varDec.getLeadingAnnotations().isEmpty() && identDefinition.getSimpleName().equals(identReturned(statements.get(statements.size() - 1)))) {
-            return replaceStatements(bl, statements, identDefinition, varDec);
-        }
-        return null;
+        return isEmpty(varDec.getLeadingAnnotations())
+                && identDefinition.getSimpleName().equals(identReturned(statements.get(statements.size() - 1)))
+                ? replaceStatements(bl, statements, identDefinition, varDec)
+                : null;
     }
 
     private static Block replaceStatements(Block bl, List<Statement> statements,
--- a/src/main/java/org/openrewrite/staticanalysis/InlineVariable.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InlineVariable.java
@@ -88,12 +88,12 @@ public class InlineVariable extends Recipe {
         return null;
     }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/main/java/org/openrewrite/staticanalysis/InlineVariable.java [hunk 1]
```diff
@@ -3,7 +3,7 @@
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * you may obtain a copy of the License at
```
punkratz312, line 6:
```suggestion
 * You may obtain a copy of the License at
```

timtebeek:
hi @punkratz312 ; Could you explain why you're refactoring `InlineVariable` here? I see this adds external dependencies like `ObjectUtils` and extracts a lot of small static methods, which I'm not sure are a clear improvement. I also don't see any tests indicating that this unlocks additional cases as compared to above, so then I'm wondering why we're changing things here.

punkratz312:
Dear @timtebeek,

Thank you for your reply. I'd like to provide some context and intent behind the changes. This is primarily a refactoring effort that applies the SOLID design principles, specifically focusing on the Single Responsibility Principle (SRP). According to clean code standards, a method should perform one clear task and be limited to 3 to 5 lines, which helps maintainability.

Clean functions that abstract away implementation details are key to ensuring the code is easily accessible and refactorable. Before the refactoring, I struggled to understand what the code was doing. However, by following the extract-till-you-drop method and considering unbloated code practices, I now have a clearer understanding.

The refactoring restructures the code without altering its functionality, and while the tests themselves were not changed, they did fail when I made mistakes during the refactoring process. This demonstrates that the tests are effective at catching issues, which is a good indication of their reliability.

While I’m not an expert on the inner workings of the code, and refactoring is essentially never finished—there is always room for improvement—the current structure adheres more closely to clean code principles. It addresses feature envy violations and improves clarity. Each function now has a single purpose, following best practices, and ensuring that no class, method, or code fragment does more than necessary.

Best Regards

punkratz312:
Still, this refactoring was just a way for me to better understand the class, as my actual goal was to adjust and copy the rule to fit my intended purpose mentioned earlier. Initially, I thought this rule would be quite similar to my needs, and that a tiny adjustment would suffice—but I was wrong. Even with over a decade of experience, this turned out to be more complex than expected. This is clearly high-end code, and without any proper introduction, it’s very difficult to grasp such complex coding, especially when it’s left unstructured as one big block, which imposes a heavy mental load.

punkratz312:
The external dependency on Apache Commons is not a problem, as it is obviously common practice. That’s why it’s included in this package and is already available in almost every codebase. Even if it’s not directly related to this utility, Spring has the same functionality. If you want to change the import, I will adjust accordingly. This approach makes the code easier to read because you see the operator or predicate first, which is an implementation detail but in this case it matters. When it comes to these details, it’s beneficial to write the simplest and most direct form possible. This can be achieved through a one-time static import, which enhances code quality and allows the predicate/function to come first instead of last, emphasizing the "first things first" principle, if you know what I mean.

timtebeek:
Hi @punkratz312 ; While I'm appreciative of your efforts to refactor this one recipe, we disagree on whether this constitutes an improvement. We've doubled the number of lines of code, added an external dependency, and made this recipe implementation diverge from others in terms of style. Folks now have to trace the flow through various methods, as opposed to at a glance seeing a reduction in scope based on nesting.

While some middle ground can perhaps be found in how much code we put into a single method, I'm more inclined to consider such efforts when they are tied to new development, as opposed to pure refactoring, as right now I'd have to carefully review change without any functional benefit. I propose we close this effort for now.

punkratz312:
Yes, I understand that many developers don't recognize the benefits of clean code or separation of concerns, so spaghetti code like this is unfortunately a common practice. However, it's actually a clear anti-pattern and often just laziness, with the code not being cleaned up after its initial creation.

Issue #362, opened 2024-10-14
Fix PMD rule `LiteralsFirstInComparisons` for `compareTo*` and `contentEquals`


punkratz312:
Simply switch the arguments; that should be a very small change to make, just similar to the Inline Variable recipe.

## What problem are you trying to solve?
https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#literalsfirstincomparisons

## Describe the situation before applying the recipe

```java
class A {
    boolean bar(String x) {
        return x.equals("2"); // should be "2".equals(x)
    }
}
```

## Describe the situation after applying the recipe

```java
class A {
    boolean bar(String x) {
        return "2".equals(x);
    }
}
```
## Further examples from the PMD:
```java
class Foo {
    boolean bar(String x) {
        return x.equals("2"); // should be "2".equals(x)
    }
    boolean bar(String x) {
        return x.equalsIgnoreCase("2"); // should be "2".equalsIgnoreCase(x)
    }
    boolean bar(String x) {
        return (x.compareTo("bar") > 0); // should be: "bar".compareTo(x) < 0
    }
    boolean bar(String x) {
        return (x.compareToIgnoreCase("bar") > 0); // should be: "bar".compareToIgnoreCase(x) < 0
    }
    boolean bar(String x) {
        return x.contentEquals("bar"); // should be "bar".contentEquals(x)
    }
}
```

punkratz312:
I would love to volunteer and implement it if you could just provide me with the snippet to flip the arguments. It should be very similar to my variant of rule 1488, InlineVariable. https://docs.openrewrite.org/recipes/staticanalysis/inlinevariable.

knutwannheden:
For the `equals()` case there is already a recipe here: https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java. The others are not covered as far as I know.

timtebeek:
Thanks for the offer to help @punkratz312 ! As Knut has indicated `String#equals` and `String#equalsIgnoreCase` are already handled in this existing recipe:
https://github.com/openrewrite/rewrite-static-analysis/blob/2163f0f05dd6fe7f4762e3486c8f16a9a5787036/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java#L32-L34

[...]

PR #365, opened 2024-10-14
PMD: LiteralsFirstInComparisons for compareTo* and contentEquals

punkratz312:
- tryout: https://github.com/openrewrite/rewrite-static-analysis/issues/362

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,26 +9,31 @@ description = "The first Static Analysis and REMEDIATION tool"
 
 val rewriteVersion = rewriteRecipe.rewriteVersion.get()
 dependencies {
-    annotationProcessor("org.openrewrite:rewrite-templating:${rewriteVersion}")
-    annotationProcessor("org.projectlombok:lombok:latest.release")
     compileOnly("org.projectlombok:lombok:latest.release")
+    annotationProcessor("org.projectlombok:lombok:latest.release")
+    testImplementation("org.projectlombok:lombok:latest.release")
+
+    implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
+    implementation("org.openrewrite:rewrite-java")
+    implementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
+    implementation("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
+    implementation("org.openrewrite:rewrite-csharp:${rewriteVersion}")
+    implementation("org.openrewrite.meta:rewrite-analysis:${rewriteVersion}")
+    implementation("org.apache.commons:commons-text:latest.release")
+
+    annotationProcessor("org.openrewrite:rewrite-templating:${rewriteVersion}")
+    implementation("org.openrewrite:rewrite-templating:${rewriteVersion}")
     compileOnly("com.google.errorprone:error_prone_core:2.+:with-dependencies") {
         exclude("com.google.auto.service", "auto-service-annotations")
     }
-    implementation("org.apache.commons:commons-text:latest.release")
-    implementation("org.openrewrite.meta:rewrite-analysis:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-csharp:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-java")
-    implementation("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-templating:${rewriteVersion}")
-    implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
-    testImplementation("com.google.code.gson:gson:latest.release")
-    testImplementation("junit:junit:4.13.2")
+
     testImplementation("org.jetbrains:annotations:24.+")
-    testImplementation("org.junit-pioneer:junit-pioneer:2.0.1")
     testImplementation("org.openrewrite:rewrite-groovy")
-    testImplementation("org.projectlombok:lombok:latest.release")
-    testRuntimeOnly("com.google.code.findbugs:jsr305:latest.release")
+    testImplementation("org.junit-pioneer:junit-pioneer:2.0.1")
+    testImplementation("junit:junit:4.13.2")
+
+    testImplementation("com.google.code.gson:gson:latest.release")
+
     testRuntimeOnly("org.openrewrite:rewrite-java-17")
+    testRuntimeOnly("com.google.code.findbugs:jsr305:latest.release")
 }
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -9,31 +9,26 @@ description = "The first Static Analysis and REMEDIATION tool"
...
-
-    implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
-    implementation("org.openrewrite:rewrite-java")
-    implementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-csharp:${rewriteVersion}")
-    implementation("org.openrewrite.meta:rewrite-analysis:${rewriteVersion}")
-    implementation("org.apache.commons:commons-text:latest.release")
-
     annotationProcessor("org.openrewrite:rewrite-templating:${rewriteVersion}")
```
punkratz312, line 12: Apply natural ordering to enable grouping. This was revisited due to the investigation into the Java version.
I can extract it of course.

image

timtebeek: Please try to minimize the changes when you open a PR, as that reduces the burden on the reviewer, and any chance of conflicts with parallel work.

Whenever we _do_ want to change such a grouping we typically do so in an automated fashion at scale with recipes, as opposed to manually. That then ensures we're immediately consistent across projects after such a recipe run.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitorTest.java
```diff
@@ -23,7 +23,7 @@
 import static org.openrewrite.java.Assertions.java;
 
 @SuppressWarnings({"ClassInitializerMayBeStatic", "StatementWithEmptyBody", "ConstantConditions"})
-class EqualsAvoidsNullTest implements RewriteTest {
+class EqualsAvoidsNullVisitorTest implements RewriteTest {
```
timtebeek, line 26: On line 30 we're testing the recipe though right? Perhaps best then to not change the class name.
```suggestion
class EqualsAvoidsNullTest implements RewriteTest {
```

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitorTest.java
timtebeek, line 1: Let's revert to the previous file name too.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitorTest.java
```diff
@@ -37,20 +37,26 @@ void invertConditional() {
...
+                        {
+                            String s = null;
+                            if(s.equals("test")) {}
+                            if(s.equalsIgnoreCase("test")) {}
+                            System.out.println(s.compareTo("test"));
+                            System.out.println(s.compareToIgnoreCase("test"));
+                            System.out.println(s.contentEquals("test"));
+                        }
+                    }
               """,
```
timtebeek, line 50: Let's not change the indentation here.
```suggestion
              public class A {
                  {
                      String s = null;
                      if(s.equals("test")) {}
                      if(s.equalsIgnoreCase("test")) {}
                      System.out.println(s.compareTo("test"));
                      System.out.println(s.compareToIgnoreCase("test"));
                      System.out.println(s.contentEquals("test"));
                  }
              }
              """,
```

timtebeek (changes requested):
Great to see these additions; just a couple of requests to reduce the changes before we merge, to make it easier to review.

build.gradle.kts
punkratz312: indeed, reverted,

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitorTest.java
punkratz312: Yes, but previously it was kind of broken because the test auto-discovery couldn’t function due to the missing naming convention. Now it’s working in every direction. I had to run the build to find the tests, so please avoid this pitfall for the next person.


image

image

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitorTest.java
punkratz312: If you want to encounter the pitfall again, I can do that, but I highly recommend not doing so. Additionally, the tests are currently somehow violating the Single Responsibility Principle (SRP), meaning that one test is somehow testing multiple classes, which is not intended. Normally, there should be a one-to-one relationship, not one-to-many.

timtebeek: The problem here stems from the `EqualsAvoidsNullVisitor` being separate from `EqualsAvoidsNull`. In most other cases the visitor is either an inline visitor in the recipe class, or nested static class. We could apply that pattern here as it seems the visitor is not used anywhere else.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitorTest.java
timtebeek: We have one in src/test/java which I would have thought would apply here:
https://github.com/openrewrite/rewrite-static-analysis/blob/692b242b9756b0b2290f56f8a05bfee0d95d5303/src/test/java/.editorconfig

build.gradle.kts
punkratz312: Obviously its an old java version now even 17?  `org.openrewrite:rewrite-java-17`

```
> Task :compileJava FAILED
/home/runner/work/rewrite-static-analysis/rewrite-static-analysis/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java:50: error: pattern matching in instanceof is not supported in -source 8
        if (superVisitMethodInvocation instanceof J.MethodInvocation methodInvocation) {
```

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -26,54 +27,65 @@
...
     public J visitMethodInvocation(J.MethodInvocation method, P p) {
-        J j = super.visitMethodInvocation(method, p);
-        if (!(j instanceof J.MethodInvocation)) {
-            return j;
-        }
-        J.MethodInvocation m = (J.MethodInvocation) j;
-        if (m.getSelect() == null) {
-            return m;
+        val superVisitMethodInvocation = super.visitMethodInvocation(method, p);
+        if (superVisitMethodInvocation instanceof J.MethodInvocation methodInvocation) {
```
timtebeek, line 50: We're keeping the source level at Java 8 for our friends still on older Java versions, allowing them to modernize their applications gradually.

PR #368, opened 2024-10-14
address: LiteralsFirstInComparisons PMD rule

punkratz312:
- tryout: https://github.com/openrewrite/rewrite-static-analysis/issues/362


src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -17,63 +17,96 @@
...
+        return getSuperIfSelectNull((J.MethodInvocation) super.visitMethodInvocation(method, p));
+    }
+
+    private J getSuperIfSelectNull(final J.MethodInvocation methodInvocation) {
+        return isNull(methodInvocation.getSelect()) ?
+                methodInvocation :
+                !(methodInvocation.getSelect() instanceof J.Literal)
+                        && methodInvocation.getArguments().get(0) instanceof J.Literal
+                        && isStringComparisonMethod(methodInvocation)
+                        ? handleNullSafetyCheck(methodInvocation, getCursor().getParentTreeCursor().getValue())
```
punkratz312, line 59: I am unsure of the original intent behind the current code, so please review and revise the methods or provide suggestions for improving the structure. I believe the current code is somewhat acceptable, but dividing each function into separate units would better adhere to the Single Responsibility Principle. This approach will focus on reducing redundancy, ensuring each function has a distinct and well-defined purpose, and avoiding 'feature envy' by keeping responsibilities isolated.

punkratz312:
@timtebeek Is this okay? Can we merge, please? If not, you can easily adjust the implementation and just keep the if adjustment and the fields. Everything else is just streamlining the implementation.

timtebeek:
Thanks again @punkratz312 ! I've got an early flight tomorrow and Friday (mostly) off, so it might be Monday before I get to a review and merge.

punkratz312:
Have a happy flight and enjoy your trip! Thank you for your guidance and support.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -17,63 +17,111 @@
...
+ * and performs optimizations to ensure null checks are correctly applied.
+ * <p>
+ * For more details, refer to the PMD best practices:
+ * <a href="https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#LiteralsFirstInComparisons">Literals First in Comparisons</a>
+ *
+ * @param <P> The type of the parent context used for visiting the AST.
+ */
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class EqualsAvoidsNullVisitor<P> extends JavaVisitor<P> {
```
punkratz312, line 52: Perhaps LiteralsFirstInComparisonsVisitor would be a better fit, as it's no longer just about equality; both equality and comparison are now merely implementation details.
```suggestion
public class LiteralsFirstInComparisonsVisitor<P> extends JavaVisitor<P> {
```

timtebeek: I'd hold off on renaming this just yet. While the visitor only appears to be used in this module, the recipe name is well established as part of [our Common static analysis issues recipe](https://docs.openrewrite.org/recipes/staticanalysis/commonstaticanalysis), which folks have created custom copies off that might break if we similarly rename the recipe.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -17,84 +17,123 @@
...
         private final J.Binary scope;
-        boolean done;
 
-        @Override
-        public @Nullable J visit(@Nullable Tree tree, P p) {
-            if (done) {
-                return (J) tree;
-            }
-            return super.visit(tree, p);
-        }
```
timtebeek, line 97: This visit method seems to have been deliberately added in https://github.com/openrewrite/rewrite-static-analysis/commit/a330fe97f5f04019d8cc5c87d77c351a6e88ca3d ; what made you remove it from here now?

timtebeek (commented):
I like how you've expanded the existing visitor to cover more cases, and by extension the recipe to do more similar replacements without the need for users to adopt any new recipe!

I also appreciate you're looking to apply best practices to the visitor itself, even if those diverge from what's common elsewhere throughout our code base. It'll take some getting used to on both side on what to (not) pick up as part of your work on recipes, as it does lead to more work to review, and at times discussions as to what's the best way forward.

Typically we view recipes and visitors as one complete unit; there's little need to optimize within such a unit, and at times patterns like the four-fold `requireNonNull(m.getSelect())` introduced here are a symptom of branching off separate methods, where the nested structure would have no such need. Going forward lets try to minimize such changes, or perhaps propose them first as part of a self-review such that there's no conflicts arising late in review.

I've left a final question about a removed earlier optimization; once that's answered we can merge this work; thanks again!

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
punkratz312: Was a mistake, thanks for the finding.

Actually i would be nice if this is tested so i would have known earlier.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
punkratz312: Yes thanks. Its just an idea and of course would be a new PR topic.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -97,8 +92,8 @@ private static J.Binary literalsFirstInComparisonsNull(J.MethodInvocation m, Exp
                 m.getPrefix(),
                 Markers.EMPTY,
                 requireNonNull(m.getSelect()),
-                build(J.Binary.Type.Equal).withBefore(SINGLE_SPACE),
-                firstArgument.withPrefix(SINGLE_SPACE),
+                JLeftPadded.build(J.Binary.Type.Equal).withBefore(Space.SINGLE_SPACE),
```
punkratz312, line 97: Build maybe okay, but the Space is DRY Space. SPACE 

If i need unwanted context i move my mouse and not clutter the code with redundancy, but its a style where most dev´s use full qualified imports.

image

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -66,21 +66,19 @@ public J visitMethodInvocation(J.MethodInvocation method, P p) {
...
+
+    private boolean isStringComparisonMethod(J.MethodInvocation methodInvocation) {
         return EQUALS.matches(methodInvocation) ||
-                !style.getIgnoreEqualsIgnoreCase() &&
-                EQUALS_IGNORE_CASE.matches(methodInvocation) ||
-                COMPARE_TO.matches(methodInvocation) ||
-                COMPARE_TO_IGNORE_CASE.matches(methodInvocation) ||
-                CONTENT_EQUALS.matches(methodInvocation);
-                && EQUALS_IGNORE_CASE.matches(methodInvocation)
-                || COMPARE_TO.matches(methodInvocation)
```
punkratz312: bad mixmatch, thx for the fix

punkratz312:
Thanks for the feedback and correction. I think we might got it now 👍

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
timtebeek: Often classes use a mix of the [two constants in Space](https://github.com/openrewrite/rewrite/blob/0e7364eb9d8d6b6e7f07e9da190b8fe375ff31ab/rewrite-java/src/main/java/org/openrewrite/java/tree/Space.java#L32-L34), only one of which has `_SPACE` in the name for the repetition. I'd prefer then to be consistent with prior use, and i/fwhen we dedice to change that, do so through [our enforced patterns in rewrite-recommendations](https://github.com/openrewrite/rewrite-recommendations/commit/88183fbd237bce56abc44a97def111ad9faf703a). I hope that context helps!

timtebeek:
Thanks again @punkratz312 !

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
knutwannheden: I suspect the idea with the name was to emphasize that the object represents a single space character. The class `Space` represents any sequence of whitespace and comments, so `Space.SINGLE` would be a bit ambiguous (a single space, tab, newline, or comment?), so `Space.SINGLE_SPACE` uses "space" in two different meanings.

punkratz312:
Thank you, @timtebeek, it was an honor and a great learning experience. I appreciate the support, guidance, and patience.

timtebeek:
You're more than welcome! Your change has just landed in the latest release:
https://github.com/openrewrite/rewrite-static-analysis/releases/tag/v1.19.0

PR #369, opened 2024-10-14
Empty lines test

Pankraz76:
fix line missalining to accordingly Java Code Conventions

https://www.oracle.com/technetwork/java/codeconventions-150003.pdf


timtebeek (commented):
Great to see this kicked off! I've posted a few quick suggestion even if this is still a draft. In general we try to have a single `@DocumentExample` annotated test, which we've shown in the docs before. That also helps folks spot which test to copy and modify if they want to report an `@Issue`.

PR #374, opened 2024-10-21
Cap the lengths of generate names in the ReplaceDuplicateStringLiterals recipe

lkerford:
We could generate some unreasonably long variable names. By default, we will cap variable names at 100 characters but people can change this was desired.

We will crop the generated variable name to the nearest "_" if there is one available

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java
@@ -190,11 +190,7 @@ public class ReplaceDuplicateStringLiterals extends Recipe {
                 String newNameString = newName.toString();
                 while (newNameString.length() > maxVariableLength){
                     int indexOf = newNameString.lastIndexOf("_");
-                    if (indexOf > -1) {
-                        newNameString = newNameString.substring(0, indexOf);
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java
```diff
@@ -181,7 +187,16 @@ private String transformToVariableName(String valueOfLiteral) {
                         prevIsLower = Character.isLowerCase(c);
                     }
                 }
-                return VariableNameUtils.normalizeName(newName.toString());
+                String newNameString = newName.toString();
+                while(newNameString.length() > maxVariableLength){
+                    int indexOf = newNameString.lastIndexOf("_");
+                    if(indexOf > -1) {
+                        newNameString = newNameString.substring(0, indexOf);
+                    } else{
```
timtebeek, line 195: I see you've adapted the Sam-style of formatting PRs ;)
```suggestion
                while (newNameString.length() > maxVariableLength){
                    int indexOf = newNameString.lastIndexOf("_");
                    if (indexOf > -1) {
                        newNameString = newNameString.substring(0, indexOf);
                    } else {
```

timtebeek (approved):

PR #375, opened 2024-10-22
Setting the maximum generated variable length to 40

lkerford:
Removing the configuration option to modify the maximum variable length and changing it from 100 to 40.

40 is already a very long variable name, and we don't think that people will be changing this max length


timtebeek (approved):

timtebeek:
Feel free to merge whenever btw; We'll do a new release tomorrow.

PR #376, opened 2024-10-22
Handle name collision between enums and new string literal in ReplaceDuplicateStringLiterals

lkerford:
Previously if a enum name would collid with newly generated string litterals names, we wouldn't catch the edge case. With this change we add a "_1" to the newly generated string literal

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java
@@ -97,7 +97,7 @@ public class ReplaceDuplicateStringLiterals extends Recipe {
                 if (duplicateLiteralsMap.isEmpty()) {
                     return classDecl;
                 }
-                Set<String> variableNames = duplicateLiteralInfo.getVariableNames();
+                Set<String> usedNames = duplicateLiteralInfo.getUsedNames();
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java
```diff
@@ -236,6 +236,15 @@ public J.VariableDeclarations.NamedVariable visitVariable(J.VariableDeclarations
                     return v;
                 }
 
+                public J.Identifier visitIdentifier(J.Identifier identifier, Integer i) {
+                    J.Identifier v = super.visitIdentifier(identifier, i);
+                    if (v.getType() instanceof JavaType.Class &&
```
lkerford, line 241: There might be a better way of doing so, if there is I'm open to suggestions

timtebeek:
Hmm; I'd have expected any such conflicts to be picked up by VariableNameUtils; did you explore any potential fix there as opposed to in the recipe?
https://github.com/openrewrite/rewrite-static-analysis/blob/5b19b47bc670dc7862839156bd304f79698f95b6/src/main/java/org/openrewrite/staticanalysis/ReplaceDuplicateStringLiterals.java#L195

lkerford:
Closing this PR because we should be making this change in the the `VariableNameUtils`, associated PR below
- https://github.com/openrewrite/rewrite/pull/4600

Issue #384, opened 2024-10-29
ReplaceDuplicateStringLiterals throws java.lang.NullPointerException on version 1.19.0


adambir:
## What version of OpenRewrite are you using?

I am using 

- Maven/Gradle plugin v5.43.2
- rewrite-static-analysis _ v1.19.0 (no error on version 1.18.0)

## How are you running OpenRewrite?

Using the mvn plugin with a multi project module 
mvn -U org.openrewrite.maven:rewrite-maven-plugin:5.43.2:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-static-analysis:1.19.0 -Drewrite.activeRecipes=org.openrewrite.staticanalysis.ReplaceDuplicateStringLiterals

## What is the smallest, simplest way to reproduce the problem?
This code snippet alongside a basic pom reproduces the issue
```java
package com.nonsense.test;

public class PojoTester {

	public static Object getValue() {
		Object[] args = null;
			args = new Object[] { "1" };
		return null;
	}
}
```

## What is the full stack trace of any errors you encountered?
```
[ERROR] Failed to execute goal org.openrewrite.maven:rewrite-maven-plugin:5.43.2:run (default-cli) on project untitled2: Execution default-cli of goal org.openrewrite.maven:rewrite-maven-plugin:5.43.2:run failed: Error while visiting src/main/java/com/whatever/PojoTester.java: java.lang.NullPointerException: Cannot invoke "Object.toString()" because the return value of "org.openrewrite.java.tree.J$Literal.getValue()" is null
[ERROR]   org.openrewrite.staticanalysis.ReplaceDuplicateStringLiterals$DuplicateLiteralInfo$1.visitVariable(ReplaceDuplicateStringLiterals.java:235)
[ERROR]   org.openrewrite.staticanalysis.ReplaceDuplicateStringLiterals$DuplicateLiteralInfo$1.visitVariable(ReplaceDuplicateStringLiterals.java:218)
[ERROR]   org.openrewrite.java.tree.J$VariableDeclarations$NamedVariable.acceptJava(J.java:5901)
[ERROR]   org.openrewrite.java.tree.J.accept(J.java:59)
[ERROR]   org.openrewrite.TreeVisitor.visit(TreeVisitor.java:250)
[ERROR]   org.openrewrite.TreeVisitor.visitAndCast(TreeVisitor.java:320)
[...]

PR #385, opened 2024-10-30
`ReplaceDuplicateStringLiterals` throws NPE from 1.19.0+ #384

adambir:
- Fix https://github.com/openrewrite/rewrite-static-analysis/issues/384

## What's changed?
Added a test for [issue #384](https://github.com/openrewrite/rewrite-static-analysis/issues/384) on the ReplaceDuplicateStringLiterals recipe.

## What's your motivation?
Allow reproduction of the bug

### Checklist


timtebeek (approved):
Thanks for the clear report here @adambir ! Turns out my IDE warned about the same potential NPE that you were seeing. Fixed now by adjusting the condition just above.

Bananeweizen:
@timtebeek Would you mind checking this fix again? The previous logic was effectively "! instanceof String", and the negation is lost now. While I can't really judge the overall condition, the code after the if-condition supports my claim: In line 236 is an explicit getValue().toString() (indicating this is _not_ instanceof String yet). Compare that to line 242, where getValue() is taken directly after checking instanceof String.

timtebeek:
Keen eye there! I've restored the original conditiona and added another to indeed handle these cases correctly, thanks!

PR #388, opened 2024-11-17
Unnecessary `return` as last statement in `void` method

### LINKED ISSUE #75: Recipe: `return` is unnecessary as the last statement in a `void` method
Before : 

```java 

private void login(String name){

        if ("peter".equals(name)){
            System.out.println("Hi admin");
            return;
        }else {
            System.out.println("Access denied");
            return;
        }
    return;
    }

```

After : 

```java 

private void login(String name){

        if ("peter".equals(name)){
            System.out.println("Hi admin");
        }else {
            System.out.println("Access denied");
        }
    }

```

mccartney:
## What's changed?
Adding a new recipe to remove `return` if it's the last statement of a `void` method.

## What's your motivation?
- Closes #75 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
@@ -44,12 +44,12 @@ public class UnnecessaryReturnAsLastStatement extends Recipe {
             public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
                 J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
                 if (TypeUtils.asPrimitive(m.getType()) == JavaType.Primitive.Void) {
-                    return m.withBody(maybeRemoveReturnAsLastStatement(m.getBody()));
+                    return m.withBody(maybeRemoveReturnFromBlock(m.getBody()));
                 }
                 return m;
             }
 
-            private J.@Nullable Block maybeRemoveReturnAsLastStatement(J.@Nullable Block b) {
+            private J.@Nullable Block maybeRemoveReturnFromBlock(J.@Nullable Block b) {
                 if (b == null) {
                     return null;
                 }
@@ -58,27 +58,27 @@ public class UnnecessaryReturnAsLastStatement extends Recipe {
                     if (lastStatement instanceof J.Return && ((J.Return) lastStatement).getExpression() == null) {
                         return null;
                     } else if (lastStatement instanceof J.If) {
-                        return maybeRemoveReturnAsLastStatement((J.If) lastStatement);
+                        return maybeRemoveReturnFromIf((J.If) lastStatement);
                     } else {
                         return lastStatement;
                     }
                 }));
             }
 
-            private J.If maybeRemoveReturnAsLastStatement(J.If ifStatement) {
-                ifStatement = ifStatement.withThenPart(maybeRemoveReturnAsLastStatement(ifStatement.getThenPart()));
+            private J.If maybeRemoveReturnFromIf(J.If ifStatement) {
+                ifStatement = ifStatement.withThenPart(maybeRemoveReturnFromStatement(ifStatement.getThenPart()));
                 J.If.Else elze = ifStatement.getElsePart();
                 if (elze != null) {
-                    return ifStatement.withElsePart(elze.withBody(maybeRemoveReturnAsLastStatement(elze.getBody())));
+                    return ifStatement.withElsePart(elze.withBody(maybeRemoveReturnFromStatement(elze.getBody())));
                 }
                 return ifStatement;
             }
 
-            private Statement maybeRemoveReturnAsLastStatement(Statement s) {
+            private Statement maybeRemoveReturnFromStatement(Statement s) {
                 if (s instanceof J.Block) {
-                    return maybeRemoveReturnAsLastStatement((J.Block) s);
+                    return maybeRemoveReturnFromBlock((J.Block) s);
                 } else if (s instanceof J.If) {
-                    return maybeRemoveReturnAsLastStatement((J.If) s);
+                    return maybeRemoveReturnFromIf((J.If) s);
                 } else {
                     return s;
                 }
--- a/src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.staticanalysis;
 
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
```diff
@@ -0,0 +1,87 @@
...
+                }
+
+                List<Statement> statements = b.getStatements();
+                if (statements.isEmpty()) {
+                    return b;
+                }
+                Statement lastStatement = statements.get(statements.size() - 1);
+                List<Statement> allButLast = statements.subList(0, statements.size() - 1);
+                if (lastStatement instanceof J.Return) {
+                    return b.withStatements(allButLast);
```
knutwannheden, line 62: One thing I am missing here is the check if the return has an expression or not.

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
```diff
@@ -0,0 +1,87 @@
...
+                    return b.withStatements(newStatements);
+                } else {
+                    return b;
+                }
+            }
+
+            @Override
+            public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+                J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
+                if (m == null) {
```
knutwannheden, line 80: Since this visitor never removes methods, this check should not be necessary.

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
mccartney: Right. There's no check for `void` return type.

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
mccartney: 7112d6e

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
mccartney: In bcc951a9e47eaaa9d9e1eb04189d9dc60db8549f adding two lines of defense:

- the whole method being a void method
- the return expression having an empty expression

Not fully sure if you prefer to have both (I've checked both work alone).

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
```diff
@@ -0,0 +1,90 @@
...
+                } else {
+                    return b;
+                }
+            }
+
+            @Override
+            public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+                J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
+                if (TypeUtils.asPrimitive(m.getType()) == JavaType.Primitive.Void) {
+                    return m.withBody(maybeRemoveReturnAsLastStatement(m.getBody()));
```
timtebeek, line 83: If we're only potentially removing the very last statement, then I think this would be a [good use case for `ListUtils.mapLast`](https://github.com/openrewrite/rewrite/blob/ae9083973ca333639ffc4b2231822bbdf7da4e6e/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java#L112-L137), as in: `return m.withBody(m.getBody().withStatements(ListUtils.mapLast(m.getBody().getStatements(), s -> { ... if mere return then return null.... })))`

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
```diff
@@ -0,0 +1,90 @@
...
+                } else {
+                    return m;
+                }
```
timtebeek, line 86: No need for an `else` if the previous block returns.
```suggestion
                }
               return m;
```

src/test/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatementTest.java
```diff
@@ -0,0 +1,123 @@
...
+          java(
+            """
+              class Hello {
+                  int world(int i) {
+                      return i + 436;
+                  }
+              }
+              """));
+    }
+
```
timtebeek, line 122: It might be good to add a test case similar to what we have in https://github.com/openrewrite/rewrite-static-analysis/commit/f1ce05b697a8787d77bfb001476495450cc30959 for a nested `J.NewClass` or `J.Lambda`.

knutwannheden (commented):
I think some more tests would be good. For example removing the `return` here would be correct:

```java
        Consumer<Integer> c = i -> {
            return;
        };
```

while removing the return here would not be correct:
```java
        while (true) {
            return;
        }
```

src/main/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatement.java
mccartney: 411156e

src/test/java/org/openrewrite/staticanalysis/UnnecessaryReturnAsLastStatementTest.java
mccartney: 3325148629f54c5362a96df3bc59f4418051f123

mccartney:
> I think some more tests would be good.

Added some in 3325148629f54c5362a96df3bc59f4418051f123.

Kindly requesting another review.

mccartney:
BTW, is there some easy way to run the Code Suggester locally to avoid the noise in the pull request itself?

timtebeek:
Had another look; thanks so far @mccartney ; Since I think you'll enjoy the puzzle I've added a new unit test that fails. Let me know if you can't or won't enjoy working it out. :)

mccartney:
>  Since I think you'll enjoy the puzzle I've added a new unit test that fails.

b48dd6a

timtebeek (approved):
Approved from my side already; thanks for the thorough tests!

I've replaced the three methods you had with the same name but different arguments with a single recursive method, as the flow was harder to follow across those individual methods.

timtebeek:
Saving others a click: unrelated test failure now in:
```
RemoveRedundantTypeCastTest > changeTypeCastInReturn() FAILED
    org.opentest4j.AssertionFailedError: [Expected recipe to complete in 0 cycles, but took at least one more cycle. Between the last two executed cycles there were changes to "Test.java"] 
```
- Possibly following https://github.com/openrewrite/rewrite/pull/4688

knutwannheden:
> Saving others a click: unrelated test failure now in:
> 
> ```
> RemoveRedundantTypeCastTest > changeTypeCastInReturn() FAILED
>     org.opentest4j.AssertionFailedError: [Expected recipe to complete in 0 cycles, but took at least one more cycle. Between the last two executed cycles there were changes to "Test.java"] 
> ```
> 
> * Possibly following [Make `FindMissingTypes` stricter for method invocations rewrite#4688](https://github.com/openrewrite/rewrite/pull/4688)

Indeed. I've been struggling for a while with that `TypeUtils#isAssignableTo()`, where I've so far neglected the fact that we also need to consider the "direction" of the assignability. While we can assign a `List<String>` type to `<T extends Collection<String>>` when passing in a parameter, the same doesn't work when assigning a local variable. I will add an overload with an additional `direction` parameter, so that we stay backwards compatible.

knutwannheden:
The type assignability check in Java is tricky. I think we might even have some cases where we can't do it correctly, as we don't have all the information we need in our type attribution. In any case I think this PR should now fix the test failure we have here: https://github.com/openrewrite/rewrite/pull/4696

mccartney:
If I read the GHA output correctly, the failure is with:
```
InstanceOfPatternMatchTest > If > typeParameters_3() FAILED
    org.opentest4j.AssertionFailedError: [Expected recipe to complete in 0 cycles, but took at least one more cycle. Between the last two executed cycles there were changes to "A.java"] 
    expected: 
```
now.

knutwannheden:
That was resolved on main. I just merged in the main branch.

knutwannheden (approved):
LGTM.

timtebeek:
Thanks both! As a quick check I'm running this recipe against all Apache projects: https://app.moderne.io/results/5TXPCrrF0
Welcome to use the platform to explore the results, or even push the recipe suggested improvements to other projects! :)

knutwannheden:
All the changes in that recipe run look good to me.

mccartney:
> As a quick check I'm running this recipe against all Apache projects: https://app.moderne.io/results/5TXPCrrF0
Welcome to use the platform to explore the results, or even push the recipe suggested improvements to other projects!

I am hesitant to do this as the Moderne app requires kind of generous permissions (read/write code to all my private repos) to even see the changes.

timtebeek:
Rest assured we're not after your private projects at all; we merely need access to your email address to identify you, and then access to your projects if you choose to fork any OSS projects. I do see how the dialog could come across differently; we're exploring options on our side for that in parallel, but there should be no reason not to use the platform.

PR #390, opened 2024-11-23
EqualsAvoidsNullTest

punkratz312:


punkratz312:
@knutwannheden, feel free to take over. The order doesn’t matter much, but the constantList test is failing in production. I appreciate it!

PR #395, opened 2024-11-26
Adjust `CatchClauseOnlyRethrows` to work with C#

Laurens-W:
## What's changed?

Take into account a possible `J.Empty` expression for a throws statement

@OlegDokuka @knutwannheden @timtebeek 

## Any additional context

In C# an implicit throw like this
```
catch (Exception ex)
{
    throw;
}
```
is valid code and it's mapped to a `J.Empty` expression

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
+++ b/src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
@@ -17,7 +17,9 @@ package org.openrewrite.staticanalysis;
 
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
+import org.openrewrite.SourceFile;
 import org.openrewrite.TreeVisitor;
+import org.openrewrite.csharp.tree.Cs;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.Expression;
@@ -113,6 +115,13 @@ public class CatchClauseOnlyRethrows extends Recipe {
                 }
 
                 Expression exception = ((J.Throw) aCatch.getBody().getStatements().get(0)).getException();
+
+                // In C# an implicit rethrow is possible
+                if (getCursor().firstEnclosing(SourceFile.class) instanceof Cs &&
+                    exception instanceof J.Empty) {
+                    return true;
+                }
+
                 JavaType catchParameterType = aCatch.getParameter().getType();
                 if (!(catchParameterType instanceof JavaType.MultiCatch)) {
                     JavaType.FullyQualified catchType = TypeUtils.asFullyQualified(catchParameterType);
--- a/src/test/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrowsTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrowsTest.java
@@ -16,19 +16,13 @@
 package org.openrewrite.staticanalysis;
 
 import org.junit.jupiter.api.Test;
-import org.openrewrite.DocumentExample;
-import org.openrewrite.FileAttributes;
-import org.openrewrite.InMemoryExecutionContext;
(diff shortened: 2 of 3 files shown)


src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
```diff
@@ -113,6 +115,13 @@ private boolean onlyRethrows(J.Try.Catch aCatch) {
                 }
 
                 Expression exception = ((J.Throw) aCatch.getBody().getStatements().get(0)).getException();
+
+                // In C# an implicit rethrow is possible
+                if (getCursor().firstEnclosing(SourceFile.class) instanceof Cs &&
+                    exception instanceof J.Empty) {
```
timtebeek, line 121: How certain are we that this is always modeled as a `J.Empty`, and not as let's say `null`?

src/test/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrowsTest.java
```diff
@@ -326,4 +332,55 @@ void foo() throws IOException {
...
+
+                  assertThat(output.getMembers().get(0)).isInstanceOf(J.ClassDeclaration.class);
+                  assertThat(((J.ClassDeclaration)output.getMembers().get(0)).getBody().getStatements().get(0)).isInstanceOf(J.MethodDeclaration.class);
+                   md = (J.MethodDeclaration) ((J.ClassDeclaration)output.getMembers().get(0)).getBody().getStatements().get(0);
+                  assertThat(md.getBody().getStatements().get(0)).isInstanceOf(J.Throw.class);
+            }
+            )
+          )
+        );
+    }
```
timtebeek, line 378: I've revised this a bit for readability; I hope you agree.
There's no need for the in between `instanceof` checks; those will fail anyway if incorrect.
```suggestion
                          throw e; // `e` is removed below
                      }
                  }
              }
              """,
            spec -> spec.beforeRecipe(compUnit -> {
                  // C# can rethrow the caught exception implicitly and so the `e` Identifier is removed by the inline visitor below
                  Cs.CompilationUnit cSharpCompUnit = (Cs.CompilationUnit) new JavaVisitor<ExecutionContext>() {
                      @Override
                      public J visitThrow(J.Throw thrown, ExecutionContext ctx) {
                          if (thrown.getException() instanceof J.Identifier) {
                              return thrown.withException(new J.Empty(Tree.randomId(), Space.EMPTY, Markers.EMPTY));
                          }
                          return thrown;
                      }
                  }.visit(JavaToCsharp.compilationUnit(compUnit), new InMemoryExecutionContext());

                  Cs.CompilationUnit after = (Cs.CompilationUnit) new CatchClauseOnlyRethrows().getVisitor()
                    .visit(cSharpCompUnit, new InMemoryExecutionContext());

                  var classA = (J.ClassDeclaration) after.getMembers().get(0);
                  var methodFoo = (J.MethodDeclaration) classA.getBody().getStatements().get(0);
                  var firstStatement = methodFoo.getBody().getStatements().get(0);
                  assertThat(firstStatement).isInstanceOf(J.Throw.class); // The `try/catch` block removed
              }
            )
          )
        );
    }
```

timtebeek: Instead of `beforeRecipe` I've briefly explore `mapBeforeRecipe`, but requires us to return a Java complication unit still.

timtebeek (approved):
Approved after my changes; but the one question about that J.Empty remains.

src/main/java/org/openrewrite/staticanalysis/CatchClauseOnlyRethrows.java
timtebeek: Mostly asking since elsewhere we had problems when recipes were chained; so by and of themselves they'd be fine, but when chained together a previous recipe might `return null`. Although thinking about it; this would then just not match and make no changes presumably; that's fine.

PR #396, opened 2024-11-26
Improve recipes and excluded them from C# codebase if needed

jevanlingen:
## What's changed?
Excluded CSharp for some recipes + made some recipes a little bit more modern

## What's your motivation?
Some recipes don't work for C#.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/StringLiteralEquality.java
+++ b/src/main/java/org/openrewrite/staticanalysis/StringLiteralEquality.java
@@ -22,7 +22,6 @@ import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markers;
 import org.openrewrite.staticanalysis.java.JavaFileChecker;
 
-import java.time.Duration;
 import java.util.Collections;
 import java.util.Set;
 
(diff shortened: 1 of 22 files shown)


timtebeek (approved):
Looks good; what part of this is still in draft? Or can we already merge this increment?

src/main/java/org/openrewrite/staticanalysis/StringLiteralEquality.java
```diff
@@ -110,23 +116,6 @@ private J.Unary asNegatedUnary(J.MethodInvocation mi) {
...
-                    } else if (binary.getOperator() == J.Binary.Type.NotEqual) {
-                        after = asNegatedUnary(asEqualsMethodInvocation(binary)).withPrefix(binary.getPrefix());
-                    }
-                    if (after != null) {
-                        doAfterVisit(new EqualsAvoidsNull().getVisitor());
-                        return after;
-                    }
-                }
-                return super.visitBinary(binary, ctx);
-            }
```
timtebeek, line 129: Did you check which commit had added this override? That might have missed adding a test. Not sure why it was removed now?

jevanlingen: I did not remove it, I just moved the method upwards (so public method first, then the private helper methods).

timtebeek: I can see it's _old_, but when running with coverage it does appear to be hit.
image

timtebeek: Ah it's merely moved up; that's at times hard to spot in a diff.

PR #397, opened 2024-11-27
Adjust `MethodNameCasing` to work with C# 

Laurens-W:
## What's changed?

Introduce interfacing and factory pattern to dynamically retrieve a `NamingConvention` to be applied for given sourcefile

## What's your motivation?

Introduction of C# as a supported language

@timtebeek @OlegDokuka @knutwannheden 

### Checklist


knutwannheden:
Possibly we should leverage the service mechanism we have in JavaSourceFile. We should pull that up API to SourceFile anyway one of these days. Just a thought.

src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
```diff
@@ -113,37 +104,9 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
                     !method.isConstructor() &&
                     !simpleName.startsWith("_") &&
                     !STANDARD_METHOD_NAME.matcher(simpleName).matches()) {
```
knutwannheden, line 106: Keeping this matching here seems wrong to me. That should somehow also be part of the `NamingService`.

src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
```diff
@@ -113,37 +104,9 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
-                                    }
-                                } else {
-                                    standardized.append(c);
-                                }
-                            }
-                        }
-                    }
-
-                    String toName = standardized.toString();
+                    NamingService service = getCursor().firstEnclosing(SourceFile.class).service(NamingService.class);
```
knutwannheden, line 108: The `JavaVisitor` in fact has a `service()` method which is much easier to use. We might want to pull that up to `TreeVisitor` at some point.

We might also have to cache the services a bit more as "breadcrumbs" along the cursor path, so that the service lookups don't get too expensive for deeply nested elements. But that is also something for another day.

src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
```diff
@@ -113,37 +104,9 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
+                    String toName = service.getMethodName(normalized);
```
knutwannheden, line 109: I would rename this method to something else which includes the adjective or verb "standard(ize)" or "normal(ize)". I will let you inspire yourself here: https://sonarsource.github.io/rspec/#/rspec/S100/java

src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
```diff
@@ -110,40 +100,10 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
-                                }
-                            }
-                        }
-                    }
-
-                    String toName = standardized.toString();
+                    NamingService service = service(NamingService.class);
+                    String toName = service.standardizeMethodName(normalized);
                     if (!StringUtils.isBlank(toName) && !StringUtils.isNumeric(toName) &&
                         !methodExists(method.getMethodType(), toName)) {
```
timtebeek, line 108: `standardizeMethodName` is not guaranteed to give back a different method name; do we already have a test to check expecting no changes? Or should we add a comparison here between the old and new name?

timtebeek (approved):
Approved already, but needs a rerun of the tests after this PR is merged:
- https://github.com/openrewrite/rewrite/pull/4725

src/main/java/org/openrewrite/staticanalysis/MethodNameCasing.java
Laurens-W: Think it's best to introduce a check here, better safe then sorry

timtebeek (approved):

PR #398, opened 2024-11-27
`EqualsAvoidsNull` should flip arguments for constants

punkratz312:
Hi,

I might have found an issue in my local project where around 90 violations flagged by PMD are not being replaced correctly. Some of these violations seem related to a list and method arguments.

Could you please take over and enable the necessary tests to confirm if this is indeed an issue?

Thanks in advance!  

@timtebeek 
@knutwannheden 
@ggerbaud

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
@@ -57,11 +57,8 @@ public class EqualsAvoidsNullVisitor<P> extends JavaVisitor<P> {
     @Override
     public J visitMethodInvocation(J.MethodInvocation method, P p) {
         J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, p);
-        if (m.getSelect() != null &&
-            !(m.getSelect() instanceof J.Literal) &&
-            !m.getArguments().isEmpty() &&
-            (m.getArguments().get(0) instanceof J.Literal || m.getArguments().get(0) instanceof J.FieldAccess) &&
-            isStringComparisonMethod(m)) {
+        if (m.getSelect() != null && !(m.getSelect() instanceof J.Literal) &&
+            isStringComparisonMethod(m) && hasCompatibleArgument(m)) {
 
             maybeHandleParentBinary(m);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -60,9 +60,15 @@ public J visitMethodInvocation(J.MethodInvocation method, P p) {
         if (m.getSelect() != null &&
             !(m.getSelect() instanceof J.Literal) &&
             !m.getArguments().isEmpty() &&
-            m.getArguments().get(0) instanceof J.Literal &&
+            (m.getArguments().get(0) instanceof J.Literal || m.getArguments().get(0) instanceof J.FieldAccess) &&
```
timtebeek, line 63: This is the relevant change here; in theory that FieldAccess might itself be nullable, whereas the old left hand side is not. That makes this a challenging change to make reliably without potentially making things worse as well.

timtebeek: Added a check for the flags of the field type; with that we're good to merge.

timtebeek (approved):
Thanks for the help here @punkratz312 ! Nice to see these cases covered as well.

punkratz312:
wow so quick 🥇  well done 🍀

pankratz76:
this is wrong imho

timtebeek:
Ah sorry to hear; in which way? What kind of changes would you propose we make now?

pankratz76:
Sorry for causing this issue. Using the recipes in the codebase itself should have revealed this. Ideally, there should be no dirty code on `main`, assuming all GitHub Actions are applied beforehand. We could then easily implement a check to ensure that the recipes are idempotent. So, running them after merging into `main` should reveal no changes, as they must be applied beforehand ideally.

I experienced a similar situation where the Maven call required multiple steps to cover everything, or it had already flipped. However, if that were the case, it would have also resulted in an infinite loop locally, which seems unlikely. I should have experienced this right away, as there is no exit from that condition.

pankratz76:
> Addressing references is tricky and not covered by other projects like PMD and Checkstyle. Implementing this breaks the idempotency, causing: #472
> 
> Changing a reference with another reference makes no sense to me, which is this feature kind of about. Its like fixing one leak by taking from another one thats not leaking ATM. We just flip for infinity as we can not truly determine which one is null, only by changing the arguments (a reference with another reference).

no PMD is making the assumption so this feature seems correct.

https://github.com/openrewrite/rewrite-static-analysis/pull/473#issuecomment-2720640459

PR #403, opened 2024-12-05
Add new `UseTargetNewOperator`

jevanlingen:
## What's changed?
Add a `UseTargetNewOperator` recipe as C# variant of Java's UseDiamondOperator recipe.

## What's your motivation?
There was an idea of using the `UseDiamondOperator`, but it turns out it's way easier to just make a new recipe for the target-typed new operator instead..

## What is does:
It covers both variable declaration and assignments:

image

### Checklist


jevanlingen:
Todo: 
- There Is still a case left, the code `IList<int> a = new List<int>();` cannot be simplified
- Unit tests

Issue #410, opened 2024-12-18
Use `compileOnly` configuration for optional language dependencies


knutwannheden:
## What problem are you trying to solve?
Currently, the `rewrite-static-analysis` has runtime dependencies on language bindings like `rewrite-csharp` and `rewrite-kotlin` (more to follow). These can be "heavy" and if the user doesn't care about them, they shouldn't have to bring them in.

## Describe the solution you'd like
These dependencies (except for `rewrite-java` of course) should be optional. Thus, a dependency on `rewrite-static-analysis` should not automatically pull them in. Instead the client application should provide them and `rewrite-static-analysis` should mark them as "provided" (Maven parlance), which I believe would translate to the `compileOnly` Gradle configuration.

Additionally, the recipes need to make sure to not fail with runtime exceptions, if these dependencies are "missing". Thus, utility visitors like `CSharpFileChecker` should be based on reflection. (Possibly it would make sense to have a generic visitor in `rewrite-core`, which can be instantiated with a class name, but this is optional and maybe not even a good idea.)

timtebeek:
Have we given any thought to how this would then work in the platform & CLI? Those would have to explicitly put rewrite-csharp and others on the classpath when running against such projects, as by default they only load a minimal isolated classloader I think.

PR #411, opened 2024-12-23
Use compileOnly configuration for optional language dependencies

jevanlingen:
## What's changed?
Groovy, Kotlin and CSharp dependencies are changed to `compileOnly`. The filechecker-classes have been adjusted with a reflection check 

## What's your motivation?
- https://github.com/openrewrite/rewrite-static-analysis/issues/410

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,9 +14,12 @@ dependencies {
 
     implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
     implementation("org.openrewrite:rewrite-java")
-    implementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-csharp:${rewriteVersion}")
+    compileOnly("org.openrewrite:rewrite-groovy:${rewriteVersion}")
+    testImplementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
+    compileOnly("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
+    testImplementation("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
+    compileOnly("org.openrewrite:rewrite-csharp:${rewriteVersion}")
+    testImplementation("org.openrewrite:rewrite-csharp:${rewriteVersion}")
     implementation("org.openrewrite.meta:rewrite-analysis:${rewriteVersion}")
     implementation("org.apache.commons:commons-text:latest.release")
 
--- a/src/main/java/org/openrewrite/staticanalysis/csharp/CSharpFileChecker.java
+++ b/src/main/java/org/openrewrite/staticanalysis/csharp/CSharpFileChecker.java
@@ -20,14 +20,17 @@ import org.openrewrite.Tree;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.csharp.tree.Cs;
 import org.openrewrite.marker.SearchResult;
+import org.openrewrite.staticanalysis.internal.ClassPathUtils;
 
 /**
  * Add a search marker if vising a CSharp file
  */
 public class CSharpFileChecker<P> extends TreeVisitor<Tree, P> {
+    private static final boolean IS_CSHARP_AVAILABLE = ClassPathUtils.isAvailable("org.openrewrite.csharp.tree.Cs$CompilationUnit");
+
     @Override
     public @Nullable Tree visit(@Nullable Tree tree, P p) {
-        if (tree instanceof Cs.CompilationUnit) {
+        if (IS_CSHARP_AVAILABLE && tree instanceof Cs.CompilationUnit) {
             return SearchResult.found(tree);
         }
         return tree;
--- a/src/main/java/org/openrewrite/staticanalysis/internal/ClassPathUtils.java
+++ b/src/main/java/org/openrewrite/staticanalysis/internal/ClassPathUtils.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.staticanalysis.internal;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class ClassPathUtils {
+    public static boolean isAvailable(String className) {
+        try {
+            Class.forName(className);
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
+    }
+}
--- a/src/main/java/org/openrewrite/staticanalysis/groovy/GroovyFileChecker.java
+++ b/src/main/java/org/openrewrite/staticanalysis/groovy/GroovyFileChecker.java
@@ -20,14 +20,17 @@ import org.openrewrite.Tree;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.marker.SearchResult;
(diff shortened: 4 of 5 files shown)


build.gradle.kts
```diff
@@ -14,9 +14,12 @@ dependencies {
 
     implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
     implementation("org.openrewrite:rewrite-java")
-    implementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-kotlin:${rewriteVersion}")
-    implementation("org.openrewrite:rewrite-csharp:${rewriteVersion}")
+    compileOnly("org.openrewrite:rewrite-groovy:${rewriteVersion}")
+    testImplementation("org.openrewrite:rewrite-groovy:${rewriteVersion}")
```
timtebeek, line 18: This was already in the original, but the version of rewrite-groovy is managed by rewrite-bom.
```suggestion
    compileOnly("org.openrewrite:rewrite-groovy")
    testImplementation("org.openrewrite:rewrite-groovy")
```

jkschneider:
I'd like to use the old nebula provided configuration plugin instead here, such that the dependency winds up with a `provided` scope. Most of the tooling we have that executes recipes have runtime dependencies on language bindings and "provide" them in a parent classloader scheme that is highly similar to the original intent of the `provided` scope in Maven (in the old days when J2EE containers provided servlet types, etc.).

https://github.com/nebula-plugins/gradle-extra-configurations-plugin?tab=readme-ov-file#using-the-provided-plugin

src/main/java/org/openrewrite/staticanalysis/csharp/CSharpFileChecker.java
```diff
@@ -20,14 +20,17 @@
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.csharp.tree.Cs;
 import org.openrewrite.marker.SearchResult;
+import org.openrewrite.staticanalysis.internal.ClassPathUtils;
 
 /**
  * Add a search marker if vising a CSharp file
  */
 public class CSharpFileChecker<P> extends TreeVisitor<Tree, P> {
+    private static final boolean IS_CSHARP_AVAILABLE = ClassPathUtils.isAvailable("org.openrewrite.csharp.tree.Cs$CompilationUnit");
```
timtebeek, line 29: I'd perhaps keep it simple and just look for the outer wrapping class, not the inner class.
```suggestion
    private static final boolean IS_CSHARP_AVAILABLE = ClassPathUtils.isAvailable("org.openrewrite.csharp.tree.Cs");
```

src/main/java/org/openrewrite/staticanalysis/internal/ClassPathUtils.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.openrewrite.staticanalysis.internal;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class ClassPathUtils {
+    public static boolean isAvailable(String className) {
+        try {
+            Class.forName(className);
```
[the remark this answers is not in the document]
timtebeek: Only just now saw [this comment](https://github.com/openrewrite/rewrite-static-analysis/pull/411#issuecomment-2559823283); indeed makes most sense to add those `<provided>` dependencies, which are missing right now after a publish to Maven local.
```xml
  <dependencies>
    <dependency>
      <groupId>org.jetbrains</groupId>
      <artifactId>annotations</artifactId>
      <version>26.0.1</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.openrewrite</groupId>
      <artifactId>rewrite-java</artifactId>
      <version>8.43.0-SNAPSHOT</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.openrewrite.meta</groupId>
      <artifactId>rewrite-analysis</artifactId>
      <version>2.15.0-SNAPSHOT</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.apache.commons</groupId>
      <artifactId>commons-text</artifactId>
      <version>1.13.0</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.openrewrite</groupId>
      <artifactId>rewrite-templating</artifactId>
      <version>1.21.0-SNAPSHOT</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
```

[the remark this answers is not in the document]
jkschneider: You should always use this form and set the `initialize` boolean to `false` so that static initializers etc don't get executed unintentionally.

image

> is if the classloader used here would contain the C# / Groovy / Kotlin classes we are likely to parent class load in the CLI & Platform

@timtebeek yes it would, because by default `forName` uses the classloader of `ClassPathUtils`, which will be in the same classloader as the recipe whose parent classloader contains the language specific classes.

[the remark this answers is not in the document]
jkschneider: I would much prefer this utility goes in `openrewrite/rewrite`'s `ReflectionUtils` as it is likely to be used in other recipe modules.

image

build.gradle.kts
```diff
@@ -1,7 +1,9 @@
 plugins {
     id("org.openrewrite.build.recipe-library") version "latest.release"
     id("org.openrewrite.build.moderne-source-available-license") version "latest.release"
+    id("com.netflix.nebula.provided-base") version "10.0.1"
 }
+apply(plugin = "com.netflix.nebula.provided-base")
```
jkschneider, line 6: You shouldn't have both a plugin `id` block above and _also_ an `apply` call. This `apply` mechanism is an old way of applying plugins.

jevanlingen: Thx for the info! I am still trying to make it work; I don't have much experience with Gradle, so this was my first attempt. At the moment, the `provided` functions do not work at all 😞.

jevanlingen: Ah, now I get it! The plugin does not do anything anymore, it's only an [empty](https://github.com/nebula-plugins/gradle-extra-configurations-plugin/blob/main/src/main/groovy/nebula/plugin/extraconfigurations/ProvidedBasePlugin.groovy) plugin today:
> LOGGER.warn("nebula.provided-base plugin does nothing. Use the 'compileOnly' configuration")

See alse: [history](https://github.com/nebula-plugins/gradle-extra-configurations-plugin/commits/main/src/main/groovy/nebula/plugin/extraconfigurations/ProvidedBasePlugin.groovy).

What do you propose now @jkschneider? Start going for the `compileOnly` scope? What do you think @timtebeek ?

---
FYI: there is another [archived project](https://github.com/spring-attic/propdeps-plugin) with the same intention.

src/main/java/org/openrewrite/staticanalysis/internal/ClassPathUtils.java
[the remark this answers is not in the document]
jevanlingen: Done:
- https://github.com/openrewrite/rewrite/pull/4810

src/main/java/org/openrewrite/staticanalysis/csharp/CSharpFileChecker.java
```diff
@@ -19,15 +19,18 @@
...
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.csharp.tree.Cs;
+import org.openrewrite.internal.ReflectionUtils;
 import org.openrewrite.marker.SearchResult;
 
 /**
  * Add a search marker if vising a CSharp file
  */
 public class CSharpFileChecker<P> extends TreeVisitor<Tree, P> {
+    private static final boolean IS_CSHARP_AVAILABLE = ReflectionUtils.isClassAvailable("org.openrewrite.csharp.tree.Cs");
```
sambsnyd, line 29: Let's make an `isLanguageAvailable` function part of `ReflectionUtils` or otherwise available in the rewrite core modules as this functionality will end up used in many places

sambsnyd (approved):

src/main/java/org/openrewrite/staticanalysis/csharp/CSharpFileChecker.java
jevanlingen: @jkschneider did already partly suggest that, as I moved the `isClassAvailable` to the core:
- https://github.com/openrewrite/rewrite/pull/4810

Are you saying you want to go one step further, by creating a `isLanguageAvailable(Enum language)` kind of function in the core?

sambsnyd: That's good enough for now, we can always change it later 👍

Issue #413, opened 2024-12-24
`UseLambdaForFunctionalInterface` should not convert when code uses a static field from enum constructor 


jevanlingen:
## Problem
When running the `UseLambdaForFunctionalInterface` recipe, the code did no longer compile for a enum class. This enum class was a little special, as in the constructor an anonymous class was used with a reference to a `static` field. After conversion to a lambda, compilation error started to appear.

## What is the smallest, simplest way to reproduce the problem?
```java
@Test
void dontUseLambdaWhenEnumAccessesStaticFieldFromConstructor() {
    rewriteRun(
      //language=java
      java(
        """
          import java.time.LocalDate;
          import java.time.format.DateTimeFormatter;
          enum Test {
              A, B;

              private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");

              Test() {
                  Runnable r = new Runnable() {
                      @Override
                      public void run() {
                          DATE_FORMAT.format(LocalDate.now());
                      }
                  };
              }
          }
          """
      )
    );
}
```

## What did you expect to see?
No code change (so anonymous class should not be converted).

## What did you see instead?
A conversion:

```java
import java.time.LocalDate;
import java.time.format.DateTimeFormatter;
enum Test {
    A, B;

    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");

    Test() {
        Runnable r = () -> DATE_FORMAT.format(LocalDate.now());
    }
}
```

which leads to compile error: `Accessing static field from enum constructor is not allowed`.

## Extra info
The reason this error message appears: `static` fields in anonymous classes in the constructor are valid, but the same is not true for lambdas. This is because the loading order of enums is different from that of classes:

<img width="752" alt="image" src="https://github.com/user-attachments/assets/ba627e03-c1ba-4f6f-a1a4-f38a2e86745a" />

PR #415, opened 2024-12-24
`UseLambdaForFunctionalInterface` recipe should not convert when code uses a static field from enum constructor

jevanlingen:
## What's changed?
No more lambda conversion for `UseLambdaForFunctionalInterface` recipe when the anonymous class is within a constructor of a Enum class and uses a static field of this very class.

## What's your motivation?
- https://github.com/openrewrite/rewrite-static-analysis/issues/413

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UseLambdaForFunctionalInterface.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UseLambdaForFunctionalInterface.java
@@ -96,6 +96,10 @@ public class UseLambdaForFunctionalInterface extends Recipe {
                         StringBuilder templateBuilder = new StringBuilder();
                         J.MethodDeclaration methodDeclaration = (J.MethodDeclaration) n.getBody().getStatements().get(0);
 
+                        if (isLambdaInEnumConstructorWithStaticField(getCursor(), methodDeclaration)) {
+                            return n;
+                        }
+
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for taking this on! From a reasoning and maintenance perspective I want to propose an alternative: not make any change within Enums.
```java
            @Override
            public J visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
                if (classDecl.getKind() == J.ClassDeclaration.Kind.Type.Enum) {
                    return classDecl;
                }
                return super.visitClassDeclaration(classDecl, ctx);
            }
```
Given that we rarely expect any lambdas within enum classes, I think it's fair to then just ignore any such cases if they occur. It would greatly simplify the logic here, and cover 3 out of the 5 tests you've written here, with the other 2 ignored to likely little effect. Saves us from then having to reason about and maintain these conditionals for any such edge cases. Any thoughts on that?

jevanlingen:
> Given that we rarely expect any lambdas within enum classes

Well actually: "Given that we rarely expect any ~lambdas~ anonymous classes convertible to lambda's within enum classes", which I think will be even less frequently used than lambda within enum classes. So I agree. As I was writing the logic, more and more edge cases started to appear. Eventually the logic became quite specific and indeed difficult to read.

So let's get to that. I will simplify the code and modify the test accordingly 😎.

timtebeek (approved):

PR #419, opened 2024-12-27
Lambda

punkratz312:
enable `lambdaGenerics` and enjoy yourself @timtebeek 

einen guten Rusch ins neuen Jahr 🍀

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
@@ -47,16 +47,11 @@ import static java.util.Objects.requireNonNull;
 public class EqualsAvoidsNullVisitor<P> extends JavaVisitor<P> {
 
     private static final String JAVA_LANG_STRING = "java.lang.String ";
-    private static final MethodMatcher EQUALS = new MethodMatcher(JAVA_LANG_STRING +
-            "equals(java.lang.Object)");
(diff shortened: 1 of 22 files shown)


src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
```diff
@@ -90,21 +96,19 @@ private boolean hasCompatibleArgument(J.MethodInvocation m) {
     }
 
     private boolean isStringComparisonMethod(J.MethodInvocation methodInvocation) {
-        return EQUALS.matches(methodInvocation) ||
-               !style.getIgnoreEqualsIgnoreCase() &&
-               EQUALS_IGNORE_CASE.matches(methodInvocation) ||
-               COMPARE_TO.matches(methodInvocation) ||
-               COMPARE_TO_IGNORE_CASE.matches(methodInvocation) ||
-               CONTENT_EQUALS.matches(methodInvocation);
+        return METHOD_MATCHERS.stream().anyMatch(matcher -> matcher.matches(methodInvocation))
```
timtebeek, line 99: While I understand the appeal of `anyMatch`, we try to limit our use of the streams API due to the new objects it creates when executing, which adds up at scale.

punkratz312: resolved, thanks for the hint.

timtebeek (approved):
Thanks for helping showcase what's not yet covered @punkratz312 ! We'll merge this as is, and then any fix can be picked up and reviewed separately. 🙏🏻 All the best towards the end of the year!

PR #439, opened 2025-01-11
Don't remove necessary whitespace in EqualsAvoidNull

### LINKED ISSUE #434: EqualsAvoidsNullVisitor removes necessary whitespace
maven 5.47.3

```java
	boolean isCompact(final Foo foo) {
		String property = someCall();
		return property != null && property.equals(COMPACT);
	}
```

## What did you see instead?

```java
	boolean isCompact(final Foo foo) {
		String property = someCall();
		returnCOMPACT.equals(property);
	}
```

The refactoring of the equals is generally okay. Note the missing space after the `return` keyword however. Probably caused by https://github.com/openrewrite/rewrite-static-analysis/blob/dfdb3eb3035a0d5d4454fa4fba87d7f6d7fa4798/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java#L118-L121. I would expect that a refactoring recipe should not touch the whitespace there. And if necessary, there is surely something in the formatting/style area that can be invoked instead?

Given my current knowledge, I could only try removing the code that trims the whitespace. Not sure if that's a reasonable approach.

Bananeweizen:
- Fixes #434, but makes another test fail.

## What's changed?

Avoid unconditionally removing the whitespace prefix of the optimized expression.

## What's your motivation?

It depends on the parent, whether removing whitespace is syntactially valid.

This is incomplete, as another test fails now. We get `if( "test".equals(s)) {}` (note the unwanted space after the opening brace) in another test case, which comes from optimizing `if(s != null && s.equals("test"))` and still using the space character which was after the `&&` operator.

I'm not sure about the correct way to proceed. When I worked with domain specific languages in another project, the formatter of the language (used in the serializer) would enforce additional whitespace in a case like this ("return", ADDITIONAL_SPACE_INSERTED_HERE, returnValue) when processing the return statement. But I've not found any concept like that in the source. It looks like the expressions really only hold their own parsed space, which makes conditionally removing unwanted space like in this visitor much harder.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
@@ -157,7 +157,7 @@ public class EqualsAvoidsNullVisitor<P> extends JavaVisitor<P> {
         public J visitBinary(J.Binary binary, P p) {
             if (scope.isScope(binary)) {
                 done = true;
-                return binary.getRight().withPrefix(Space.EMPTY);
+                return binary.getRight();
             }
             return super.visitBinary(binary, p);
         }
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNullVisitor.java
@@ -157,7 +157,7 @@ public class EqualsAvoidsNullVisitor<P> extends JavaVisitor<P> {
         public J visitBinary(J.Binary binary, P p) {
             if (scope.isScope(binary)) {
                 done = true;
-                return binary.getRight();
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the help here @Bananeweizen ! In high over we see two patterns to fix such issues:
1. assign the prefix of the original expression to the replacement
2. call `maybeAutoFormat` to ensure a minimum between keywords and expressions

Did you try either of those already? And would you want to keep at it or would you like for one of us to take over?

Bananeweizen:
@timtebeek I'd be happy with you taking over. I understand the idea of maybeAutoformat, but wouldn't know _where_ to apply it, since I assume it's needed on the parent of the optimized expression, and the visitor only deals with the expression itself, not with its parent.
If I get it right, then you want to revert the changed line in EqualsAvoidsNullVisitor before applying maybeAutoFormat.

PR #440, opened 2025-01-12
Restore double negation test in `SimplifyBooleanExpressionTest`

Bananeweizen:
## What's changed?

The test wasn't implemented according to its intention. The intended double negation was completely missing.

Someone might want to check if this was optimized away at some point in the past by applying the rewrite recipes to its own code. If so, other places (in test code) might be affected by wrongly having been refactored.

### Checklist


timtebeek (approved):
Thanks for noticing! Indeed a curious case. I can't think of any recipe that would have removed the original; perhaps a botched merge was to blame.

Issue #443, opened 2025-01-13
Invalid output from recipe UnnecessaryThrows with constructor checked exception applied to field initialization


sebastianhaeni:
## What version of OpenRewrite are you using?
I am using

- Maven plugin v6.0.0
- rewrite-static-analysis v2.0.1

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.0.0</version>
    <configuration>
        <exportDatatables>true</exportDatatables>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.UnnecessaryThrows</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-static-analysis</artifactId>
            <version>2.0.1</version>
        </dependency>
    </dependencies>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?
```java
public class Foo {
    String str = bar();
    Foo() throws Exception {} // `throws` is required for initialization of `str`
    String bar() throws Exception { throw new Exception("bar"); }
}
```

## What did you expect to see?

No change because of the checked exception.

## What did you see instead?

```java
public class Foo {
    String str = bar();
    Foo() {} // `throws` was removed, but now this no longer compiles
    String bar() throws Exception { throw new Exception("bar"); }
}

```
This does not compile anymore as the constructor does not declare the checked exception anymore.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
In general yes, but I think I don't have time.

timtebeek:
Oh wow, interesting edge case! Didn't know exceptions could propagate from field initializers to constructors, but it makes sense. As a quick safe guard we can choose not to remove any exceptions from constructors. A more involved solution would look for field initializers that throw exceptions, and retain those on the constructors. A draft PR with just the above test would already help!

sebastianhaeni:
Added a failing test in #444

PR #444, opened 2025-01-13
Skip constructors in `UnnecessaryThrows` for now

sebastianhaeni:
## What's changed?
This adds a test for issue #443


timtebeek:
Thanks @sebastianhaeni ! Figured go for the easier correctness fix right now to not change constructors at all. We can separately explore a proper fix there as well, but this way we at least do no harm until then.

Issue #442, opened 2025-01-13
EqualsAvoidsNull wrongly inverts a.compareTo(b)


Bananeweizen:
## What is the smallest, simplest way to reproduce the problem?

EqualsAvoidsNull is intended to only improve comparisons with null. However, it unconditionally inverts a.compareTo(b) even for non-null arguments. That's completely wrong (because compareTo() implements an _ordered_ comparison), but unfortunately the refactored code still compiles.

## What did you expect to see?

Existing unit tests are wrong and would uncover this issue, if the expectation is fixed. See for example https://github.com/openrewrite/rewrite-static-analysis/blob/3609cd78468ec299ef32f7934ce0d710d942982e/src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java#L49-L62.
`s.compareTo("test")` is not the same as `"test".compareTo(s)`.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

I would just remove the 2 compareTo* matchers, not sure if that is the best fix.

timtebeek:
Hmm interesting; this was first introduced by @punkratz312 I think in 
- https://github.com/openrewrite/rewrite-static-analysis/pull/368

Perhaps these cases should have been split off into a separate issue so as to not confuse the purpose of `EqualsAvoidsNull`, nor invert the comparison accidentally.

timtebeek:
I've pulled those two out; we could reintroduce those in a separate recipe which folks can then choose to apply or exclude as needed.

PR #448, opened 2025-01-23
Fix PMD rule `LiteralsFirstInComparisons` for `compareTo*` and `contentEquals`

punkratz312:
- https://github.com/openrewrite/rewrite-static-analysis/issues/362


src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -54,24 +55,23 @@ public Duration getEstimatedEffortPerOccurrence() {
...
-            @Override
-            public J visit(@Nullable Tree tree, ExecutionContext ctx) {
-                if (tree instanceof JavaSourceFile) {
-                    JavaSourceFile cu = (JavaSourceFile) requireNonNull(tree);
-                    EqualsAvoidsNullStyle style = cu.getStyle(EqualsAvoidsNullStyle.class);
-                    if (style == null) {
-                        style = Checkstyle.equalsAvoidsNull();
+        return Preconditions.check(
+//                new UsesMethod<>("java.lang.String *quals*(..)"),
+                new UsesMethod<>("java.lang.String *(..)"),
```
punkratz312, line 60: Hi @timtebeek,
I need help finding a better regex for equals and compareTo. The current one is just a temporary fix—it might work, but it's inefficient.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -54,24 +55,23 @@ public Duration getEstimatedEffortPerOccurrence() {
...
+                        return (J) tree;
                     }
-                    return new EqualsAvoidsNullVisitor<>(style).visitNonNull(cu, ctx);
                 }
-                //noinspection DataFlowIssue
-                return (J) tree;
-            }
-        };
-        return Preconditions.check(
-                new UsesMethod<>("java.lang.String *quals*(..)"),
```
punkratz312, line 73: I noticed this by accident while wondering why the recipe wasn’t applied. I think it would be better to prioritize a functional approach—putting the most important things first. In a functional style, it’s easier to see and understand the flow right away.

Thanks for considering this!

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
```diff
@@ -447,6 +499,15 @@ public class A {
                       System.out.println(s.compareToIgnoreCase("test"));
                   }
               }
+              """,
+            """
+              public class A {
```
punkratz312, line 504: I’m not sure why we don’t have a second comparison method here—you guys probably know better. Looking at the code, it seems like the goal is to avoid an NPE.

This test was the only one failing, which caught my attention, so I decided to conclude it and ask the team for their thoughts.

Thanks!

punkratz312 (commented):
feel free to takeover and give feedback, thx.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
MBoegers: You know that preconditions can be conjuctive/disjuctive chaind? 
```java
Preconditions.check(
    Preconditions.or(
                new UsesMethod<>("java.lang.String equals(..)"),
                new UsesMethod<>("java.lang.String equalsIgnoreCase(..)")));
```

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
MBoegers: If the visitor is not reused in other recipes, I tend to move the visitor inside the `getVisitor` Method. Maybe that would make it a little bit easier to reason.

punkratz312: I'm not sure if I got it right. I moved the class from a dedicated class file into the recipes one. It would also be possible to make an anonymous class where we currently initialise the external class, following the separation of concerns principle: `return new EqualsAvoidsNullVisitor`.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
punkratz312: a99ae408d27006508d0778b66bf4b5691dd621f9

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
MBoegers: Yes, the anonymous class is what's used all around the place, but I would see any readability concern ;)
What I use to do is extract methods from the `visit*` methods, if I handle different paths. 
Maybe that can help a little bit, but I had no deeper look yet.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
MBoegers: What _comparison_ method are you missing here?

punkratz312: I don’t understand why there is intentionally only one code block. This then cause the test to fail because the modification made was not expected. However, after concluding the test as expected (by switching the arguments), everything works fine.

So, this might be a leftover issue, or perhaps I’m misunderstanding the indentation.

Thanks!

@timtebeek 

image

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
```diff
@@ -64,6 +64,58 @@ public class A {
...
+    @Test
+    void invertConditional_compareTo() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              public class A {
+                  {
+                      String s = null;
+                      if(s.compareTo("test")) {}
```
timtebeek, line 77: I'm wondering if we should add & then invert any comparison with `< 0` here for instance, as indicated in https://github.com/openrewrite/rewrite-static-analysis/issues/362#issuecomment-2588152102. As it stands I think we're using an `int` in an `if`. 😕

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
MBoegers: Because `compareTo`is a comparison, and therefore the flip from `s.compareTo("test")` to `"test".compareTo(s)` would change the semantics. See https://github.com/openrewrite/rewrite-static-analysis/issues/442 for further infos.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
MBoegers: If we decide to keep the flip for `compareTo`, we have to also migrate the comparison. Otherwise, we would also flip the semantic, see https://github.com/openrewrite/rewrite-static-analysis/issues/442.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
punkratz312: thanks for clarification

punkratz312: Even the PMD documentation highlights this issue, so the rule should be revised to prioritize logic over comparison, accepting the risk of a rare NPE rather than enforcing a change that could introduce logical inconsistencies.

I would definitely prefer the safer approach, considering that an NPE (NullPointerException) is, in my opinion, a real programming issue. It’s better to expose such issues explicitly rather than silently altering the comparison logic, which could introduce subtle bugs rather than just mitigating the rare occurrence of an unhandled NPE.

`Note: Switching literal positions for compareTo and compareToIgnoreCase may change the result. See the PMD documentation for examples:`

https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#literalsfirstincomparisons

image

punkratz312:
deepseek managed to inline while chatGPT failed ^^

timtebeek:
Thanks for the work here @punkratz312 ; The automated suggestions are coming in because there's still differences seen after running the following against your branch: https://docs.openrewrite.org/recipes/recipes/rewrite/openrewriterecipebestpractices
Would you mind applying those changes?

punkratz312:
of course my friend. done and dusty

punkratz312:
@timtebeek the auto correction does not work. i applied it once before and wonder why. its. not working. see the. ci. My commit is green. The github action is failing.

image

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
MBoegers: ATM, the implementation exclud the flip of `<` and `>` needed for `compareTo`, this can be done later.

MBoegers (approved):
I removed some test that were already covered and merged the two visitors.
Great work!

PR #451, opened 2025-01-24
feat: prevent UnnecessaryCatch from removing `catch (Throwable t)`

auke-:
Similar to `catch (Exception e)` it is undesired to remove `catch (Throwable t)` blocks as it changes the runtime behavior.

## What's changed?
The UnnecessaryCatch rule did always remove `catch (Throwable t)`, which changes runtime behavior.

## What's your motivation?
Although the use of Throwable isn't recommended there are use cases. When `catch (Throwable t)` is removed it changes the runtime behavior.

### Checklist


timtebeek:
Hi @auke- ; thanks for the immediate fix here! Makes sense indeed.

timtebeek (approved):

Issue #461, opened 2025-02-11
`java.lang.IndexOutOfBoundsException` Index 0 out of bounds for length 0


punkratz312:
checkout: https://github.com/pmd/pmd

run: `mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-static-analysis:RELEASE -Drewrite.activeRecipes=org.openrewrite.staticanalysis.CodeCleanup -Drewrite.exportDatatables=true`

```
[INFO] PMD Distribution Packages .......................... FAILURE [01:43 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:33 min
[INFO] Finished at: 2025-02-11T09:25:25+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.openrewrite.maven:rewrite-maven-plugin:6.1.1:run (default-cli) on project pmd-dist: Execution default-cli of goal org.openrewrite.maven:rewrite-maven-plugin:6.1.1:run failed: Error while visiting pmd-java/src/test/resources/net/sourceforge/pmd/lang/java/ast/jdkversiontests/java21/ExhaustiveSwitch.java: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
[ERROR]   java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
[ERROR]   java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
[ERROR]   java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
[ERROR]   java.base/java.util.Objects.checkIndex(Objects.java:385)
[ERROR]   java.base/java.util.ArrayList.get(ArrayList.java:427)
[ERROR]   org.openrewrite.staticanalysis.DefaultComesLastVisitor.isDefaultCase(DefaultComesLastVisitor.java:189)
[ERROR]   org.openrewrite.staticanalysis.DefaultComesLastVisitor.isDefaultCaseLastOrNotPresent(DefaultComesLastVisitor.java:173)
[ERROR]   org.openrewrite.staticanalysis.DefaultComesLastVisitor.visitSwitch(DefaultComesLastVisitor.java:44)
[...]

PR #462, opened 2025-02-11
DefaultComesLast should work on Java 21 Object switch

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/461

Fails for now with what likely needs an upstream fix first:
```
Caused by: java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "org.openrewrite.java.tree.J$Identifier.getSimpleName()" is null
	at org.openrewrite.java.JavaPrinter.visitCase(JavaPrinter.java:488)
	at org.openrewrite.java.JavaPrinter.visitCase(JavaPrinter.java:35)
	at org.openrewrite.java.tree.J$Case.acceptJava(J.java:1118)
	at org.openrewrite.java.tree.J.accept(J.java:58)
	at org.openrewrite.TreeVisitor.visit(TreeVisitor.java:250)
	... 57 more
```

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
@@ -15,7 +15,6 @@
  */
 package org.openrewrite.staticanalysis;
 
-import jdk.vm.ci.meta.JavaType;
 import lombok.AllArgsConstructor;
 import lombok.Value;
 import lombok.With;
@@ -28,6 +27,7 @@ import org.openrewrite.internal.RecipeRunException;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.service.ImportService;
+import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Marker;
 import org.openrewrite.marker.Markers;
 import org.openrewrite.staticanalysis.csharp.CSharpFileChecker;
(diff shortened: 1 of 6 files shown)


punkratz312 (approved):
nice one

timtebeek:
There's further similar replacements we have to make across this and a few other projects to avoid issues on Java 21, and phase out the deprecated method:
- https://app.moderne.io/results/6HFfmN8wJ

image

src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
```diff
@@ -27,7 +28,6 @@
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.service.ImportService;
-import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Marker;
```
timtebeek, line 31:
```suggestion
import org.openrewrite.java.tree.*;
import org.openrewrite.marker.Marker;
```

src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.staticanalysis;
 
+import jdk.vm.ci.meta.JavaType;
```
timtebeek, line 18:
```suggestion
```

timtebeek (commented):
Fix what IntelliJ messed up.

PR #464, opened 2025-02-13
fix missing ObjectEquals

punkratz312:
this is last one. The lombok stuff is working now. 👍


timtebeek (approved):
Great to see, thanks for adding this missing case!

PR #466, opened 2025-02-16
recipe: replace Arrays#asList() by List#of()

kthoms:
## What's changed?

Adds a new recipe

## What's your motivation?

We have lots of usages in our code base and they are reported as warnings by default

The second test is failing. I need help here with the implementation to fix it.
```
Failed to run recipe at Cursor{MethodInvocation->JRightPadded(element=Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8, 9, 10), after=Space(comments=<0 comments>, whitespace=<empty>))->Block->MethodDeclaration->JRightPadded(element=MethodDeclaration{A{name=foo,return=void,parameters=[]}}, after=Space(comments=<0 comments>, whitespace=<empty>))->Block->ClassDeclaration->CompilationUnit->root}
```

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ReplaceArraysAsListWithListOf.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ReplaceArraysAsListWithListOf.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for kicking this off @kthoms ; we had so far held off on a recipe like this because `List.of` returns immutable collections, whereas `Arrays.asList` returns a new mutable array list. That means this is not yet a guaranteed safe replacement, and could lead to runtime issues. What are your thoughts on that?

kthoms:
You are right, Tim. Although most of the time an array would be converted to a iist for the sake to use the values for API that requires lists, it is not always true and might even break code.

Maybe we could think about restricting it to test code? There we are very likely safe and it is where most usages of the method will be.

Also, I forgot that a check/test is needed that List.of() should not be used when Arrays.asList() is used with a single argument that is an array.

Do you think I should extend the code in that sense?

PR #468, opened 2025-02-18
Fix RemoveUnusedPrivateMethods `org.opentest4j.AssertionFailedError`

Pankraz76:
WIP: Fix RemoveUnusedPrivateMethods `removeUnusedPrivateMethodsChainedUsage`

```
org.opentest4j.AssertionFailedError: [Expected recipe to complete in 1 cycle, but took at least one more cycle. Between the last two executed cycles there were changes to "Test.java"] 
```

@timtebeek feel free to takeover. Imho this occurred in production as well, so thats why I found it.


timtebeek:
So this is one of those rare cases where multiple cycles could be expected; the trick then is to use `Repeat` as shown in my commit, thanks!

timtebeek (approved):
Thanks for the help here @Pankraz76 !

Pankraz76:
Thanks for resolving

Issue #472, opened 2025-03-07
EqualsAvoidsNull not idempotent


protocol7:
## What version of OpenRewrite are you using?

* org.openrewrite:rewrite-core:8.47.2
* org.openrewrite.recipe:rewrite-static-analysis:2.3.0

## What is the smallest, simplest way to reproduce the problem?

EqualsAvoidsNull will non-idempotently swap places of constants and string literals. We see this in our daily runs of the recipe in that the same code will switch back and forth every other day. The test case below is a minimised version of our real code.

I assume that the preferred order would be to have the string literal first.

```java
  @Test
  void equalsAvoidsNullNonIdempotent() {
    rewriteRun(
        spec -> spec.recipe(new EqualsAvoidsNull()),
        // language=java
        java(
            """
            package com.helloworld;

            public class Foo {
                private static final String FOO = "";

                public void foo() {
                    FOO.equals("");
                    "".equals(FOO);
                }
            }
            """));
  }
```

```
org.opentest4j.AssertionFailedError: [Expected recipe to complete in 0 cycles, but took at least one more cycle. Between the last two executed cycles there were changes to "com/helloworld/Foo.java"]
expected:
  "package com.helloworld;

  public class Foo {
      private static final String FOO = "";

      public void foo() {
          FOO.equals("");
          "".equals(FOO);
      }
  }"
 but was:
  "package com.helloworld;

  public class Foo {
      private static final String FOO = "";

      public void foo() {
          "".equals(FOO);
          FOO.equals("");
      }
  }"

```

pankratz76:
thanks for the report. We will have a look into this.

pankratz76:
imho still open: https://github.com/openrewrite/rewrite-static-analysis/pull/487

pankratz76:
@protocol7 have you retested the issue? I think its still open.

protocol7:
Yes, now works as expected on latest OpenRewrite. This issue is closed, and I'm not sure I have enough context to know if #487 also required.

protocol7:
Currently we only have the raw strings case, but that might change as we're adding more code to be exposed to OpenRewrite.

pankratz76:
All right thanks - then we gonna meet up again, until fixed. 

Cheers.

PR #473, opened 2025-03-10
fix Issue 472 `EqualsAvoidsNull` not idempotent add missing check on `compareTo` and `compareToIgnoreCase`

pankratz76:
fixes: https://github.com/openrewrite/rewrite-static-analysis/issues/472


src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
```diff
@@ -473,4 +473,104 @@ public class A {
...
+                          "FOO".compareToIgnoreCase("BAR");
+                      }
+                  }
+                  """
+              )
+            );
+        }
+
+        /**
+         * ATM we cannot determine this case without breaking {@link ReplaceConstantMethodArg#one}
```
pankratz76, line 501: @timtebeek i was not able to determine the states correctly. Maybe you can fix this case. Its possible to make the test work but not without breaking the pre existing once.

pankratz76:
I think we cannot determine if it's a string literal or a raw string written in quotes.  
This needs to be determined in order to check when to flip.  

Right now, it seems that we get the same information—which might be correct—for each situation, making it impossible for us to detect.  
It's always the `java.lang.String` and not just a simple `String` that comes as a type.  

Issue:
`java.lang.String` and `java.lang.String`
image

`java.lang.String` and `java.lang.String`
image

`java.lang.String` and `String`
image

pankratz76:
@timtebeek @MBoegers would you please give me a hint, how to solve this?

MBoegers:
hi @pankratz76 hav you tried using [`TypeUtils`](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java#L102)? it helped me in most of this cases.

src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
MBoegers: Did you tried to solve it with the `DataFlow` APIs `ConstantFold` [feature](https://github.com/openrewrite/rewrite-analysis/blob/main/src/main/java/org/openrewrite/analysis/constantfold/ConstantFold.java#L41)? It determines the constant value if possible.

pankratz76:
> hi @pankratz76 hav you tried using [`TypeUtils`](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java#L102)? it helped me in most of this cases.

Thanks for the hint.

It’s not able to detect it, but I’m wondering how it should, since the type appears to be correct—`java.lang.String`. Simply checking the type doesn’t seem sufficient. We need to determine the source of the string:

If it references a field, we can modify it.
If it points to a constant or a higher-scoped field/reference, we should prioritize that.
Just comparing two strings alone makes it difficult to determine the correct one without additional context.

mixmatch:

image


string on string not determinable: 

image

pankratz76:
as least not its clear that the assumption made by PMD is then although our way to go imho, just to go with the flow. Still we then face the issue to track back the reference and prefer constants first as requested initially.

`that if you declare a constant string (that is a final static field of type String), it would never be initialized with null `

https://github.com/pmd/pmd/issues/5590#issuecomment-2720399235

pankratz76:
## How to Build the Code Tree as Checkstyle Works  

Can you show me how to build the code tree as Checkstyle works?  
[Checkstyle PR #16435](https://github.com/checkstyle/checkstyle/pull/16435)  

I already tried adding the user type like seen elsewhere, assuming this would then also be applied.  
But I don't see the variable declaration.  

This needs to be built before making the actual flip to determine which string reference should be preferred.  

Image

greg-at-moderne:
FYI, #485 has been merged.

pankratz76:
thanks

pankratz76:
> FYI, #485 has been merged.

We need to implement reference declaration backward parsing and check the location of the references we want to consider for swapping. Only then can we determine if the swap makes sense.

PR #488, opened 2025-03-25
rewise early exit to positive condition in `EqualsAvoidsNull`

pankratz76:
## What's changed?
* rewise early exit to positive condition in `EqualsAvoidsNull`
* separate concerns: `one thing only`

## What's your motivation?
* clean code

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
@@ -76,10 +76,10 @@ public class EqualsAvoidsNull extends Recipe {
                 new JavaVisitor<ExecutionContext>() {
                     @Override
                     public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
-                        return visitMethodInvocation((J.MethodInvocation) super.visitMethodInvocation(method, ctx));
+                        return checkLiteralsFirstInComparisons((J.MethodInvocation) super.visitMethodInvocation(method, ctx));
                     }
 
-                    private J visitMethodInvocation(J.MethodInvocation m) {
+                    private J checkLiteralsFirstInComparisons(J.MethodInvocation m) {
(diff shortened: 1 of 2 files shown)


pankratz76:
@greg-at-moderne kindly request feedback

greg-at-moderne (commented):
TBH, I don't fully find it better with the overloaded standard `visit...` method.

pankratz76:
> TBH, I don't fully find it better with the overloaded standard `visit...` method.

yes thats non a good story telling. Structuring it with SOC is ok, right? Its just about finding the proper size and name. Please see updates version. Thx

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -76,104 +76,110 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                            return applyLiteralsFirstInComparisons(m, m.getArguments().get(0));
                         }
+                        return m;
+                    }
 
-                        maybeHandleParentBinary(m, getCursor().getParentTreeCursor().getValue());
-                        Expression firstArgument = m.getArguments().get(0);
-
+                    private J applyLiteralsFirstInComparisons(J.MethodInvocation m, Expression firstArgument) {
+                        checkPotentialNullCheck(m, getCursor().getParentTreeCursor().getValue());
```
pankratz76, line 92: I don't know how `checkPotentialNullCheck` is working, but from the storytelling here, I would expect this to act as an early exit. A null check does not get flipped, which doesn’t make sense because null is checked again below in `literalsFirstInComparisonsNull`, so it must be treated as well.  

Therefore, it seems more like a treatment or preparation step and should be renamed to `preparePotentialNullCheck` to better reflect its purpose in allowing the process to continue.

pankratz76:
https://github.com/openrewrite/rewrite-static-analysis/pull/497

PR #489, opened 2025-03-26
Fix several issues in InstanceOfPatternMatch

### LINKED ISSUE #308: InstanceOfPatternMatch gives wrong type parameter
I am using

- OpenRewrite v6.16.2
- Maven/Gradle plugin v6.16.2
- org.openrewrite.recipe:rewrite-migrate-java:2.18.1

I am using the Gradle plugin, and my project is a multi module project.

```
initscript {
    repositories {
        maven { url "https://plugins.gradle.org/m2" }
    }
    dependencies {
        classpath("org.openrewrite:plugin:6.16.2")
    }
}

rootProject {
    plugins.apply(org.openrewrite.gradle.RewritePlugin)
    dependencies {
        rewrite("org.openrewrite.recipe:rewrite-spring:latest.release")
        rewrite("org.openrewrite.recipe:rewrite-migrate-java:2.18.1")
    }

    afterEvaluate {
        if (repositories.isEmpty()) {
            repositories {
                mavenCentral()
            }
        }
    }
}
```

and run

```
./gradlew --init-script rewrite.gradle rewriteRun -Drewrite.activeRecipe=org.openrewrite.staticanalysis.InstanceOfPatternMatch
```

A code snippet can be something simple like this.

```java
        if(obj instanceof List) {
            test((List)list);
        }

     static public String test(List<String> list){
```

```java
        if(obj instanceof List list) {
            test(list);
        }

     static public String test(List<String> list){
```

## What did you see instead?

```java
        if(obj instanceof List<?> list) {
            test(list);
        }

     static public String test(List<String> list){
```

`List<?>` is not the same as `List` , this will result in a compilation error like this:

```
 error: incompatible types: List<CAP#1> cannot be converted to List<String>
            test(list));
            ^
  where CAP#1 is a fresh type-variable:
    CAP#1 extends Object from capture of ?
```

Indicate if this is something you would like to work on, and how we can best support you in doing so.

### LINKED ISSUE #482: InstanceOfPatternMatch does not always handle generics correctly

I am using

- Maven plugin v6.3.2
- rewrite-migrate-java v3.4.0

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.3.2</version>
    <configuration>
        <exportDatatables>true</exportDatatables>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.InstanceOfPatternMatch</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>3.4.0</version>
        </dependency>
    </dependencies>
</plugin>
```

```java
package org.example;

public class Main {
    public class Bar<T> {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar) {
            @SuppressWarnings("unsafe")
            Bar<String> barString = (Bar) o;
            System.out.println(barString);
        }
    }
}
```

```java
package org.example;

public class Main {
    public class Bar<T> {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar barString) {
            System.out.println(barString);
        }
    }
}
```
Note: casting to `Bar<?>` would not be correct in all cases, even if it would work in this one.

## What did you see instead?

```java
package org.example;

public class Main {
    public class Bar<T> {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar<String> barString) {
            System.out.println(barString);
        }
    }
}
```

The migration succeeds, but if you try to compile the changes afterwards javac fails with
```
java.lang.Object cannot be safely cast to org.example.Main.Bar<java.lang.String>
```

If I manage to do so with reasonable effort, yes.

### LINKED ISSUE #483: InstanceOfPatternMatch does not shadow/rename properly

I am using

- Maven plugin v6.3.2
- rewrite-migrate-java v3.4.0

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.3.2</version>
    <configuration>
        <exportDatatables>true</exportDatatables>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.InstanceOfPatternMatch</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>3.4.0</version>
        </dependency>
    </dependencies>
</plugin>
```

```java
package org.example;

public class Main {
    public class Bar {}
    Object bar;
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar) {
            Bar bar = (Bar)o; 
            System.out.println(bar);
        }
    }
}
```

```java
package org.example;

public class Main {
    public class Bar {}
    Object bar;
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar bar) { 
            System.out.println(bar);
        }
    }
}
```

## What did you see instead?

```java
package org.example;

public class Main {
    public class Bar {}
    Object bar;
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar bar1) { 
            System.out.println(bar);
        }
    }
}
```

The migration succeeds, but if you try to compile the changes afterwards javac fails with
```
non-static variable bar cannot be referenced from a static context
```
Note: depending on the setup this could also very much fail silently and thereby introduce bugs.

If I manage to do so with reasonable effort, yes.

### LINKED ISSUE #484: InstanceOfPatternMatch does not work correctly with enums

I am using

- Maven plugin v6.3.2
- rewrite-migrate-java v3.4.0

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.3.2</version>
    <configuration>
        <exportDatatables>true</exportDatatables>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.InstanceOfPatternMatch</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>3.4.0</version>
        </dependency>
    </dependencies>
</plugin>
```

```java
package org.example;

public class Main {
    public enum Foo {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Foo) {
            Foo e = (Foo) o;
            System.out.println(e);
        } 
    }
}
```

```java
package org.example;

public class Main {
    public enum Foo {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Foo e) {
            System.out.println(e);
        } 
    }
}
```

## What did you see instead?

```java
package org.example;

public class Main {
    public enum Foo {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Foo e1) {
            System.out.println(e);
        } 
    }
}
```
Note: this behavior seems to be specific to enums, the pattern works fine with regular classes.

The migration succeeds, but if you try to compile the changes afterwards javac fails with
```
cannot find symbol
[ERROR]   symbol:   variable e
[ERROR]   location: class org.example.Main
```

If I manage to do so with reasonable effort, yes.

cortlepp:
## What's changed?

- always use the type in the cast for the new pattern and not the previously checked `instanceof` type.
- when a variable name is already present in the source, don't search for duplicates and just take it as it is. Otherwise we would have to make unnecessary changes when the variable happens to be shadowing something else.
- refuse to pattern-match non-wildcard generic types, as this cannot be checked due to type erasure.

## What's your motivation?

- Fix #308
- Fix #482 
- Fix #483 
- Fix #484

I'll annotate some areas in the files themselves.

@greg-at-moderne 
### Checklist


src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
```diff
@@ -251,23 +251,22 @@ public J.InstanceOf processInstanceOf(J.InstanceOf instanceOf, Cursor cursor) {
...
                     randomId(),
                     Space.build(" ", emptyList()),
                     Markers.EMPTY,
                     emptyList(),
                     name,
-                    type,
-                    null));
+                    typeCastTypeTree.getType(),
+                    null))
+                    .withClazz(typeCastTypeTree.withPrefix(currentTypeTree.getPrefix()).withId(Tree.randomId()));
```
cortlepp, line 265: It seems superfluous or possibly even wrong that we have to pass the type to the Identifier and also set it as `clazz` on the `J.InstanceOf`. Maybe this could be avoided with Java template, but I was unsure how one would look for this case.

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
```diff
@@ -251,23 +251,22 @@ public J.InstanceOf processInstanceOf(J.InstanceOf instanceOf, Cursor cursor) {
...
 
-            J currentTypeTree = instanceOf.getClazz();
-            TypeTree typeCastTypeTree = computeTypeTreeFromTypeCasts(instanceOf);
-            // If type tree from type cast is not parameterized then NVM. Instance of should already have proper type
-            if (typeCastTypeTree instanceof J.ParameterizedType) {
-                J.ParameterizedType parameterizedType = (J.ParameterizedType) typeCastTypeTree;
-                result = result.withClazz(parameterizedType.withId(Tree.randomId()).withPrefix(currentTypeTree.getPrefix()));
+            if (typeCastTypeTree instanceof J.ParameterizedType && ((J.ParameterizedType) typeCastTypeTree).getTypeParameters().stream().anyMatch(e -> !(e instanceof J.Wildcard))) {
```
cortlepp, line 267: It seems to me that this should already be covered by `isAcceptableTypeCast()`, but apparently it isn't. Why?

timtebeek (commented):
Thanks a lot for the multiple fixes here @cortlepp ! @greg-at-moderne will have a look shortly and see if we can get this in before the release

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
greg-at-moderne: Also - since you mentioned - would it make sense to replace the `if` condition with a call to `isAcceptableTypeCast`?

greg-at-moderne (approved):
I appreciate the great contribution. I consider it a piece of high quality work. Well tested, thoughtful and solving a few overlapping cases in one go!

There's one minor outstanding issue and I think we are good to merge.

(Also, I acknowledge your questions on some of the linked issues, I will get to them).

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
cortlepp: I guess I thought that by giving the identifier a type in line 263, this should determine the type for the entire expression and I should not have to set it again as the `clazz`. To me it seems like `clazz` should always be the same as the type passed in line 263, therefore I thought maybe I was using the class wrong or I misunderstood something generally.

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
cortlepp: I finally found out why `isAcceptableTypeCast()` did not work: we only applied it to casts, but not variable assignments. It turns out that instead of really fixing the Issue I had messed up my initially reported test case so that it was rejected by another check that had already been in place. The latest commit now removes my previous "fix" and applies the `isAcceptableTypeCast()` check to both casts and assignments. For this I had to change it to now work with `JavaType` and not `J` (what is the difference???).

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
```diff
@@ -201,11 +201,13 @@ public void registerTypeCast(J.TypeCast typeCast, Cursor cursor) {
...
-                                variablesToDelete.put(instanceOf, new VariableAndTypeTree(parent.getValue(),
-                                        requireNonNull(parent.firstEnclosing(J.VariableDeclarations.class).getTypeExpression())));
+                                    !variablesToDelete.containsKey(instanceOf)) {
+                                if (isAcceptableTypeCast(((J.VariableDeclarations.NamedVariable) parent.getValue()).getType())) {
+                                    variablesToDelete.put(instanceOf, new VariableAndTypeTree(parent.getValue(),
+                                            requireNonNull(parent.firstEnclosing(J.VariableDeclarations.class).getTypeExpression())));
+                                }
                             } else {
                                 replacements.put(typeCast, instanceOf);
                             }
```
cortlepp, line 213: In the following line (214) the current cursor is now still placed into `contextScopes` even if the `if` in line 207 fails. I don't really understand what it being in the `contextScopes` means, so I'm not sure whether we should do nothing in the else branch (as it is now) or if we should `continue` (The test suite passes either way).

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
greg-at-moderne: `JavaType` represents information on a type.

`J` is a superclass (well, interface) of all nodes in a Java LST.

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
greg-at-moderne: Sorry for a slow response. I needed to find time to check out your code locally in order to share an informed comment.

So the `contextScopes` variable purpose seems to be for making sure we only process these `instanceof` statements whose variables are later used. This is how I read the [line 252](https://github.com/openrewrite/rewrite-static-analysis/blob/2183b3872609b04c471b63a7e95c1b32dc9aaff6/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java#L252) as confirmed with the `multipleVariablesOnlyOneUsed` test case.

So having said that I think it's clearer if we don't add to the `contextScopes` map if this particular instanceof is not to be processed. I.e. when the `if isAcceptableTypeCast` evaluates to false.

I didn't try that, but probably one could add a test case similar to `multipleVariablesOnlyOneUsed`, but the one not used would be used, but not acceptable to cast.

I hope it helps.

greg-at-moderne:
Does it make sense to you?
Not trying to put any pressure or anything like this. Just wondering if you are blocked by anything here?

cortlepp:
Hey, I'm currently on vacation until next week, I'll take a look at it as soon as I return. (Sorry I should've said something, I just forgot)

src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
cortlepp: See my latest changes, I hope I understood you correctly :). The testcase I added fails with an NPE on the previous commit, so your theory seems to be correct. Thanks for looking into it!

greg-at-moderne:
Not sure what's happening with Github, might be some glitch. I have clicked the button to approve CI run a few times already.

cortlepp:
I think the run went through now (on your merge commit).

greg-at-moderne (approved):
I am satisfied with this change. I appreciate the contribution.

arodionov:
It looks like the current PR create an issue https://github.com/openrewrite/rewrite-static-analysis/issues/528

Issue #428, opened 2025-01-07
AnnotateNullableMethods should have a parameter for the FQN of the nullable type


Bananeweizen:
## What problem are you trying to solve?
AnnotateNullableMethods currently adds a _fixed_ Nullable annotation, the one that OpenRewrite uses itself. The fully qualified name for that nullable type should be parameterizable, as many projects use other nullable types. Even OpenRewrite changed its code base just recently to switch to JSpecify.

## Describe the solution you'd like
one new argument for the FQN to use, with the default of the current value to not cause migration effort

## Have you considered any alternatives or workarounds?
Right now people with other annotations also can just run this recipe and then do a search-replace of the import.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes. Is there a convention for naming parameters or would you suggest a name? "nullableType"?

timtebeek:
hi @Bananeweizen ; thanks for the suggestion! One thing to check perhaps before we make the nullable annotation configurable, is if your desired argument supports this style of use:
https://github.com/openrewrite/rewrite-static-analysis/blob/37c8568b65a2232da30bac01e32eb777bc29576f/src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java#L247

If yes, then fine to make it optionally configurable with the existing value the default; Otherwise it might complicate matters a bit.

Bananeweizen:
@timtebeek Your question is basically: Is my null annotation a declaration annotation (Java 5) or a type annotation (Java 8) using `@Target({...})`. AFAIK all type annotations fit your needs (or the FQN case like `java.io.@Nullable File`) and similar. My null annotation (Eclipse JDT) definitely fits, and I would also expect other major null annotations to be type annotations nowadays.

timtebeek:
Perfect, then indeed ok to add a `nullableType` option to `AnnotateNullableMethods` which takes a fully qualified class name. 🙏🏻

jevanlingen:
Fixed with :arrow_up:.

PR #491, opened 2025-04-01
Added option to provide a custom FQN to AnnotateNullableMethods recipe

JohannisK:
## What's changed?
Added option to provide a custom FQN to AnnotateNullableMethods recipe

## What's your motivation?
As requested in https://github.com/openrewrite/rewrite-static-analysis/issues/428

@jevanlingen 
@Laurens-W 
@greg-at-moderne

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
+++ b/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
@@ -35,7 +35,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public class AnnotateNullableMethods extends Recipe {
 
     @Option(displayName = "`@Nullable` annotation class",
-            description = "The fully qualified name of the @Nullable annotation. The annotation should have `@Target(TYPE_USE)`. Defaults to `org.jspecify.annotations.Nullable`",
+            description = "The fully qualified name of the @Nullable annotation. The annotation should be meta annotated with `@Target(TYPE_USE)`. Defaults to `org.jspecify.annotations.Nullable`",
             example = "org.jspecify.annotations.Nullable",
             required = false)
     @Nullable
@@ -52,7 +52,7 @@ public class AnnotateNullableMethods extends Recipe {
     public String getDescription() {
         return "Add `@Nullable` to non-private methods that may return `null`. " +
                 "By default `org.jspecify.annotations.Nullable` is used, but through the `nullableAnnotationClass` option a custom annotation can be provided." +
-                "When providing a custom `nullableAnnotationClass` that annotation should have `@Target(TYPE_USE)`." +
+                "When providing a custom `nullableAnnotationClass` that annotation should be meta annotated with `@Target(TYPE_USE)`." +
                 "This recipe scans for methods that do not already have a `@Nullable` annotation and checks their return " +
                 "statements for potential null values. It also identifies known methods from standard libraries that may " +
                 "return null, such as methods from `Map`, `Queue`, `Deque`, `NavigableSet`, and `Spliterator`. " +
--- a/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
+++ b/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
@@ -60,20 +60,26 @@ public class AnnotateNullableMethods extends Recipe {
     }
 
     @Override
+    public Validated<Object> validate() {
+        return super.validate()
+                .and(Validated.test("nullableAnnotationClass", "<message>", nullableAnnotationClass,
+                        it -> it == null || it.contains(".")));
+    }
+
+    @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         String fullyQualifiedName = nullableAnnotationClass != null ? nullableAnnotationClass : DEFAULT_NULLABLE_ANN_CLASS;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
```diff
@@ -31,10 +30,18 @@
...
 import java.util.concurrent.atomic.AtomicBoolean;
 
+@Value
+@EqualsAndHashCode(callSuper = false)
 public class AnnotateNullableMethods extends Recipe {
 
-    private static final String NULLABLE_ANN_CLASS = "org.jspecify.annotations.Nullable";
-    private static final AnnotationMatcher NULLABLE_ANNOTATION_MATCHER = new AnnotationMatcher("@" + NULLABLE_ANN_CLASS);
+    @Option(displayName = "`@Nullable` annotation class",
+            description = "The fully qualified name of the @Nullable annotation. The annotation should have `@Target(TYPE_USE)`. Defaults to `org.jspecify.annotations.Nullable`",
```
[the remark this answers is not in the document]
Laurens-W: 1. I think rewording it to `should be meta annotated with ...` would be better
2. I would create a separate recipe and look at whether we can provide a set of accepted values here (I've seen this done elsewhere, just can't think of where right now)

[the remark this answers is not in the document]
JohannisK: 1. Agreed, I'll make the change.
2. I've been in doubt about this... it's not unthinkable new "Nullable" frameworks might be released in the future, to for larger companies to have their own implementations.

src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
```diff
@@ -43,33 +50,39 @@ public String getDisplayName() {
...
+                "This recipe scans for methods that do not already have a `@Nullable` annotation and checks their return " +
+                "statements for potential null values. It also identifies known methods from standard libraries that may " +
+                "return null, such as methods from `Map`, `Queue`, `Deque`, `NavigableSet`, and `Spliterator`. " +
+                "The return of streams, or lambdas are not taken into account.";
     }
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
+        String fullyQualifiedName = nullableAnnotationClass != null ? nullableAnnotationClass : DEFAULT_NULLABLE_ANN_CLASS;
+        String fullyQualifiedPackage = fullyQualifiedName.substring(0, fullyQualifiedName.lastIndexOf('.'));
```
jevanlingen, line 65: This will break if someone would provide a non valid `nullableAnnotationClass`. I think we need to override the `validate` method of the recipe with something like:

```java
@Override
    public Validated<Object> validate() {
        return super.validate()
                .and(Validated.test("nullableAnnotationClass", "<message>", nullableAnnotationClass,
                        it -> it == null || iit.contains(".")));
    }
```

src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
```diff
@@ -43,33 +50,39 @@ public String getDisplayName() {
...
+        String simpleName = fullyQualifiedName.substring(fullyQualifiedName.lastIndexOf('.') + 1);
+        AnnotationMatcher nullableAnnotationMatcher = new AnnotationMatcher("@" + fullyQualifiedName);
```
jevanlingen, line 67: Inline this one (even if you use it twice).

src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
JohannisK: Done :)

src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
JohannisK: Done

jevanlingen (approved):

src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
```diff
@@ -43,33 +50,45 @@ public String getDisplayName() {
...
     public String getDescription() {
-        return "Add the `@org.jspecify.annotation.Nullable` to non-private methods that may return `null`. " +
-               "This recipe scans for methods that do not already have a `@Nullable` annotation and checks their return " +
-               "statements for potential null values. It also identifies known methods from standard libraries that may " +
-               "return null, such as methods from `Map`, `Queue`, `Deque`, `NavigableSet`, and `Spliterator`. " +
-               "The return of streams, or lambdas are not taken into account.";
+        return "Add `@Nullable` to non-private methods that may return `null`. " +
+                "By default `org.jspecify.annotations.Nullable` is used, but through the `nullableAnnotationClass` option a custom annotation can be provided." +
+                "When providing a custom `nullableAnnotationClass` that annotation should be meta annotated with `@Target(TYPE_USE)`." +
+                "This recipe scans for methods that do not already have a `@Nullable` annotation and checks their return " +
```
timtebeek, line 56: Would you mind adding a space after each of the end one line dots here? It's often forgotten, but when these strings get concatenated the missing spaces will make the docs hard to read.

Issue #356, opened 2024-10-10
`org.openrewrite.staticanalysis.NoDoubleBraceInitialization` drops constructor collection parameter


blipper:
```java
class A {
    void foo() {
        Map<String, String> a = Map.of("foo", "bar");
        Map<String, String> b = new HashMap<>(a) {{
            put("irrelevantKey", "irrelevantValue");
        }};
    }
}
```

## What did you expect to see?

```java
class A {
    void foo() {
        Map<String, String> a = Map.of("foo", "bar");
        Map<String, String> b = new HashMap<>(a);
        b.put("irrelevantKey", "irrelevantValue");
    }
}
```

## What did you see instead?

```java
class A {
    void foo() {
        Map<String, String> a = Map.of("foo", "bar");
        Map<String, String> b = new HashMap<>();   // <= No `a` here anymore
        b.put("irrelevantKey", "irrelevantValue");
    }
}
```

## What is the full stack trace of any errors you encountered?

```
stacktrace output here
```

arodionov:
The mentioned issue is not reproduced. 

But there is another one, in a class level initialisation: 

Before:
```
class A {                                               
     Map<String, String> a = Map.of("foo", "bar");  
     Map<String, String> b = new HashMap<>(a) {{    
          put("irrelevantKey", "irrelevantValue");   
     }};                                            
}                                                       
```

After:
```  
class A {
    Map<String, String> a = Map.of("foo", "bar");
    Map<String, String> b;
    {
        b = new HashMap<>();        // <= No `a` here anymore
        b.put("irrelevantKey", "irrelevantValue");
    }
}
```

PR #492, opened 2025-04-01
Fix missing constructor argument issue

arodionov:
- fixed issue https://github.com/openrewrite/rewrite-static-analysis/issues/356
- added tests


timtebeek (approved):
Nice! Love the simpler solution here.

PR #495, opened 2025-04-06
RemoveRedundantTypeCast should not remove required downcast

protocol7:
Adds a failing test case for where RemoveRedundantTypeCast should not remove a downcast that is necessary for chained calls to get matching types.


timtebeek:
Thanks for the runnable example!

protocol7:
Nice work!

PR #487, opened 2025-03-25
WIP: fix `EqualsAvoidsNull` not idempotent

pankratz76:
## What's changed?
test `EqualsAvoidsNull` idempotency

## What's your motivation?
* tests [EqualsAvoidsNull not idempotent #472](https://github.com/openrewrite/rewrite-static-analysis/issues/472)

PR #496, opened 2025-04-07
Increased idempotency of EqualsAvoidsNull by avoiding too many swaps

greg-at-moderne:
## What's changed?

Increasing the idempotency of `EqualsAvoidsNull` recipe. Specifically preventing thrashing between expressions like `FOO.equals(BAR)` changed to `BAR.equals(FOO)` changed to `FOO.equals(BAR)`, etc.

The test code has been taken from #487 (kudos to @pankratz76).

## What's your motivation?

- fixes additional problems reported in #472


pankratz76:
nice to see this going on, thanks.

PR #497, opened 2025-04-08
Reduce `complexity` and `coupling` by increasing `cohesion` in `EqualsAvoidsNull`

Pankraz76:
Reduce `complexity` and `coupling` by increasing `cohesion` in `EqualsAvoidsNull`

Push the well done approach provided by @greg-at-moderne further to simply and separate the concerns done in #496.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
@@ -40,10 +40,14 @@ public class EqualsAvoidsNull extends Recipe {
     private static final String JAVA_LANG_STRING = "java.lang.String";
     private static final String JAVA_LANG_OBJECT = "java.lang.Object";
 
-    private static final MethodMatcher EQUALS_STRING = new MethodMatcher(JAVA_LANG_STRING + " equals(" + JAVA_LANG_OBJECT + ")");
-    private static final MethodMatcher EQUALS_OBJECT = new MethodMatcher(JAVA_LANG_OBJECT + " equals(" + JAVA_LANG_OBJECT + ")");
-    private static final MethodMatcher EQUALS_IGNORE_CASE = new MethodMatcher(JAVA_LANG_STRING + " equalsIgnoreCase(" + JAVA_LANG_STRING + ")");
-    private static final MethodMatcher CONTENT_EQUALS = new MethodMatcher(JAVA_LANG_STRING + " contentEquals(java.lang.CharSequence)");
+    private static final MethodMatcher EQUALS_STRING = new MethodMatcher(JAVA_LANG_STRING +
+            " equals(" + JAVA_LANG_OBJECT + ")");
+    private static final MethodMatcher EQUALS_OBJECT = new MethodMatcher(JAVA_LANG_OBJECT +
+            " equals(" + JAVA_LANG_OBJECT + ")");
+    private static final MethodMatcher EQUALS_IGNORE_CASE = new MethodMatcher(JAVA_LANG_STRING +
+            " equalsIgnoreCase(" + JAVA_LANG_STRING + ")");
+    private static final MethodMatcher CONTENT_EQUALS = new MethodMatcher(JAVA_LANG_STRING +
+            " contentEquals(java.lang.CharSequence)");
 
     @Override
     public String getDisplayName() {
@@ -52,7 +56,9 @@ public class EqualsAvoidsNull extends Recipe {
 
     @Override
     public String getDescription() {
-        return "Checks that any combination of String literals is on the left side of an `equals()` comparison. Also checks for String literals assigned to some field (such as `someString.equals(anotherString = \"text\"))`.";
+        return "Checks that any combination of String literals is on the left side of an `equals()` comparison. " +
+                "Also checks for String literals assigned to some field " +
+                "(such as `someString.equals(anotherString = \"text\"))`.";
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +91,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                    private @NotNull J literalsFirstInComparisonsNullCheck(J.MethodInvocation m,
+                                                                           Expression firstArgument) {
                         return firstArgument.getType() == JavaType.Primitive.Null ?
                                 literalsFirstInComparisonsNull(m, firstArgument) :
                                 literalsFirstInComparisons(m, firstArgument);
-
                     }
 
-                    private boolean hasCompatibleArgument(J.MethodInvocation m) {
-                        if (m.getArguments().isEmpty()) {
```
pankratz76, line 95: ## 🛠️ Fix - Structure & Clarity

These cases are never on the same level – it's always either/or. By avoiding early exits and using a positive program flow, we can rely on returning `false` from below, making the happy path clear right away.

By doing this, we use **indentation** to give **visual indication** for this special case.  

The idea is to **support** code understanding right away without needing to read the whole or boilerplate code in full detail.

This is **not** what we want to support in the first place –  
We try to follow:
- **SOC/SRP** (Separation of Concerns / Single Responsibility Principle)
- Avoid **feature envy**
- Do **one thing only** in functions

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -171,14 +200,13 @@ private J.Binary literalsFirstInComparisonsNull(J.MethodInvocation m,
 
                     private J.MethodInvocation literalsFirstInComparisons(J.MethodInvocation m,
                                                                           Expression firstArgument) {
-                        if (!(firstArgument instanceof J.Literal) && !(m.getSelect() instanceof J.Literal)) {
```
pankratz76, line 174: Try to provide the conditions with proper inline comments or extract them into cleanly named methods. In my opinion, this is better than using overhead in double `if` statements that are technically the same. There’s no need to split them explicitly.

In this case, it’s the opposite, but kind of the same as above. It’s always the same, so we want to communicate this right away. In this case, the indentation and additional `if` are, in my eyes, boilerplate and overhead as well.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -17,12 +17,23 @@
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import org.jetbrains.annotations.NotNull;
 import org.jspecify.annotations.Nullable;
-import org.openrewrite.*;
```
pankratz76, line 21: lacking config fixed in: #498

pankratz76:
kindly request your feedback on this @timtebeek @greg-at-moderne

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +90,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                    private boolean hasCompatibleArgument(J.MethodInvocation m) {
-                        if (m.getArguments().isEmpty()) {
-                            return false;
-                        }
-                        Expression firstArgument = m.getArguments().get(0);
-                        if (firstArgument instanceof J.Literal) {
-                            return true;
-                        }
-                        if (firstArgument instanceof J.FieldAccess) {
-                            firstArgument = ((J.FieldAccess) firstArgument).getName();
```
pankratz76, line 103: # Immutable Code Principles

Having variables is never a good thing and should be avoided in favor of the final state, as this code structure is much more stable and adheres to best practices.

What can be `unmodifiable` and `final` should be, to make things `clear` and `avoid undesirable code` that often leads to **side effects** due to **unstable** state. 

Overbearing, clustered code that does too many things together and is coupled over the same fields can lead to the worst-case scenario: changing everything everywhere.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -52,7 +56,9 @@ public String getDisplayName() {
 
     @Override
     public String getDescription() {
-        return "Checks that any combination of String literals is on the left side of an `equals()` comparison. Also checks for String literals assigned to some field (such as `someString.equals(anotherString = \"text\"))`.";
```
pankratz76, line 55: TLTR

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +80,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                                return isStaticOrFinal(((J.FieldAccess) firstArgument).getName().getFieldType());
+                            } else if (firstArgument instanceof J.Identifier) {
+                                return isStaticOrFinal(((J.Identifier) firstArgument).getFieldType());
+                            }
                         }
                         return false;
                     }
 
+                    private boolean isStaticOrFinal(JavaType.@Nullable Variable fieldType) {
+                        return fieldType != null && fieldType.hasFlags(Flag.Static, Flag.Final);
```
pankratz76, line 121: hide impl. detail

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +80,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                            } else if (firstArgument instanceof J.FieldAccess) {
+                                return isStaticOrFinal(((J.FieldAccess) firstArgument).getName().getFieldType());
+                            } else if (firstArgument instanceof J.Identifier) {
+                                return isStaticOrFinal(((J.Identifier) firstArgument).getFieldType());
+                            }
                         }
                         return false;
                     }
 
+                    private boolean isStaticOrFinal(JavaType.@Nullable Variable fieldType) {
```
pankratz76, line 120: good story telling

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +80,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                                m.getSelect() instanceof J.Literal) {
                             return m;
                         }
-
                         maybeHandleParentBinary(m, getCursor().getParentTreeCursor().getValue());
-                        Expression firstArgument = m.getArguments().get(0);
+                        return literalsFirstInComparisonsNullCheck(m, m.getArguments().get(0));
+                    }
 
+                    private J literalsFirstInComparisonsNullCheck(J.MethodInvocation m,
```
pankratz76, line 99: SOC

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +80,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
                 new JavaVisitor<ExecutionContext>() {
+
                     @Override
                     public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
-                        J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
+                        return visitMethodInvocation((J.MethodInvocation) super.visitMethodInvocation(method, ctx));
+                    }
 
-                        if (!isStringComparisonMethod(m) || !hasCompatibleArgument(m) || m.getSelect() instanceof J.Literal) {
+                    private J visitMethodInvocation(J.MethodInvocation m) {
```
pankratz76, line 89: SOC - as here is the action happening; we get the data we **actually need** and extract the DTO stuff to outer world we do not care and know about.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +80,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
                     private void maybeHandleParentBinary(J.MethodInvocation m, final Tree parent) {
-                        if (parent instanceof J.Binary) {
-                            if (((J.Binary) parent).getOperator() == J.Binary.Type.And &&
-                                    ((J.Binary) parent).getLeft() instanceof J.Binary) {
-                                J.Binary potentialNullCheck = (J.Binary) ((J.Binary) parent).getLeft();
-                                if (isNullLiteral(potentialNullCheck.getLeft()) &&
-                                        matchesSelect(potentialNullCheck.getRight(), requireNonNull(m.getSelect())) ||
-                                    isNullLiteral(potentialNullCheck.getRight()) &&
-                                            matchesSelect(potentialNullCheck.getLeft(), requireNonNull(m.getSelect()))) {
-                                    doAfterVisit(new JavaVisitor<ExecutionContext>() {
```
pankratz76, line 128: 3 ifs (indentations are 3 concerns) and countless conditions later in **one block/context** we doing some `doAfterVisit` is just too much.

src/main/java/org/openrewrite/staticanalysis/EqualsAvoidsNull.java
```diff
@@ -74,83 +80,95 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                                                done = true;
-                                                return binary.getRight().withPrefix(binary.getPrefix());
-                                            }
-                                            return super.visitBinary(binary, ctx);
-                                        }
-                                    });
+                        if (parent instanceof J.Binary &&
+                                ((J.Binary) parent).getOperator() == J.Binary.Type.And &&
+                                ((J.Binary) parent).getLeft() instanceof J.Binary) {
+                            potentialNullCheck(m, (J.Binary) parent, (J.Binary) ((J.Binary) parent).getLeft());
```
pankratz76, line 136: here we know whats going on `potentialNullCheck` reducing the mental load into smaller scopes making it easy/possible to digest.

pankratz76:
if you want, lets have a quick call about this and review together.
As its just `extract till you drop` its easy to `reproduce and showcase`. @greg-at-moderne

PR #498, opened 2025-04-08
Fix format in test

pankratz76:


src/test/java/org/openrewrite/staticanalysis/EqualsAvoidsNullTest.java
```diff
@@ -25,604 +25,613 @@
...
-                """
-                  public class Foo {
-                      private static final String FOO = null;
-                      private static final String BAR = null;
-                      public void bar() {
-                          BAR.equals(FOO);
-                      }
-                  }
-                  """
-              )
```
pankratz76, line 556: we already have 2 spaced indentations mixed with 4. I do not care whats the value, but it should be set and done.

Checkstyle is doing in its test this smaller scope too and google is on this train too. So we might hop on: https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation

pankratz76:
kindly request your feedback on this @timtebeek @greg-at-moderne

timtebeek:
hi @pankratz76 ; appreciate the help; it's been proving difficult to get to all the reviews lately, and then we tend to prioritize those that fix an issue or add a feature over those that purely improve the code. Appreciate your patience as we work through a backlog of issues, as I know you've contributed quite a few into that last category recently. With those larger restructurings it's also quite a bit of effort to ensure it all functionally works the same, as we can't solely rely on the unit tests that we have, but need to verify these at scale as well.

pankratz76:
> at scale as well

automation is key, as we cannot rely on something we do not measure.
So if we can leverage something like [gatling](https://docs.gatling.io/tutorials/scripting-intro/) this would help us to get ritt of this issue right away.

Issue #513, opened 2025-04-16
CompareEnumsWithEqualityOperator inverts conditions in some cases


julmud:
## What version of OpenRewrite are you using?

I am using the maven plugin with the RELEASE version (so hopefully the latest one) directly from the command line.

## How are you running OpenRewrite?

I am using the Maven plugin via the command line, and my project is a single module project.
```bash
mvn org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-static-analysis:RELEASE -Drewrite.activeRecipes=org.openrewrite.staticanalysis.CompareEnumsWithEqualityOperator
```

## What is the smallest, simplest way to reproduce the problem?

```java
public class App
{
    private enum EnumA {
        VALUE_A, VALUE_B;
    }

    private static class ObjectA {
        private EnumA enumA;
        public EnumA getEnumA() {
            return enumA;
        }

        public ObjectA(EnumA enumA) {
            this.enumA = enumA;
        }
    }

    public static void case1() {
        List<ObjectA> list = List.of(new ObjectA(EnumA.VALUE_A), new ObjectA(EnumA.VALUE_B));
        /* Input code:
         *   if (!list.stream().filter(e -> e.getEnumA().equals(EnumA.VALUE_A)).toList().isEmpty()) {
         * is wrongly rewritten by the CompareEnumsWithEqualityOperator rule as:
         *   if (list.stream().filter(e -> e.getEnumA() != EnumA.VALUE_A).toList().isEmpty()) {
         */
        if (!list.stream().filter(e -> e.getEnumA().equals(EnumA.VALUE_A)).toList().isEmpty()) {
            System.out.println("Not empty");
        }
    }

    public static void case2() {
        EnumA value1 = null;
        EnumA value2 = null;
        EnumA value3 = null;
        EnumA value4 = null;

        /* Input code:
         *   boolean hasMultipleValues = !(value1 != null ? value1.equals(value2) : (value2 == null)) ||
         *                              !(value3 != null ? value3.equals(value4) : (value4 == null));
         * is wrongly rewritten by the CompareEnumsWithEqualityOperator rule as:
         *   boolean hasMultipleValues = value1 != null ? value1 != value2 : (value2 == null) ||
         *                               value3 != null ? value3 != value4 : (value4 == null);
         */
[...]

PR #516, opened 2025-04-16
CompareEnumsWithEqualityOperator should not invert equality

JohannisK:
## What's changed?
- Added tests proving bug
- Added checks for ternary expressions
- Added checks for lambdas

## What's your motivation?
Fixes:
 - https://github.com/openrewrite/rewrite-static-analysis/issues/513

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperatorTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperatorTest.java
@@ -272,7 +272,51 @@ class CompareEnumsWithEqualityOperatorTest implements RewriteTest {
               import java.util.List;
               class Test {
                   void method(List<ObjectA> list) {
+                      if (list.stream().filter(e -> e.getEnumA().equals(A.FOO)).toList().isEmpty()) {}
+                      if (list.stream().filter(e -> !(e.getEnumA().equals(A.FOO))).toList().isEmpty()) {}
+                  }
+              }
+              """,
+            """
+              import a.A;
+              import java.util.List;
+              class Test {
+                  void method(List<ObjectA> list) {
+                      if (list.stream().filter(e -> e.getEnumA() == A.FOO).toList().isEmpty()) {}
+                      if (list.stream().filter(e -> e.getEnumA() != A.FOO).toList().isEmpty()) {}
+                  }
+              }
+              """
+          )
+        );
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperator.java
```diff
@@ -56,7 +56,7 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
                 if (enumEquals.matches(m) && m.getSelect() != null) {
-                    Cursor parent = getCursor().dropParentUntil(is -> is instanceof J.Unary || is instanceof J.Block || is instanceof J.Binary);
+                    Cursor parent = getCursor().dropParentUntil(is -> is instanceof J.Unary || is instanceof J.Block || is instanceof J.Binary || is instanceof J.Ternary || is instanceof J.Lambda);
```
JohannisK, line 59: I focussed on a quick bug fix for the reported issue, but haven't given extensive thought whether or not this list is now complete.
Should I do that?

src/test/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperatorTest.java
```diff
@@ -247,4 +247,156 @@ void method(A arg0) {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/513")
```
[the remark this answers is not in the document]
greg-at-moderne: I think it's fine to link. We prefer concise unit tests (for multiple reasons). The example in the original issue wasn't that concise. So what you suggest here is actually good.

src/main/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperator.java
greg-at-moderne: Nothing comes to my mind, so I cannot suggest anything specific.
I think we are fine. I personally support positive changes to the code, even if they are not perfect (i.e. they don't solve all possible problems).

greg-at-moderne (approved):

Issue #315, opened 2024-07-13
`NeedsBraces` deletes comments, or moves comments around in confusing, inconsistent way


valters:
## What version of OpenRewrite are you using?

I am using

- Maven plugin 5.35.0
- rewrite-static-analysis 1.11.0

## How are you running OpenRewrite?
Maven plugin.
Minimal project to repro the issue: https://github.com/valters/test-openrewrite
Please see steps in readme.

## What is the smallest, simplest way to reproduce the problem?

I am trying to rewrite following code:

```java
import java.io.File;

public class WithCommentPlacingInBraces {

  public void shouldKeepTrackOfComments(boolean success) {

    // keep track of comments corresponding to lines
    if (success)
      System.out.println("success comment belongs here: but somehow failure comment ends up here"); // success!
    else
      System.out.println("failure comment belongs here: but somehow comment at end of line gets moved before else?"); // failure! this should stay here, firmly inside `else` block
    // this line should stay in place. the comments at end of line, above, should not get mixed up! but somehow this gets moved into `if {}` block?
  }

  public int shouldKeepTrackOfBlockComment(File f) {

    if (f.length() > 0) {
      System.out.println("file is not empty");
    }
    /*
     * this comment should not be lost.
     * it would be great if it went inside the `else {}` block.
     */
    else
      System.out.println("where did the comment go?");
    return 0;
  }

  public int shouldKeepTrackOfComments(String s) {
    if ("ok".equals(s))
      return 1;
      // this comment should not be lost.
      // it would be great if it went inside the `if {}` block.
    else {
      System.out.println("where did the comment go?");
      // nothing
    }
    return 0;
  }
}
```

## What did you expect to see?

Comments should be preserved.
If comments are moved, they should move into block that they logically belong to.

## What did you see instead?

```diff
diff --git a/source-problems/src/main/java/ch/vingolds/testopenrewrite/WithCommentPlacingInBraces.java b/source-problems/src/main/java/ch/vingolds/testopenrewrite/WithCommentPlacingInBraces.java
index 8dba36a..485fdb2 100644
--- a/source-problems/src/main/java/ch/vingolds/testopenrewrite/WithCommentPlacingInBraces.java
+++ b/source-problems/src/main/java/ch/vingolds/testopenrewrite/WithCommentPlacingInBraces.java
@@ -7,33 +7,29 @@ public class WithCommentPlacingInBraces {
   public void shouldKeepTrackOfComments(boolean success) {
 
[...]

PR #518, opened 2025-04-16
Fix Comment handling in NeedsBraces

Jammy-Louie:
## What's changed?

Improve comment handling for NeedBraces recipe for if statements. Comments would be placed incorrectly or dropped if they trailed the if statement.

## What's your motivation?

- fixes https://github.com/openrewrite/rewrite-static-analysis/issues/315

### Checklist


src/main/java/org/openrewrite/staticanalysis/NeedBraces.java
```diff
@@ -69,20 +69,42 @@ private static class NeedBracesVisitor extends JavaIsoVisitor<ExecutionContext>
          * We can use that to our advantage by saying if you aren't a block (e.g. a single {@link Statement}, etc.),
          * then we're going to make this into a block. That's how we'll get the code bodies surrounded in braces.
          */
-        private <T extends Statement> J.Block buildBlock(Statement owner, T element) {
-            J j = getCursor().getParentTreeCursor().getValue();
+        private <T extends Statement> J.Block buildBlock(T element) {
```
Jammy-Louie, line 72: remove owner as an argument because it is no longer needed

src/main/java/org/openrewrite/staticanalysis/NeedBraces.java
```diff
@@ -69,20 +69,42 @@ private static class NeedBracesVisitor extends JavaIsoVisitor<ExecutionContext>
...
+                int currentIndex = statements.indexOf(rootElement);
+                boolean last = currentIndex == statements.size() - 1;
+                Space trailingComment = last ? block.getEnd() : statements.get(currentIndex + 1).getPrefix();
+
+                if (!trailingComment.isEmpty() && !trailingComment.getWhitespace().contains("\n")) {
+                    end = trailingComment;
+                    if (last) {
+                        blockParentCursor.putMessage("removeEndComments", true);
+                    } else {
+                        blockParentCursor.<List<Integer>>computeMessageIfAbsent("replaced", k -> new ArrayList<>()).add(currentIndex);
```
Jammy-Louie, line 103: messages can appear at the end of the block for cases such as.

```
static void method() {
    if (true) return; // if comment
    else return; // else comment
}
```

for this case the `// else comment` would be located at the `end` of the block

src/main/java/org/openrewrite/staticanalysis/NeedBraces.java
```diff
@@ -111,6 +133,11 @@ private <T extends Statement> J.Block buildBlock(Statement owner, T element) {
         @Override
         public J.Block visitBlock(J.Block block, ExecutionContext ctx) {
             J.Block bl = super.visitBlock(block, ctx);
+            Boolean removeEndComments = getCursor().pollMessage("removeEndComments");
+            if (removeEndComments != null) {
+                bl = bl.withEnd(bl.getEnd().withComments(Collections.emptyList()));
+                bl = maybeAutoFormat(block, bl, ctx);
+            }
```
Jammy-Louie, line 140: remove comments when the trailing comments found at the end of the block

src/main/java/org/openrewrite/staticanalysis/NeedBraces.java
```diff
@@ -140,7 +167,15 @@ public J.If visitIf(J.If iff, ExecutionContext ctx) {
...
-                J.Block b = buildBlock(elem, elem.getThenPart());
+                J.Block b;
+                if (elem.getElsePart() != null && !elem.getElsePart().getPrefix().getComments().isEmpty()) {
+                    Space end = elem.getElsePart().getPrefix();
+                    elem = elem.withElsePart(elem.getElsePart().withPrefix(Space.EMPTY));
+                    b = buildBlock(elem.getThenPart()).withEnd(end);
+                } else {
+                    b = buildBlock(elem.getThenPart());
+                }
+
```
Jammy-Louie, line 178: trailing comments for if statements can appear as a prefix in the `else`

example:
```
static void method() {
    if (true) return; // if comment
    else return; // else comment
}
```

src/main/java/org/openrewrite/staticanalysis/NeedBraces.java
```diff
@@ -155,7 +190,14 @@ public J.If.Else visitElse(J.If.Else else_, ExecutionContext ctx) {
             J.If.Else elem = super.visitElse(else_, ctx);
             boolean hasAllowableBodyType = elem.getBody() instanceof J.Block || elem.getBody() instanceof J.If;
             if (!needBracesStyle.getAllowSingleLineStatement() && !hasAllowableBodyType) {
-                J.Block b = buildBlock(getCursor().getParentTreeCursor().getValue(), elem.getBody());
+                Space prefix = elem.getPrefix();
+                Statement body = elem.getBody();
+
+                if(!prefix.getComments().isEmpty() && prefix.getWhitespace().contains("\n")) {
+                    body = body.withPrefix(prefix);
+                }
```
Jammy-Louie, line 198: this handles the case where braces appear on the if statement but not the else

```
if (true) {
    return;
} 
/*
* do something
/*
else return;

**Comment by Jammy-Louie** [comment]
testing locally on `openrewrite`

only 1 change suggestion and appears to be correct

Screenshot 2025-04-22 at 11 10 53 AM

**Comment by Jammy-Louie** [comment]
testing locally on `apache`

suggested changes appear to be correct.

Sample change:

Screenshot 2025-04-22 at 11 13 58 AM

although there does appear to be issues when running against Javascript files:

Screenshot 2025-04-22 at 11 17 11 AM

Issue appears to be present before my changes as I've tested the recipe from main and still see the same results.

File: src/main/java/org/openrewrite/staticanalysis/NeedBraces.java [hunk 6]
```diff
@@ -69,20 +69,42 @@ private static class NeedBracesVisitor extends JavaIsoVisitor<ExecutionContext>
...
-                J.Block block = (J.Block) j;
+
+            Cursor currentCursor = getCursor();
+            while (
+                    currentCursor != null &&
+                            currentCursor.getParent() != null &&
+                            !(currentCursor.getParent().getValue() instanceof J.Block)
+            ) {
+                currentCursor = currentCursor.getParent();
+            }
```
greg-at-moderne, line 83: `dropParentUntil` is likely a 1-1 replacement for this

Jammy-Louie: `dropParentUntil` doesn't quite do what I need to do as it will return the cursor on the predicate condition on its value, `instanceof J.Block`. I need it to return the cursor when the parent is is `instanceof J.Block`.

I couldn't really use `JRightPadded` because it could appear multiple times because it could be nested as mentioned here https://github.com/openrewrite/rewrite-static-analysis/pull/518#discussion_r2054259090

greg-at-moderne: Ah. Forgot the `Cursor` class is not strongly typed :(. It uses `Object`

src/main/java/org/openrewrite/staticanalysis/NeedBraces.java
```diff
@@ -111,6 +133,11 @@ private <T extends Statement> J.Block buildBlock(Statement owner, T element) {
         @Override
         public J.Block visitBlock(J.Block block, ExecutionContext ctx) {
             J.Block bl = super.visitBlock(block, ctx);
+            Boolean removeEndComments = getCursor().pollMessage("removeEndComments");
+            if (removeEndComments != null) {
```
Jenson3210, line 137: Should we not use the `Boolean.TRUE.equals(Object)` to not have a classCastException when this message would ever be incorrectly set to a String/int....
Also, if the message would be set to false, the if block is still executing in the as-is implementation. 

```suggestion
            if (Boolean.TRUE.equals(getCursor().pollMessage("removeEndComments"))) {
```

Jenson3210 (approved):
I see @timtebeek has pushed my proposed change, so other that this LGTM!

src/test/java/org/openrewrite/staticanalysis/NeedBracesTest.java
```diff
@@ -323,6 +333,177 @@ static void method() {
...
                   }
+                  static void commentWhile(Object obj) {
+                      while (true) { // while comment
+                      }
+                      while (true) {
+                          return; // while comment with body
+                      }
+                      do {
+                          obj.notify(); // do while comment
+                      } while (true);
```
timtebeek, line 344: Certainly understand choices have to be made at some point, but here I had sort of expected
```suggestion
                          obj.notify();
                      } while (true); // do while comment
```

src/test/java/org/openrewrite/staticanalysis/NeedBracesTest.java
```diff
@@ -323,6 +333,177 @@ static void method() {
...
+              class Test {
+                  static void method() {
+                      if (true) {
+                          return;
+                      } else {
+                          /*
+                           * comment should be in else block
+                           */
+                          return;
+                      }
```
timtebeek, line 392: Similarly, I had expected the following to minimize the diff with what was there before. All we ought to be doing is adding braces, where possible.
```suggestion
                      }
                      /*
                       * comment should be in else block
                       */
                      else {
                          return;
                      }
```

timtebeek (commented):
Some optional suggestions, fully realize these might be awkward to make work, and this is already a good improvement, so we need not drag it out. Could be worth briefly exploring before a merge.

PR #520, opened 2025-04-17
DefaultComesLast should correctly support fall through cases

JohannisK:
## What's changed?
Reimplemented logic in DefaultComesLast Visitor

## What's your motivation?
The code now assumes every case is a fall through case, best case scenario it falls through 0 times.
It handles the bug correctly.
 
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/4

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -100,7 +100,7 @@ public class DefaultComesLastVisitor<P> extends JavaIsoVisitor<P> {
         return casesWithDefaultLast;
     }
 
-    private @NonNull List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
+    private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
         J.Case defaultCase = null;
         List<J.Case> preDefault = new ArrayList<>();
         List<J.Case> postDefault = new ArrayList<>();
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -32,7 +32,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.toList;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
@@ -44,7 +45,7 @@ public class DefaultComesLastVisitor<P> extends JavaIsoVisitor<P> {
         J.Switch s = visitAndCast(switch_, p, super::visitSwitch);
 
         if (!isDefaultCaseLastOrNotPresent(switch_)) {
-            List<J.Case> cases = s.getCases().getStatements().stream().map(J.Case.class::cast).collect(Collectors.toList());
+            List<J.Case> cases = s.getCases().getStatements().stream().map(J.Case.class::cast).collect(toList());
             List<J.Case> casesWithDefaultLast = maybeReorderCases(cases, p);
             if (casesWithDefaultLast != null) {
                 boolean changed = true;
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -114,7 +114,7 @@ public class DefaultComesLastVisitor<P> extends JavaIsoVisitor<P> {
                 defaultStatements.addAll(aCase.getStatements());
             }
         }
-        List<Statement> defaultCaseStatements = defaultCase.getStatements();
+        List<Statement> defaultCaseStatements = defaultCase != null ? defaultCase.getStatements() : Collections.emptyList();
         defaultStatements.addAll(0, defaultCaseStatements);
         List<J.Case> fixedCases = new ArrayList<>(preDefault.size() + postDefault.size() + 1);
         fixedCases.addAll(maybeUpdatePreDefaultCases(preDefault, defaultCaseStatements));
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -17,6 +17,7 @@ package org.openrewrite.staticanalysis;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import org.jetbrains.annotations.NotNull;
 import org.openrewrite.Tree;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
@@ -26,9 +27,7 @@ import org.openrewrite.java.tree.Space;
 import org.openrewrite.java.tree.Statement;
 import org.openrewrite.marker.Markers;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 import java.util.stream.Collectors;
 
 @Value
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/DefaultComesLastTest.java
```diff
@@ -320,4 +353,148 @@ static int coverage(Object obj) {
...
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/4")
+    @Test
+    void moveDefaultToLastWithFallThroughAndStatements() {
```
JohannisK, line 401: Is this behavior desirable?

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -42,116 +45,157 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
...
+
+    private boolean isFallthroughCase(J.Case aCase) {
+        return aCase.getStatements().isEmpty() ||
+              !(aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Break ||
+                    aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Continue ||
+                    aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Return ||
+                    aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Throw);
+    }
+
+    private J.Case removeCaseEnd(J.Case aCase) {
```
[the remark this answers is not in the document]
JohannisK: Renamed to removeBreak

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -42,116 +45,157 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
...
 
-            boolean changed = true;
-            if (cases.size() == fixedCases.size()) {
-                changed = false;
-                for (int i = 0; i < cases.size(); i++) {
-                    if (cases.get(i) != fixedCases.get(i)) {
-                        changed = true;
-                        break;
-                    }
+    private J.Switch maybeUpdateSwitch(List<J.Case> cases, List<J.Case> casesWithDefaultLast, J.Switch s) {
```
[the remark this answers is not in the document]
JohannisK: Me neither :)

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -42,116 +45,139 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
 
         if (!isDefaultCaseLastOrNotPresent(switch_)) {
             List<J.Case> cases = s.getCases().getStatements().stream().map(J.Case.class::cast).collect(Collectors.toList());
```
jevanlingen, line 47:
```suggestion
            List<J.Case> cases = ListUtils.map(s.getCases().getStatements(), J.Case.class::cast));
```

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -41,117 +45,140 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
...
-                    fixedCases.add(0, aCase);
+                if (changed) {
+                    s = s.withCases(s.getCases().withStatements(casesWithDefaultLast.stream().map(Statement.class::cast).collect(toList())));
                 }
             }
+        }
+        return s;
+    }
+
+    private @Nullable List<J.Case> maybeReorderCases(List<J.Case> cases, P p) {
```
jevanlingen, line 70: I would go with a non-nullable here and return an `emptyList()`.

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -41,117 +45,140 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
...
-                        new J.Break(Tree.randomId(), Space.EMPTY, Markers.EMPTY, null),
-                        p
-                );
-                statementsOfCaseBeingMoved.add(breakStatement);
-
-                lastNotGroupedWithDefault = lastNotGroupedWithDefault.withStatements(
-                        ListUtils.map(statementsOfCaseBeingMoved, stmt -> autoFormat(stmt, p, getCursor()))
-                );
-                fixedCases.set(fixedCases.size() - 1, lastNotGroupedWithDefault);
+    private @NonNull List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
```
jevanlingen, line 103:
```suggestion
    private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
```

No need for this annotation, as there is already a JSpecify package-info.java file in this folder.

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -41,117 +45,140 @@ public J.Switch visitSwitch(J.Switch switch_, P p) {
...
-            fixedCases.addAll(casesGroupedWithDefault);
-            if (defaultCase != null) {
-                if (defaultCase.getStatements().stream().reduce((s1, s2) -> s2)
-                        .map(stat -> stat instanceof J.Break || stat instanceof J.Continue || isVoidReturn(stat))
-                        .orElse(false)) {
-                    List<Statement> fixedDefaultStatements = new ArrayList<>(defaultCase.getStatements());
-                    fixedDefaultStatements.remove(fixedDefaultStatements.size() - 1);
-                    fixedCases.add(defaultCase.withStatements(fixedDefaultStatements));
+        }
+        List<Statement> defaultCaseStatements = defaultCase.getStatements();
```
jevanlingen, line 118: Method invocation 'getStatements' may produce 'NullPointerException'

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
JohannisK: It's a bit ugly, but null has special meaning. It means we found a switch case with a body, and should return the original (t's completely ignored)

jevanlingen (approved):

src/test/java/org/openrewrite/staticanalysis/DefaultComesLastTest.java
timtebeek: hmm; I'm not so sure; imagine there's 15 lines of code instead of one print statement; then you'd be duplicating a lot of logic, as part of a recipe that should only reorder, or make no change at all.

greg-at-moderne: My way of thinking was that the user explicitly asked for this when they called 'DefaultComesLast'.

(no strong feelings, feel free to ignore my comments)

timtebeek (changes requested):
I've discussed this with Johan, he's going to explore two options
1. Either move the whole block of fall through cases that include `default` to the end
2. Or not make any change at all when there's fallthrough we can't handle well.

src/test/java/org/openrewrite/staticanalysis/DefaultComesLastTest.java
JohannisK: I've changed the logic, and added tests to show the behavior.
Now it should:
- Move entire fall though blocks to the bottom if they contain a default statement
- If the default case can be moved down without touching statements the default case is moved down (it checks this by verifying the default case and all cases following the default case (except for the last one) do not contain statements themselves. If that is true statements of the last case are moved to the default case, and the default case is moved down.

JohannisK:
> I've discussed this with Johan, he's going to explore two options
> 
> 1. Either move the whole block of fall through cases that include `default` to the end
> 2. Or not make any change at all when there's fallthrough we can't handle well.

I've adjusted the recipe: https://github.com/openrewrite/rewrite-static-analysis/pull/520#discussion_r2050695648

timtebeek:
As discussed might be good to test this against some well known projects still before we merge; `apache/poi` could be interesting as they have a convention of default first, so that gives quite some cases to ealuate.

sambsnyd:
There may well be further improvements possible but this looks like it's clearly an improvement over the existing form of the recipe. Thanks @JohannisK !

PR #523, opened 2025-04-23
Fixes for DefaultComesLast

JohannisK:
## What's changed?
Resolved small issues with DefaultComesLast

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -100,7 +100,7 @@ public class DefaultComesLastVisitor<P> extends JavaIsoVisitor<P> {
         return casesWithDefaultLast;
     }
 
-    private @NonNull List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
+    private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
         J.Case defaultCase = null;
         List<J.Case> preDefault = new ArrayList<>();
         List<J.Case> postDefault = new ArrayList<>();
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -159,7 +159,7 @@ public class DefaultComesLastVisitor<P> extends JavaIsoVisitor<P> {
     }
 
     private J.Case removeBreak(J.Case aCase) {
-        if (!aCase.getStatements().isEmpty() && aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Break) {
+        if (!aCase.getStatements().isEmpty() && aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Break && ((J.Break) aCase.getStatements().get(aCase.getStatements().size() - 1)).getLabel() == null) {
             aCase = aCase.withStatements(aCase.getStatements().subList(0, aCase.getStatements().size() - 1));
         }
         return aCase;
--- a/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
@@ -32,7 +32,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.toList;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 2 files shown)


JohannisK:
I think it's functionally correct, but it does misbehave while formatting. I can't seem to reproduce the issues in the test.
It makes the following change where the indentation seems wrong. I don't see a reason for this when looking at the original source: https://github.com/apache/cassandra-accord/blob/trunk/accord-maelstrom/src/main/java/accord/maelstrom/MaelstromRequest.java
```
diff --git a/accord-maelstrom/src/main/java/accord/maelstrom/MaelstromRequest.java b/accord-maelstrom/src/main/java/accord/maelstrom/MaelstromRequest.java
index 01621f7..10fdda3 100644
--- a/accord-maelstrom/src/main/java/accord/maelstrom/MaelstromRequest.java
+++ b/accord-maelstrom/src/main/java/accord/maelstrom/MaelstromRequest.java
@@ -122,15 +122,16 @@ org.openrewrite.staticanalysis.DefaultComesLast
             Key key = MaelstromKey.readKey(in);
             switch (op)
             {
-                default: throw illegalState("Invalid op: " + op);
                 case "r":
                     in.nextNull();
                     buildReadKeys.add(key);
                     break;
                 case "append":
-                    Datum value = Datum.read(in);
-                    buildKeys.add(key);
-                    update.merge(key, new Value(value), Value::append);
+                Datum value = Datum.read(in);
+                buildKeys.add(key);
+                update.merge(key, new Value(value), Value::append);
+                break;
+                default: throw illegalState("Invalid op: " + op);
             }
             in.endArray();
         }
```

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -133,33 +133,35 @@ private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
 
     private List<J.Case> addBreakToLastCase(List<J.Case> cases, P p) {
         return ListUtils.mapLast(cases, e -> {
-            if (isFallthroughCase(e)) {
+            if (e != null && isFallthroughCase(e.getStatements())) {
```
jevanlingen, line 136: `e` cannot be null here, so no need for this check!

---

FYI: The reason why it's shown as nullable is because the mapList second argument is defined as a `UnaryOperator<@Nullable T>`. We already fixed this by adding a secondary mapList function where this argument is listed as `Function<T, @Nullable T>`, but the original function is still there for backwards compatibility.

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -133,33 +133,35 @@ private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
...
     }
 
-    private boolean isFallthroughCase(J.Case aCase) {
-        return aCase.getStatements().isEmpty() ||
-              !(aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Break ||
-                    aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Continue ||
-                    aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Return ||
-                    aCase.getStatements().get(aCase.getStatements().size() - 1) instanceof J.Throw);
+    private boolean isFallthroughCase(List<Statement> statements) {
+        if(statements.isEmpty()) {
```
jevanlingen, line 144:
```suggestion
        if (statements.isEmpty()) {
```

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -133,33 +133,35 @@ private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
...
+            return true;
+        }
+        if(statements.get(statements.size() - 1) instanceof J.Block) {
```
jevanlingen, line 147:
```suggestion
        if (statements.get(statements.size() - 1) instanceof J.Block) {
```

src/main/java/org/openrewrite/staticanalysis/DefaultComesLastVisitor.java
```diff
@@ -133,33 +133,51 @@ private List<J.Case> maybeReorderFallthroughCases(List<J.Case> cases, P p) {
...
+        }
+        Statement lastStatement = statements.get(statements.size() - 1);
+        if (lastStatement instanceof J.Block) {
+            return isFallthroughCase(((J.Block) lastStatement).getStatements());
+        }
+        return !(lastStatement instanceof J.Break ||
+                lastStatement instanceof J.Continue ||
+                lastStatement instanceof J.Return ||
+                lastStatement instanceof J.Throw ||
+                lastStatement instanceof J.Yield);
```
JohannisK, line 155: @timtebeek I missed this one, but checked it out.
I'm not sure if it's relevant now, because we skip all switch expressions. I don't think it's wrong to leave this here for when we would want to support them in future.
Am I correct that yield has no meaning in traditional switch statements?

timtebeek (approved):
Nice round of improvements; let's merge this iteration as well and move onto something new. :)

PR #530, opened 2025-04-27
fix: use box for primitives when rewriting `instanceof` (#528)

### LINKED ISSUE #528: InstanceOfPatternMatch recipe replaces wrapper types with primitives in `instanceof`
It looks like the current PR https://github.com/openrewrite/rewrite-static-analysis/pull/489 replacing wrapper types with primitives in `instanceof`

Test to reproduce (should be added to `InstanceOfPatternMatchTest`: 

```
@Test
    void matchingPrimitiveVariableInBody() {
        rewriteRun(
          version(
          //language=java
          java(
            """
              public class A {
                  void test(Object o, Integer integer) {
                      if (o instanceof Integer) {
                          for (int j = 0; j < (int) o; j++) {
                          }
                      }
                  }
              }
              """,
            """
              public class A {
                  void test(Object o, Integer integer) {
                      if (o instanceof Integer integer1) {
                          for (int j = 0; j < integer1; j++) {
                          }
                      }
                  }
              }
              """
          ), 17)
        );
    }
```

The produced code is non-compilable:
```
public class A {
      void test(Object o, Integer integer) {
          if (o instanceof int integer1) {
              for (int j = 0; j < integer1; j++) {
              }
          }
      }
  } 
```

cortlepp:
## What's changed?

We now only use the boxed variants of primitives when rewriting `instanceof`.

## What's your motivation?

- fix #528 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
@@ -259,8 +259,16 @@ public class InstanceOfPatternMatch extends Recipe {
                 return instanceOf;
             }
             String name = patternVariableName(instanceOf, cursor);
-            TypeTree typeCastTypeTree = computeTypeTreeFromTypeCasts(instanceOf);
-            J currentTypeTree = instanceOf.getClazz();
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the immediate fix @cortlepp !

Issue #532, opened 2025-04-28
Incorrect type substitution for `var` variable


arodionov:
Incorrect type substitution in `instanceof` if a variable has type `var` in a body.

Having the following test for `InstanceOfPatternMatchTest`:

```
    @Test
    void varVariables() {
        rewriteRun(
          version(
            //language=java
            java(
              """
                public class A {
                    void test(Object o) {
                        if (o instanceof String) {
                            final var str = (String) o;
                            System.out.println(str.toUpperCase());
                        }
                    }
                }
                """,
              """
                public class A {
                    void test(Object o) {
                        if (o instanceof String str) {
                            System.out.println(str.toUpperCase());
                        }
                    }
                }
                """
            ), 17)
        );
    }
```

The actual output produced non-compiling code with incorrect type in `instanceof`:

```
public class A {
    void test(Object o) {
        if (o instanceof var str) {
            System.out.println(str.toUpperCase());
        }
    }
}
```

PR #533, opened 2025-04-28
Fix type substitution for `var` variable

arodionov:
- fix https://github.com/openrewrite/rewrite-static-analysis/issues/532


timtebeek (approved):

arodionov:
@cortlepp  added a minor additional fix

Issue #481, opened 2025-03-24
InstanceOfPatternMatch can produce name clashes


cortlepp:
## What version of OpenRewrite are you using?

I am using

- Maven plugin v6.3.2
- rewrite-migrate-java v3.4.0

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.3.2</version>
    <configuration>
        <exportDatatables>true</exportDatatables>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.InstanceOfPatternMatch</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>3.4.0</version>
        </dependency>
    </dependencies>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
package org.example;

public class Main {
    public class Bar {}
    public class FooBar {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar) {
            System.out.println(((Bar)o));
            if (o instanceof FooBar) {
                System.out.println(((FooBar)o));
            }
        }
    }
}
```

## What did you expect to see?

```java
package org.example;


public class Main {
    public class Bar {}
    public class FooBar {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar bar) {
            System.out.println(bar);
            if (o instanceof FooBar fooBar) {
                System.out.println(fooBar);
            }
        }
    }
}
```

## What did you see instead?

```java
package org.example;

public class Main {
    public class Bar {}
    public class FooBar {}
    public static void main(String[] args) {
        Object o = new Object();
        if (o instanceof Bar bar) {
            System.out.println(bar);
            if (o instanceof FooBar bar) {
                System.out.println(bar);
            }
        }
    }
}
```

## What is the full stack trace of any errors you encountered?

The migration succeeds, but if you try to compile the changes afterwards javac fails with
```
variable bar is already defined in method main(java.lang.String[])
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

If I manage to do so with reasonable effort, yes.

[...]

Issue #334, opened 2023-02-06
`InstanceOfPatternMatch` declares variable with conflicting name


knutwannheden:
The `InstanceOfPatternMatch` declares a pattern variable with a conflicting name if a variable with that name already exists in the `if` body after the corresponding type cast as in the following example:

```java
void test(Object o) {
  if (o instanceof String) {
    System.out.println((String) o);
    String string = "x";
  }
}
```

Currently the recipe will change this to the following illegal code:
```java
void test(Object o) {
  if (o instanceof String string) {
    System.out.println(string);
    String string = "x";
  }
}
```

Fixing openrewrite/rewrite#2776 should address this issue and the corresponding test case can be enabled again.
- openrewrite/rewrite#2776

timtebeek:
@koppor  The second case should see some improvement already after
- https://github.com/openrewrite/rewrite-static-analysis/pull/175

uhafner:
I have a similar (but a little bit different) example in one of my classes that cause the same problem:

```
private class JSONObject {
  Object getObject() {
      return new Object();
  }
}
public class A {
  void test(Object o) {
      if (o instanceof JSONObject) {
          Object jsonObject = ((JSONObject) o).getObject();
          if (jsonObject instanceof JSONObject) {
              Object result = ((JSONObject) jsonObject).getObject();
          }
      }
  }
}
```

This produces:

```
         if (o instanceof JSONObject object) {
              Object jsonObject = object.getObject();
              if (jsonObject instanceof JSONObject object) {
                  Object result = object.getObject();
              }
          }
```

PR #538, opened 2025-04-30
Fix clashing variable names generated by `InstanceOfPatternMatch`

arodionov:
In addition to the `cursor` scope, added the enclosing `J.If` scope for searching conflicting variable names.

- Fix https://github.com/openrewrite/rewrite-static-analysis/issues/334
- Fix https://github.com/openrewrite/rewrite-static-analysis/issues/481
- Fix https://github.com/openrewrite/rewrite/issues/2787
- It also can be a workaround for https://github.com/openrewrite/rewrite/issues/2776


timtebeek (approved):
Great to see these cleared out; I _think_ there might still be an issue with a triple nested `if`, but since this is already an improvement for a common case I propose we merge as is, and can address any other cases if reported.

PR #541, opened 2025-05-04
Enable custom Serial Version UID

yuri1969:
## What's changed?

To be able to choose from `1`, `1L`, or any other value.

## What's your motivation?

Codebases often go for the `1L` value, so having an option to use that value is cool.

Unit test coverage, maybe?

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/AddSerialVersionUidToSerializableTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/AddSerialVersionUidToSerializableTest.java
@@ -57,7 +57,6 @@ class AddSerialVersionUidToSerializableTest implements RewriteTest {
         );
     }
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/staticanalysis/AddSerialVersionUidToSerializableTest.java
```diff
@@ -57,6 +57,34 @@ public class Example implements Serializable {
         );
     }
 
+    @DocumentExample
```
timtebeek, line 60: We try to only have a single example for our docs, as these would be very similar.
```suggestion
```

timtebeek (approved):
Thanks for the addition here! Should help folks standardize if they had a certain preference.

PR #542, opened 2025-05-05
CustomImportOrder recipe

mankoffs:
## What's changed?
Added CustomImportOrder recipe

Dependant pull request  - https://github.com/openrewrite/rewrite/pull/5381

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/CustomImportOrder.java
+++ b/src/main/java/org/openrewrite/staticanalysis/CustomImportOrder.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


MBoegers:
I also added a small test to check if we are idempotent ;) 
Works locally, we start by merging https://github.com/openrewrite/rewrite/pull/5381 as soon as CI is green

MBoegers (approved):

mankoffs:
> I also added a small test to check if we are idempotent ;) Works locally, we start by merging [openrewrite/rewrite#5381](https://github.com/openrewrite/rewrite/pull/5381) as soon as CI is green

@MBoegers I also have a question, should I add this recipe here https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/resources/META-INF/rewrite/static-analysis.yml#L22 and here https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/resources/META-INF/rewrite/common-static-analysis.yml#L22?

MBoegers:
I think both are good matches ;)

PR #544, opened 2025-05-06
Reordering Annotations

### LINKED ISSUE #543: Java recipe to reorder annotations

I'd like to be able to tackle inconsistent order of frequently used annotations in the code.

e.g.
```
    @Issue("https://github.com/openrewrite/rewrite/issues/2973")
    @Test
    @ExpectedToFail
    void explicitImplementationClassInApi() {
```
vs.
```
        @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/112")
        @ExpectedToFail("only directly returned ternaries are taken into account")
        @Test
        void doReplaceNestedOrAssignmentTernaryWithIfElse() {
```
in our own code, etc. There are other combinations.

## Describe the solution you'd like

Maybe a recipe to instil some deterministic order (alphabetic) or maybe some smart way of passing priorities/orders - e.g. `@Test` should always come first, etc.

steve-aom-elliott:
- Fixes #543 

Reordering annotations (alphabetically only for now on method declarations alone).

Looking to define boundaries for what should be included in the changes for annotation reordering. As per the linked issue, the starting point would be a default alphabetical ordering of annotations (the ability to pass in what `Comparator<J.Annotation>` to use will be upcoming). It's been suggested to start with just method declarations and move on from there after, but I'm open to thoughts and suggestions.

## What's changed?
- Added a `ReorderAnnotations` recipe for reordering annotations (default alphabetic) on `J.MethodDeclaration`
- Added simple tests for the reordering for method declarations (only using the default sorting)

## Related issue with additional fixes that should be implemented first
- https://github.com/openrewrite/rewrite/issues/5392

@timtebeek 
@greg-at-moderne 

### Checklist
- [X] I've added unit tests to cover both positive and negative cases for method declarations
- [x] I've implemented handling of class declarations
- [x] I've implemented handling of field declarations
- [ ] I've added unit tests to cover both positive and negative cases for class declarations
- [ ] I've added unit tests to cover both positive and negative cases for field declarations
- [ ] I've implemented the ability to provide what `Comparator<J.Annotation>` you would like to use to sort the annotations on a given LST entry point

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/AddSerialAnnotationToSerialVersionUID.java
+++ b/src/main/java/org/openrewrite/staticanalysis/AddSerialAnnotationToSerialVersionUID.java
@@ -50,8 +50,8 @@ public class AddSerialAnnotationToSerialVersionUID extends Recipe {
         return Duration.ofMinutes(1);
     }
 
-    @Override
     @NonNull
+    @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return Preconditions.check(
                 Preconditions.and(
--- a/src/main/java/org/openrewrite/staticanalysis/BooleanChecksNotInverted.java
+++ b/src/main/java/org/openrewrite/staticanalysis/BooleanChecksNotInverted.java
@@ -51,8 +51,8 @@ public class BooleanChecksNotInverted extends Recipe {
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return new JavaVisitor<ExecutionContext>() {
 
-            @SuppressWarnings("ConstantConditions")
             @Override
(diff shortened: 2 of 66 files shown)


timtebeek:
@Laurens-W no rush on this at all ofcourse, but it's odd to see that we introduce a newline for the simplest case, and there's weird indentation and shuffling of comments for the more complex case. The first I'd at least expect to be fixed; the second is optional.

Laurens-W (approved):
LGTM

timtebeek:
There's one change here that I'm not quite happy with, and was helpfully pointed out here:
- https://github.com/openrewrite/rewrite-static-analysis/issues/543#issuecomment-2855819351
image

PR #546, opened 2025-05-07
SimplifyConstantIfBranchExecution leaves unreachable code

### LINKED ISSUE #399: SimplifyConstantIfBranchExecution leaves unreachable code

I am using

- org.openrewrite.recipe:rewrite-static-analysis:1.20.0
-  org.openrewrite.recipe:rewrite-recipe-bom:2.22.0

Java test case

```java

import org.openrewrite.staticanalysis.SimplifyConstantIfBranchExecution;
import org.openrewrite.test.RewriteTest;

class SimplifyTest implements RewriteTest {

    @Test
    void test() {
        rewriteRun(
                spec -> spec.recipe(new SimplifyConstantIfBranchExecution()),
                //language=java
                java(
                        """
                            public class A {
                                public boolean test() {
                                    if (true) {
                                        Object x;
                                        return true;
                                    }
                                    return false;
                                }
                            }
                        """,
                        """
                             public class A {
                                 public boolean test() {
                                     Object x;
                                     return true;
                                 }
                             }
                         """
                )
        );
    }
}
```

```java
public class A {
    public boolean test() {
        if (true) {
            Object x;
            return true;
        }
        return false;
    }
  }
```

```java
public class A {
     public boolean test() {
         Object x;
         return true;
     }
 }
```

## What did you see instead?

```java
public class A {
     public boolean test() {
         {
            Object x;
            return true;
         }
         return false;
     }
 }
```

No error, just invalid output.

Unable to contribute.

steve-aom-elliott:
- fixes #399 

Certain cases where there is an object assignment in an unneeded `if` block in the initial code can cause `SimplifyConstantIfBranchExecution` to leave scoping braces and unreachable lines in the code. The underlying cause is due to `RemoveUnneededBlock` deciding to not remove scoping braces if a block contains variable declarations, however, if the block also contains a return, then `SimplifyConstantIfBranchExecution` which also uses `RemoveUnreachableCodeVisitor` should be able to be unwrapped because there should be no conflicts after unreachable code is removed.

## What's changed?
- Test example from submitted issue has been added.
- Deeper nested if examples included for `SimplifyConstantIfBranchExecution` tests and test example provided for `RemoveUnneededBlock` specifically covering change in logic to allow unwrapping if return statement is nested.

### Checklist
- [X] I've added further unit tests to cover both positive and negative cases

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveUnneededBlock.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveUnneededBlock.java
@@ -80,8 +80,10 @@ public class RemoveUnneededBlock extends Recipe {
                     return stmt;
                 }
 
-                // blocks are relevant for scoping, so don't flatten them if they contain variable declarations
-                if (i < statements.size() - 1 && nested.getStatements().stream().anyMatch(J.VariableDeclarations.class::isInstance)) {
+                // blocks are relevant for scoping, so don't flatten them if they contain variable declarations unless they also have returns
+                if (i < statements.size() - 1 &&
+                    nested.getStatements().stream().anyMatch(J.VariableDeclarations.class::isInstance) &&
+                    nested.getStatements().stream().noneMatch(J.Return.class::isInstance)) {
                     return stmt;
                 }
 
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Code changes look good already; let me know if you saw any results on a large run that gave you pause before a merge, or if we can already merge as is.

PR #547, opened 2025-05-08
Handle static method calls when removing unnecessary explicit type arguments.

JohannisK:
## What's changed?
Adde rules to the removing unnecessary explicit type arguments recipe to deal with static method calls.

## What's your motivation?
The recipe currently removes explicit type arguments from method calls that need it.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
@@ -20,9 +20,8 @@ import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.staticanalysis.java.JavaFileChecker;
 
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 public class UnnecessaryExplicitTypeArguments extends Recipe {
 
@@ -49,18 +48,16 @@ public class UnnecessaryExplicitTypeArguments extends Recipe {
 
                 if (m.getMethodType() != null) {
                     Object enclosing = getCursor().getParentTreeCursor().getValue();
-                    JavaType enclosingType = null;
+                    JavaType inferedType = null;
                     if (enclosing instanceof J.MethodInvocation) {
                         if (m.getMethodType().getFlags().contains(Flag.Static)) {
-                            List<String> collect = m.getMethodType().getParameterTypes().stream().flatMap(p -> {
-                                if (!(p instanceof JavaType.Parameterized)) {
-                                    return Collections.<String>emptyList().stream();
-                                }
-                                return ((JavaType.Parameterized) p).getTypeParameters().stream()
-                                        .filter(t -> t instanceof JavaType.GenericTypeVariable)
-                                        .map(e -> ((JavaType.GenericTypeVariable)e).getName());
-                            }).collect(Collectors.toList());
-                            if(!collect.containsAll(m.getMethodType().getDeclaredFormalTypeNames())) {
+                            List<String> formalTypeNames = new ArrayList<>(m.getMethodType().getDeclaredFormalTypeNames());
+                            m.getMethodType().getParameterTypes().stream()
+                                    .filter(p -> p instanceof JavaType.Parameterized)
+                                    .flatMap(p -> ((JavaType.Parameterized) p).getTypeParameters().stream())
+                                    .filter(t -> t instanceof JavaType.GenericTypeVariable)
+                                    .forEach(it -> formalTypeNames.remove(((JavaType.GenericTypeVariable) it).getName()));
+                            if (!formalTypeNames.isEmpty()) {
                                 return m;
                             }
                         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
```diff
@@ -64,7 +80,7 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
                                 .count() > 1) {
                             return m;
                         }
-                        enclosingType = enclosingMethod.getType();
+                        enclosingType = m.getMethodType().getReturnType();
```
JohannisK, line 83: This change might be to simplistic, but when the `enclosing` is a method invocation:
We're dealing with `GenericClass.<String>someTypedBuilder(String.class)`, and enclosing is `GenericClass.<String>someTypedBuilder(String.class).build()`

<String> is already kept and returned earlier if the argument of the method does not do type inference (e.g. it does not take `Class<T> arg`. If it does do type inference, the method is its own enclosing type.
E.G. 
When `GenericClass.<String>someTypedBuilder(String.class)` returns a type inferred instance `GenericClassBuild<String>`, 
`GenericClass.<String>someTypedBuilder(String.class).build()` should be the same as
```
GenericClassBuild<String> gcb = GenericClass.someTypedBuilder(String.class);
gab.build();
```

JohannisK: Assigning the returnType to `enclosingType` is a bit of an ugly way of forcing the existing code to remove explicit generic type information.
Maybe I should rename `enclosingType` to `inferedType`?
```
if (enclosingType != null && TypeUtils.isOfType(enclosingType, m.getMethodType().getReturnType())) {
    m = m.withTypeParameters(null);
}
```
if the return type of the method invocation is the same as the inferred type, the explicit type arguments may be removed?

src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
```diff
@@ -46,8 +50,20 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
                 if (m.getMethodType() != null) {
                     Object enclosing = getCursor().getParentTreeCursor().getValue();
                     JavaType enclosingType = null;
-
                     if (enclosing instanceof J.MethodInvocation) {
+                        if (m.getMethodType().getFlags().contains(Flag.Static)) {
+                            List<String> collect = m.getMethodType().getParameterTypes().stream().flatMap(p -> {
+                                if (!(p instanceof JavaType.Parameterized)) {
+                                    return Collections.<String>emptyList().stream();
```
jevanlingen, line 57:
```suggestion
                                    return Stream.empty();
```

src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
```diff
@@ -46,8 +50,20 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                                }
+                                return ((JavaType.Parameterized) p).getTypeParameters().stream()
+                                        .filter(t -> t instanceof JavaType.GenericTypeVariable)
+                                        .map(e -> ((JavaType.GenericTypeVariable)e).getName());
+                            }).collect(Collectors.toList());
+                            if(!collect.containsAll(m.getMethodType().getDeclaredFormalTypeNames())) {
+                                return m;
+                            }
```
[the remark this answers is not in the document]
jevanlingen: You could also go one step further, but dunno if that's really easier to read. Just check for yourself:

```java
.flatMap(p -> p instanceof JavaType.Parameterized ?
        ((JavaType.Parameterized) p).getTypeParameters().stream()
                .filter(t -> t instanceof JavaType.GenericTypeVariable)
                .map(e -> ((JavaType.GenericTypeVariable) e).getName()) :
        Stream.empty())
```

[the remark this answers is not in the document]
jevanlingen: Or even:
```java
.flatMap(p -> ((JavaType.Parameterized) p).getTypeParameters().stream()
.filter(t -> t instanceof JavaType.GenericTypeVariable)
.map(e -> ((JavaType.GenericTypeVariable) e).getName()))
```

src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
jevanlingen: > Maybe I should rename enclosingType to inferedType?
Yeah do that, that's way easier to understand if you read this recipe for the first time.

src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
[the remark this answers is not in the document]
JohannisK: much much better!

timtebeek (approved):
Thanks for the fix here! Great to not make incorrect changes where we can avoid it.

PR #548, opened 2025-05-08
Configurable NoFinalizedLocalVariables for excluding methods parameters

### LINKED ISSUE #514: Feature request: configurable NoFinalizedLocalVariables
It would be great to allow configuring `org.openrewrite.java.cleanup.NoFinalizedLocalVariables` to remove `final` only from locally-declared variables, so it ignores method parameters.

I want to prevent writing to variables that are passed in method arguments, and IDEA can generate the `final` there automatically, but it's a lot of noise for local variables that I would like to get rid of.

steve-aom-elliott:
- Fixes #514 

## What's changed?
- An option has been added to `NoFinalizedLocalVariables` to allow `final` method/lambda parameters to be excluded from having their `final` designation removed.
- Added ability for `final` fields declared _within_ lambdas to have their final modifier removed in the same way they get cleaned up for regular method declarations.

### Checklist
- [ ] Rebase off of main once #546 gets merged.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/NoFinalizedLocalVariables.java
+++ b/src/main/java/org/openrewrite/staticanalysis/NoFinalizedLocalVariables.java
@@ -15,6 +15,9 @@
  */
 package org.openrewrite.staticanalysis;
 
+import lombok.EqualsAndHashCode;
+import lombok.Value;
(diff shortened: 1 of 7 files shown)


timtebeek (approved):
Thanks again! Nice to see a steady stream of improvements.

PR #553, opened 2025-05-09
UnnecessaryThrows should not change API for protected methods

timtebeek:
- For https://github.com/apache/maven/pull/2291
- Needs https://github.com/openrewrite/rewrite/pull/5412

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UnnecessaryThrows.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UnnecessaryThrows.java
@@ -138,79 +138,82 @@ public class UnnecessaryThrows extends Recipe {
 
                 return m;
             }
+
+            private Set<JavaType.FullyQualified> findExceptionCandidates(J.@Nullable MethodDeclaration method) {
+
+                if (method == null || method.getMethodType() == null || method.isAbstract() || method.isConstructor()) {
+                    return Collections.emptySet();
+                }
+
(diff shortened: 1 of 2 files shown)


Jenson3210:
What should we do when we the api offers 2 levels of inheritance? 

```java
class PlainText {
    // someone allows for exceptions 
    protected void check() throws CheckedException {
        basicChecks()
    }
}
```

```java
class ConfidentialText {
    // someone allows for exceptions 
    @Override
    protected void check() throws CheckedException {
        advancedChecks()
    }
}
```

```java
class TopSecret {
    // someone allows for exceptions 
    @Override
    protected void check() throws CheckedException {
       doesNotThrow()
    }
}
```

Will this not clean up the throws clause of `ConfidentialText` and `TopSecret` whereas it might have been needed by `TopSecret`?

timtebeek:
You're right, thanks! I thought the third level would get the exceptions declared on the top level still, but turns out that's not the case. Adjusted the logic to now look for final on the method or the class on protected methods before we change the signature.

Jenson3210 (approved):
LGTM!
Potentially add a C3 which does not get changed in the tests? Non final class and non final method. Although that is also covered by your original test case, it might be very clear how it behaves and what the difference is between these 3.

timtebeek:
Thanks for the help improving the approach! On mobile here from the plane, so I'll merge as is 😅

PR #565, opened 2025-05-22
No MethodReference for parentheses

timtebeek:
### Checklist


MBoegers (commented):
@rickie thanks for bringing it up

PR #566, opened 2025-05-23
Fallthrough should not add `break;` when statements contain a guaranteed return

JohannisK:
## What's changed?
- Added test
- Added check to find guaranteed returns

## What's your motivation?
When there is a guaranteed return in a switch case, adding a `break;` will result in unreachable code.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/FallThroughVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/FallThroughVisitor.java
@@ -291,7 +291,5 @@ public class FallThroughVisitor<P> extends JavaIsoVisitor<P> {
                 return case_;
             }
         }
-
     }
-
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/FallThroughVisitor.java
```diff
@@ -219,4 +223,137 @@ public J.Case visitCase(J.Case case_, Set<J> ctx) {
...
+        }
+
+        private static class FindGuaranteedReturnsVisitor extends JavaIsoVisitor<Set<J>> {
+            private Cursor cursor;
+            private final J.Case scope;
+
+            public FindGuaranteedReturnsVisitor(Cursor cursor, J.Case scope) {
+                this.cursor = cursor;
+                this.scope = scope;
+            }
```
jevanlingen, line 251:
```suggestion
        @RequiredArgsConstructor
        private static class FindGuaranteedReturnsVisitor extends JavaIsoVisitor<Set<J>> {
            private final Cursor cursor;
            private final J.Case scope;
```

PR #568, opened 2025-05-26
`MinimumSwitchCases` ignores switches with cases with more than 1 break.

JohannisK:
## What's changed?
`MinimumSwitchCases` ignores switches with cases with more than 1 break.

## What's your motivation?
When a case contains several breaks rewriting it to an if-statement results in an if-statement with a break.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
@@ -18,12 +18,11 @@ package org.openrewrite.staticanalysis;
 import lombok.AllArgsConstructor;
 import lombok.Value;
 import lombok.With;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.Preconditions;
-import org.openrewrite.Recipe;
-import org.openrewrite.TreeVisitor;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.RecipeRunException;
+import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
```diff
@@ -286,4 +290,15 @@ public DefaultOnly() {
             id = randomId();
         }
     }
+
+    private static class BreakFinderVisitor extends JavaIsoVisitor<List<Statement>> {
+
+        @Override
+        public @Nullable J visit(@Nullable Tree tree, List<Statement> statements) {
+            if(tree instanceof J.Break) {
```
Jenson3210, line 298: Not the best coding practice -> not sure if we need to detect, but will breaks inside nested switch statements be counted also? 
Perhaps we can not traverse deeper inside other switch cases to avoid this if we would want to support?

```java
switch(a) {
    case "a" -> {
        switch (b) {
            case 0 -> break;
            default -> break;
        }
        break;
    default:
        break;
```

jevanlingen: I wonder if this is really a problem, but @JohannisK just start with writing more unit tests. I definitely think we should at least check what happens when nested switch statements are used.

src/test/java/org/openrewrite/staticanalysis/MinimumSwitchCasesTest.java
```diff
@@ -774,4 +774,75 @@ void doSomethingElse() {}
...
+                  int variableA;
+                  int variableB;
+                  void test() {
+                      Object returnValue;
+                      if (variableA == 0) {
+                          if (variableB == 0) {
+                          } else {
+                          }
+                      } else {
+                      }
```
Jenson3210, line 841: should empty else blocks be removed by calling some other recipe/in this recipe?

PR #569, opened 2025-05-26
DEV Add nullable annotation for methods with ternary null returns

stefanodallapalma:
## What's changed?

This PR updates the `AnnotateNullableMethods` recipe in order to annotate methods with ternary null returns.

#### Example

**Before:**

```java
import java.util.Random;

public class Test {

    public String getString() {
        //...
        return new Random().nextBoolean() ? "Not null" : null;
    }
}
```

**After:**

```java
import java.util.Random;

public class Test {

    public @Nullable String getString() {
        //...
        return new Random().nextBoolean() ? "Not null" : null;
    }
}
```

## What's your motivation?
Support one more use case.

### Checklist
- [] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java
@@ -283,4 +283,56 @@ class AnnotateNullableMethodsTest implements RewriteTest {
           )
         );
     }
+
+    @DocumentExample
+    @Test
+    void methodReturnsNullInTernary() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import java.util.Random;
+
+              public class Test {
+
+                  public String getString() {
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java
```diff
@@ -283,4 +283,56 @@ static class Inner {}
           )
         );
     }
+
+    @DocumentExample
+    @Test
+    void methodReturnsNullInTernary() {
```
greg-at-moderne, line 289:
```suggestion
    @Test
    void methodReturnsNullInTernary() {
```

timtebeek:
Awesome work here @stefanodallapalma ! Spotted this just now and immediately ran it to improve our API documentation:
- https://github.com/openrewrite/rewrite/pull/5519

PR #575, opened 2025-05-28
fix: remove unnecessary continue in empty while loops

### LINKED ISSUE #570: EmptyBlock adds unnecessary `continue` in empty while loops

8.53.1

Moderne

A few examples of OSS repro:
- https://github.com/apache/orc/blob/c73651d22c51cb69ef82992e0bdca61ec932cb71/java/core/src/test/org/apache/orc/impl/TestZlib.java#L78
- https://github.com/apache/james-mime4j/blob/0ee7cc830ed449d8c44c0e5e2398a79fa2f5ff89/core/src/main/java/org/apache/james/mime4j/stream/MimeEntity.java#L409
- https://github.com/spring-projects/spring-batch/blob/08c4cb16b854b773f974eeb2073a04c56a0eb6ab/spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/ChunkMessageItemWriterIntegrationTests.java#L115

The recipe adds a `continue` statement to an empty `while` loop body. It's counterproductive. If anything, it should remove `continue` if it was the only statement.

juancaCeb:
## What's changed?
Changed EmptyBlockVisitor "visitWhileLoop" and "visitDoWhileLoop" methods so that they do not add any "continue" statement. Instead, If there is a continue statement, it should remove them.

## What's your motivation?
Closes #570 

I would like to check if it's okay to put the logic of removing the continue statement on the same class. Or if not, where should it be?

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/EmptyBlockVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/EmptyBlockVisitor.java
@@ -48,9 +48,12 @@ public class EmptyBlockVisitor<P> extends JavaIsoVisitor<P> {
     public J.WhileLoop visitWhileLoop(J.WhileLoop whileLoop, P p) {
         J.WhileLoop w = super.visitWhileLoop(whileLoop, p);
 
-        if (Boolean.TRUE.equals(emptyBlockStyle.getLiteralWhile()) && isEmptyBlock(w.getBody())) {
+        if (Boolean.TRUE.equals(emptyBlockStyle.getLiteralWhile()) && w.getBody() instanceof J.Block) {
             J.Block body = (J.Block) w.getBody();
-            w = continueStatement.apply(updateCursor(w), body.getCoordinates().lastStatement());
+            List<Statement> statements = body.getStatements();
+            if (statements.size() == 1 && statements.get(0) instanceof J.Continue) {
+                w = w.withBody(body.withStatements(new ArrayList<>()));
+            }
         }
 
         return w;
@@ -60,9 +63,12 @@ public class EmptyBlockVisitor<P> extends JavaIsoVisitor<P> {
     public J.DoWhileLoop visitDoWhileLoop(J.DoWhileLoop doWhileLoop, P p) {
         J.DoWhileLoop w = super.visitDoWhileLoop(doWhileLoop, p);
 
-        if (Boolean.TRUE.equals(emptyBlockStyle.getLiteralWhile()) && isEmptyBlock(w.getBody())) {
+        if (Boolean.TRUE.equals(emptyBlockStyle.getLiteralWhile()) && w.getBody() instanceof J.Block) {
             J.Block body = (J.Block) w.getBody();
-            w = continueStatement.apply(updateCursor(w), body.getCoordinates().lastStatement());
+            List<Statement> statements = body.getStatements();
+            if (statements.size() == 1 && statements.get(0) instanceof J.Continue) {
+                w = w.withBody(body.withStatements(new ArrayList<>()));
+            }
         }
 
         return w;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/EmptyBlockVisitor.java
```diff
@@ -48,9 +48,12 @@ public class EmptyBlockVisitor<P> extends JavaIsoVisitor<P> {
...
         J.WhileLoop w = super.visitWhileLoop(whileLoop, p);
 
-        if (Boolean.TRUE.equals(emptyBlockStyle.getLiteralWhile()) && isEmptyBlock(w.getBody())) {
+        if (Boolean.TRUE.equals(emptyBlockStyle.getLiteralWhile()) && w.getBody() instanceof J.Block) {
             J.Block body = (J.Block) w.getBody();
-            w = continueStatement.apply(updateCursor(w), body.getCoordinates().lastStatement());
+            List<Statement> statements = body.getStatements();
+            if (statements.size() == 1 && statements.get(0) instanceof J.Continue) {
+                w = w.withBody(body.withStatements(new ArrayList<>()));
+            }
```
greg-at-moderne, line 56: When reading the code, it seems like it handles the case of removing an existing `continue` statement in while loop if it's the only statement. But does the code handle empty loops after the change? I suggest you add an explicit test case for this.

greg-at-moderne: In other words, there are at least two scenarios on **input**:
- a while-loop which is empty
- a while-loop which just has a continue statement and nothing else.

I think before the changes the recipe handled the first case. Your code seems to change it to handling the second case. I suggest to keep the support for case no. 1, and if you prefer we can add support for case no. 2 on top of that.

juancaCeb: Thanks for the clarifications! I do have one question though:
Originally, the behavior only added a continue statement to empty while loops. That's why I did the opposite which is to remove the continue statement. However, keeping both behaviors would introduce a conflict between them. Was the original behavior intended to do something else?

greg-at-moderne: OK. Sorry, my input might have been confusing. I've just taken a broader look and the things are more complex than I thought. Still I think there are things to fix and your contribution is welcome.

- When looking at `EmptyBlock`, its description and name, it clearly indicates empty blocks should be removed. They are not. Instead some `continue` statements are added.
- At the same time, there's a thing called `EmptyBlockStyle` which is a very detailed way of modelling what should happen with empty blocks.
- As a consequence I think we need multiple variants of the `EmptyBlock` logic. As we need to cater for multiple behaviors.
- `BlockPolicy` seems to have two values: TEXT, STATEMENT. None of them calls for removing the blocks.

My personal preference would be for the `EmptyBlock` recipe to remove blocks by default. And optionally do something else if instructed.

Thus I suggest:
- keep the `BlockPolicy` and `EmptyBlockStyle` as they are,
- change the default behavior of `EmptyBlock` - the default should remove the empty blocks
- if a valid `EmptyBlockStyle` is passed, it should be honored accordingly

@timtebeek Can you please review the suggestion?

timtebeek: I do not think we can safely remove empty while blocks; I'm somewhat surprised a recipe aimed at removing empty blocks is even touching these at all, or changing their body, but the above strikes the right balance I think.

The reason I think we can not remove empty while blocks is things like the following, which are valid & should not be removed
```java
while (input.read() != null) {} // fully consume input
```

timtebeek (commented):
Wondering a bit about if, and which changes we should be making here; as pointed out in
- https://github.com/openrewrite/rewrite-static-analysis/pull/575/files#r2132004866

Issue #572, opened 2025-05-27
InstanceofPatternMatching recipe prefers type of cast instead of instanceof check


dnl50:
## What version of OpenRewrite are you using?

Tested in current main branch commit (0a00b1d1) using a unit test.

## What is the smallest, simplest way to reproduce the problem?

When the type of the `instanceof` check and the casting afterwards differs, the type of the cast is preferred which changes the semantic of the code:

```java
public class Main {
    void test(Object o) {
        if (o instanceof Car) {
            ((Vehicle) o).start();
        }
    }

    private static abstract class Vehicle {
        abstract void start();
    }

    private static class Car extends Vehicle {
        void start() {}
    }
    
    private static class Bus extends Vehicle {
        void start() {}
    }
}
```

## What did you expect to see?

```java
public class Main {
    void test(Object o) {
        if (o instanceof Car car) {
            car.start();
        }
    }

    private static abstract class Vehicle {
        abstract void start();
   }

   private static class Car extends Vehicle {
       void start() {}
   }
}
```

## What did you see instead?

```java
public class Main {
    void test(Object o) {
        if (o instanceof Vehicle car) {
            car.start();
        }
    }

    private static abstract class Vehicle {
        abstract void start();
    }

    private static class Car extends Vehicle {
        void start() {}
    }
}
```

The migrated code matches all `Vehicle` types, whereas the original code only matched `Car`s

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
I might take a look into it

PR #577, opened 2025-05-29
Do not pull Supertype-Cast to instanceof

dnl50:
When the cast type expression is a supertype of the instanceof type, the recipe should back off. The code semantics could change otherwise, because more instances can potentially pass the check. Fixes GH-572.

## What's changed?

Type casts after `instanceof` checks which use a different type than the one used in the `instanceof` check itself will now be ignored since changing the `instanceof` type might alter the code semantics.

## What's your motivation?

Issue #572

The `isCheckedCastCompatible` might be extended even further to take generic type variables into account. Cases like these could also be migrated, since the type variables are the same:

```java
// java 11
class A<T> {
   void test(List<T> o) {
       if (o instanceof ArrayList<?>) {
           ((ArrayList<T>) o).trimToSize();
       }
   }
}

// java 16+ where the reifiable type-restriction was lifted
class A<T> {
   void test(List<T> o) {
       if (o instanceof ArrayList<T> list) {
           list.trimToSize();
       }
   }
}
```

The `isCheckedCastCompatible` might be a duplicate of existing code, but I didn't find any publicly accessible methods which perform such checks ☹️ 

In the first try I just flipped the type `TypeUtils#isAssignableTo` parameters in the `registerTypeCast` method, but then subtype casts were pulled up to the `instanceof` check, resulting in the same problem of changed semantics.

### Checklist
- [✅] I've added unit tests to cover both positive and negative cases
- [✅] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [❌] I've used the IntelliJ IDEA auto-formatter on affected files, since the formatting differs a bit from the existing code when using the IntelliJ formatter

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
@@ -188,10 +188,11 @@ public class InstanceOfPatternMatch extends Recipe {
         @SuppressWarnings("SuspiciousMethodCalls")
         public void registerTypeCast(J.TypeCast typeCast, Cursor cursor) {
             Expression expression = typeCast.getExpression();
-            JavaType type = typeCast.getClazz().getTree().getType();
+            JavaType castType = typeCast.getClazz().getTree().getType();
 
             Optional<ExpressionAndType> match = instanceOfs.keySet().stream()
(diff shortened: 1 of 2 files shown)


sambsnyd:
Thanks @dnl50 !

PR #579, opened 2025-05-31
Handle nested types in AnnotateNullableMethods

timtebeek:
- As discovered on #578

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/staticanalysis/AnnotateNullableMethodsTest.java
```diff
@@ -334,4 +335,42 @@ public String getString() {
...
+                  }
+              }
+              """,
+            """
+              import a.B;
+              import org.jspecify.annotations.Nullable;
+
+              public class Foo {
+
+                  public  B.@Nullable C bar() {
```
timtebeek, line 368: Not quite happy with this extra newline & double space yet, which are part of moving the annotation after it was added.

timtebeek: Looks like this should be fixed upstream, so we'll leave this here for now.
https://github.com/openrewrite/rewrite-static-analysis/blob/a779afabaa0d12c77df16a384520e3a64bc762bd/src/main/java/org/openrewrite/staticanalysis/java/MoveFieldAnnotationToType.java#L80-L84

PR #594, opened 2025-06-16
Order type annotations last

timtebeek:
- Follow up from https://github.com/openrewrite/rewrite-static-analysis/pull/544#issuecomment-2976148831

> There's one change here that I'm not quite happy with, and was helpfully pointed out here:
- https://github.com/openrewrite/rewrite-static-analysis/issues/543#issuecomment-2855819351
image


steve-aom-elliott (approved):
Looks good to me

Issue #592, opened 2025-06-13
UseCollectionInterfaces might fail with UnsupportedOperationException


greg-at-moderne:
## What version of OpenRewrite are you using?

rewrite-static-analysis `2.11.0-20250612.172325-31` snapshot

## How are you running OpenRewrite?

Moderne.

## What is the smallest, simplest way to reproduce the problem?
```java
    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/592")
    @Test
    void returnHashSetSize() {
        rewriteRun(
          //language=java
          java(
            """
              import java.util.HashSet;
              import java.util.Set;

              class Test {
                  public int method(Set<Integer> input) {
                      return new HashSet<>(input).size();
                  }
              }
              """
          )
        );
    }
```
in `UseCollectionInterfacesTest`

## What did you expect to see?

Leaving that as an exercise to the reader ;)

## What did you see instead?
```
Caused by: java.lang.UnsupportedOperationException: To change the return type of this new class, use withConstructorType(..)
	at org.openrewrite.java.tree.J$NewClass.withType(J.java:4792)
	at org.openrewrite.java.tree.J$NewClass.withType(J.java:4687)
	at org.openrewrite.staticanalysis.UseCollectionInterfaces$1.updateMethodInvocation(UseCollectionInterfaces.java:228)
	at org.openrewrite.staticanalysis.UseCollectionInterfaces$1.visitMethodInvocation(UseCollectionInterfaces.java:167)
```

## OSS repro

There are plenty.
https://github.com/apache/commons-rdf/blob/3237655f9715a3e8b2c6cd39c5105ce9295587fb/commons-rdf-api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java#L284 to name just one of them.

PR #517, opened 2025-04-16
UseCollectionInterfaces updates methodInvovations type info to enable import removal

JohannisK:
## What's changed?
UseCollectionInterfaces now also visits methodInvocation to change their type. This causes the imports to be removed correctly.

## What's your motivation?
Fixes:
- https://github.com/openrewrite/rewrite-static-analysis/issues/357

PR #595, opened 2025-06-16
fix: filter out methods not invoked from an identifier

pdelagrave:
## What's changed?
`UseCollectionInterfaces.visitMethodInvocation()` will now only proceed if the method was invoked on an expression being an identifier. Anonymous objects don't have an identifier so it doesn't make sense to try to migrate their identifier from a concrete class to an interface.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/592

`UseCollectionInterfaces.visitMethodInvocation()` was crashing trying to update the identifier type of an identifier-less anonymous object method invocations.

@JohannisK who worked on adding the override `UseCollectionInterfaces.visitMethodInvocation()` 2 months ago.
- https://github.com/openrewrite/rewrite-static-analysis/pull/517

Yes, see next section

The (only) change at line 156 is:
```
if (mi.getSelect() != null && mi.getSelect().getType() != null && mi.getSelect() instanceof J.Identifier) {
```
could have been:
```
if (mi.getSelect() != null && mi.getSelect().getType() != null && !(mi.getSelect() instanceof J.NewClass)) {
```
The added test case based from the bug report did crash specifically because of the `visitMethodInvocation()` was processing the invocation even if it was for an invocation on an anonymous object which didn't make sense in that context. So skipping it by checking if the expression is exactly that could make sense, but that leave a lot of other expression type that could also not make sense in that context.

Specifying that we only want to process expressions that are an identifier narrows down possibilities a lot and might avoid other bugs.

But I'd like a 2nd opinion as even if so far it looks OK and all the tests are passing, this might narrow down when `visitMethodInvocation()` runs too much?

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UseCollectionInterfaces.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UseCollectionInterfaces.java
@@ -153,7 +153,7 @@ public class UseCollectionInterfaces extends Recipe {
             @Override
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
-                if (mi.getSelect() instanceof J.Identifier && mi.getSelect().getType() != null) {
+                if ((mi.getSelect() instanceof J.Identifier || mi.getSelect() instanceof J.FieldAccess) && mi.getSelect().getType() != null) {
                     JavaType originalType = mi.getSelect().getType();
                     JavaType.FullyQualified fullyQualified = TypeUtils.asFullyQualified(originalType);
                     if (fullyQualified != null) {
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks @pdelagrave ! I've made a small change to pull forward the `instanceof` to where we had a `null` check previously, as that's then implied by the `instanceof`. I've also added a new test case that we also didn't cover before, but would be affected by the approach you've taken here, so figured good to add into the mix here and see that fixed as well. Would you be ok to tag team on that?

pdelagrave:
> I've made a small change to pull forward the instanceof to where we had a null check previously, as that's then implied by the instanceof.
duh yeah, thanks

>  I've also added a new test case that we also didn't cover before, but would be affected by the approach you've taken here, so figured good to add into the mix here and see that fixed as well.

Thanks, good to know. So in that case we'd have to add
```java
if (mi.getSelect() instanceof J.Identifier || mi.getSelect() instanceof J.FieldAccess ...)
```
Back to my question in the initial comment, unsure how to fully filter in all the right `instanceof Expression` subclass we want. So we got two now (I wasn't expecting that explicit field access with `this.` would have its own `Expression` type), is the API also fine grained and has an `Expression` implementation for things like accessing a variable of an outer scope from within a lambda for example? I wouldn't know what other `Expression` types to look for inclusion.

Issue #574, opened 2025-05-28
Add Regressiontest to verify no default import order style is applied


JohannisK:
Add a test verifying the changes in https://github.com/openrewrite/rewrite-static-analysis/pull/573

The recipe used to pick a default import order style, in the PR a change was merged so no default import order style is picked.

This change still needs a unit test to verify.

LakshyaKwatra:
Hi @timtebeek I have added the unit test for the verification of the same as a part of this PR: https://github.com/openrewrite/rewrite-static-analysis/pull/597
Please do have a look. Thanks!

PR #597, opened 2025-06-18
Add regression test to verify that no default import order style is applied

LakshyaKwatra:
## What's changed?
Added regression test to verify that no default import order style is applied as a part of this issue:
- Closes https://github.com/openrewrite/rewrite-static-analysis/issues/574 

@timtebeek 

## Additional Context
With this PR, we test the case where if no style is applied, the import order should not change.

### Checklist


timtebeek (approved):
Thanks for the help here @LakshyaKwatra ! Good to have this guarded in the case of any future changes.

PR #598, opened 2025-06-19
Fix ClassCastException when handling nested exception class names in CombineSemanticallyEqualCatchBlocks

AbhishekHerbertSamuel:
## What does this PR do?

This PR updates the `CombineSemanticallyEqualCatchBlocks` recipe to support nested class exception types (e.g., `A.Exception`, `B.Exception`). Previously, such catches would not be considered semantically equal due to the way type information was being interpreted.

### ✅ Changes Made:
- Enhanced exception type resolution to support nested class names using `J.FieldAccess`.
- Updated internal logic to avoid false negatives when comparing types involving nested classes.
- Added a dedicated test case: `catchingNestedClassName()` to validate the transformation logic.

### 🧪 Testing:
Wrote a test case and ensured that it passed.

---


jevanlingen:
Very nice @AbhishekHerbertSamuel! I cleaned the PR a little bit to make it more conform how we write our recipe code. At the moment the `removeRedundantChildClassesWithExistingMultiCatches` test does fail though. I did not do a proper review yet, do you have time to dive into this one?

AbhishekHerbertSamuel:
Hi @jevanlingen, thank you :)... I'll handle it and send the updated code.

AbhishekHerbertSamuel:
Hi @jevanlingen,
I've addressed your feedback on this PR:

Reworked the logic in combineEquivalentCatches() to correctly filter out redundant subtypes in existing multi-catch blocks.

Fixed removeRedundantChildClassesWithExistingMultiCatches so it now passes as expected.

Ensured J.FieldAccess (nested exception types) are handled correctly throughout the deduplication and transformation process.

Let me know if anything else needs refining.

PR #601, opened 2025-06-21
(RSPEC-S2221) SpecifyGenericExceptionCatches recipe

stefanodallapalma:
## Overview
Replaces generic `catch(Exception)` blocks with specific exception types that are actually thrown within the try block. This improves exception handling precision and code maintainability by catching only the exceptions that can actually occur.

## What's your motivation?
Generic exception catching with `catch(Exception e)` is considered a code smell (SonarQube rule [S2221](https://sonarsource.atlassian.net/browse/RSPEC-2221), [CWE-396](https://cwe.mitre.org/data/definitions/396.html)) because it masks specific error conditions and makes debugging harder. When developers write code that throws specific exceptions like `IOException` or `SQLException`, the catch blocks should reflect that specificity rather than using the overly broad `Exception` type.

## What it detects

* Generic `catch(Exception e)` blocks in try-catch statements
* Analyzes method calls and constructor invocations within try blocks to identify thrown exception types
* Generates appropriate single-catch or multi-catch syntax for those thrown exception types if not already caught by more specific catch clauses

**Before**:

```java
try {
    Files.readAllLines(path);
    connection.createStatement().executeQuery(sql);
} catch (Exception e) {
    log.error("Operation failed", e);
}
```

**After:**

```java
try {
    Files.readAllLines(path);
    connection.createStatement().executeQuery(sql);
} catch (IOException | SQLException e) {
    log.error("Operation failed", e);
}
```

### Implementation approach
- **Exact matching**: The recipe specifically targets `java.lang.Exception` catches (not subclasses like `RuntimeException`) to be conservative about what constitutes "generic" exception handling
- **Multi-catch generation**: Uses JavaTemplate with complete try-catch syntax to properly parse multi-catch expressions with `|` separators
- Recipe name: feel free to suggest more suitable names

### Limitations
- **Static analysis only**: Relies on method/constructor signatures for thrown exceptions; doesn't perform data flow analysis for dynamically thrown exceptions
- **Checked exceptions focus**: Primarily improves handling of checked exceptions since those must be declared in method signatures

@timtebeek

### Checklist
- [x] I've read and applied the [[recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/SpecifyGenericExceptionCatches.java
+++ b/src/main/java/org/openrewrite/staticanalysis/SpecifyGenericExceptionCatches.java
@@ -1,0 +1,184 @@
+package org.openrewrite.staticanalysis;
+
+import org.openrewrite.Cursor;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.ShortenFullyQualifiedTypeReferences;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.JavaType.FullyQualified;
+import org.openrewrite.java.tree.TypeUtils;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+
+public class SpecifyGenericExceptionCatches extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Replace `catch(Exception)` with specific exceptions thrown in the try block";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replaces `catch(Exception e)` blocks with a multi-catch block " +
+                "(`catch (SpecificException1 | SpecificException2 e)`) containing only the checked exceptions " +
+                "explicitly thrown by method or constructor invocations within the `try` block that are not " +
+                "already caught by more specific `catch` clauses. If no checked exceptions are found that " +
+                "require catching, the generic `catch` block is removed.";
+    }
+
+    @Override
+    public Set<String> getTags() {
+        return Collections.unmodifiableSet(new HashSet<>(Arrays.asList("CWE-396", "RSPEC-S2221")));
+    }
+
+    @Override
+    public JavaVisitor<ExecutionContext> getVisitor() {
+        return new GenericExceptionCatchVisitor();
+    }
+
+    private static class GenericExceptionCatchVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String JAVA_LANG_EXCEPTION = "java.lang.Exception";
+        private static final String MULTI_CATCH_SEPARATOR = "|";
+        private static final String TRY_CATCH_TEMPLATE = "try {} catch (%s %s) {}";
+
+        @Override
+        public J.Try visitTry(J.Try aTry, ExecutionContext ctx) {
+            J.Try t = super.visitTry(aTry, ctx);
+
+            if (hasGenericCatch(t)) {
+                Set<JavaType> caughtExceptions = getCaughtExceptions(t);
+                Set<JavaType> thrownExceptions = getThrownExceptions(t);
+                thrownExceptions.removeAll(caughtExceptions); // Remove exceptions that are already specifically caught
+
+                if (!thrownExceptions.isEmpty()) {
+                    List<J.Try.Catch> updatedCatches = t.getCatches().stream()
+                            .map(c -> updateCatchIfGeneric(c, thrownExceptions))
+                            .collect(Collectors.toList());
+
+                    return t.withCatches(updatedCatches);
+                }
+            }
+
+            return t;
+        }
+
+        private static boolean isGenericCatch(J.Try.Catch aCatch) {
+            FullyQualified fq = TypeUtils.asFullyQualified(aCatch.getParameter().getType());
+            if (fq != null) {
+                String fqn = fq.getFullyQualifiedName();
+                return TypeUtils.fullyQualifiedNamesAreEqual(JAVA_LANG_EXCEPTION, fqn);
+            }
+            return false;
+        }
+
+        private boolean hasGenericCatch(J.Try aTry) {
+            for (J.Try.Catch c : aTry.getCatches()) {
+                if (isGenericCatch(c)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        private Set<JavaType> getCaughtExceptions(J.Try aTry) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/SpecifyGenericExceptionCatches.java
```diff
@@ -0,0 +1,199 @@
...
+        private static final String TRY_CATCH_TEMPLATE = "try {} catch (%s %s) {}";
+
+        @Override
+        public J.Try visitTry(J.Try aTry, ExecutionContext ctx) {
+            J.Try t = super.visitTry(aTry, ctx);
+
+            if (hasGenericCatch(t)) {
+                Set<JavaType> caughtExceptions = getCaughtExceptions(t);
+                Set<JavaType> thrownExceptions = getThrownExceptions(t);
+                thrownExceptions.removeAll(caughtExceptions); // Remove exceptions that are already specifically caught
```
stefanodallapalma, line 73: Forgot to add maybeAddImport for these exceptions

stefanodallapalma: Although I guess it's not really needed as `ShortenFullyQualifiedTypeReferences` does it already

greg-at-moderne:
Thanks for the great contribution.
I am looking into you test cases and am wondering what the behavior should be for:
```
                      try {
                          new MockThrowingClass().throwsBothIOAndSQL();
                      } catch (IOException | Exception e) {
                          System.out.println("Caught: " + e.getMessage());
                      }
```

stefanodallapalma:
Hey @greg-at-moderne 

That's an interesting case! My take is that Exception should ideally be replaced by SQLException for better specificity. However, directly using `catch (IOException | SQLException e)` isn't possible because IOException and SQLException are not disjoint types. The compiler will flag this as an error since IOException is a subclass of Exception.

stefanodallapalma:
Hi @timtebeek and @greg-at-moderne, since there hasn't been any recent activity on this PR, do you have a timeline for merging and releasing this recipe? While I can use it as a standalone recipe in our organization's project, it would be great to have it available out-of-the-box from rewrite-static-analysis.

timtebeek:
hi @stefanodallapalma ; thanks for the help! For a bit of context: Greg is out in Denmark enjoying his holiday, while I was out in Luxembourg enjoying mine, so it's taking us a little longer to get through everything. 😅 

I'll try to fit this in when I can, but hard to say exactly when that'll be. Know that the work is appreciated though! Reviews are just taking a little longer over the summer.

timtebeek:
> @timtebeek I see. I didn't mean to be pushy, just wanted to check in. Take your time! 😊

No worries at all; the context is just to let you know we strive to get things in quickly where we can, but life at times gets in the way. :)

src/main/java/org/openrewrite/staticanalysis/OnlyCatchDeclaredExceptions.java
```diff
@@ -0,0 +1,136 @@
...
+                    if (isGenericCatch(c)) {
+                        // Find declared thrown exceptions that are not already specifically caught
+                        Set<JavaType> declaredThrown = getDeclaredThrownExceptions(t);
+                        declaredThrown.removeAll(getCaughtExceptions(t));
+                        if (!declaredThrown.isEmpty()) {
+                            return multiCatchWithDeclaredExceptions(c, declaredThrown);
+                        }
+                    }
+                    return c;
+                }));
```
stefanodallapalma, line 73:
```suggestion
                // Find declared thrown exceptions that are not already specifically caught
                Set<JavaType> declaredThrown = getDeclaredThrownExceptions(t);
                declaredThrown.removeAll(getCaughtExceptions(t));
                return t.withCatches(ListUtils.map(t.getCatches(), c -> {
                    if (isGenericCatch(c) && !declaredThrown.isEmpty()) {
                            return multiCatchWithDeclaredExceptions(c, declaredThrown);
                    }
                    return c;
                }));
```

To avoid iterating over the try block multiple times?

timtebeek:
Seeing somewhat debatable results for let's say:
```diff
diff --git a/maven-scm-providers/maven-scm-providers-git/maven-scm-provider-jgit/src/main/java/org/apache/maven/scm/provider/git/jgit/command/checkout/JGitCheckOutCommand.java b/maven-scm-providers/maven-scm-providers-git/maven-scm-provider-jgit/src/main/java/org/apache/maven/scm/provider/git/jgit/command/checkout/JGitCheckOutCommand.java
index 2c745ff..8558d2b 100644
--- a/maven-scm-providers/maven-scm-providers-git/maven-scm-provider-jgit/src/main/java/org/apache/maven/scm/provider/git/jgit/command/checkout/JGitCheckOutCommand.java
+++ b/maven-scm-providers/maven-scm-providers-git/maven-scm-provider-jgit/src/main/java/org/apache/maven/scm/provider/git/jgit/command/checkout/JGitCheckOutCommand.java
@@ -19,6 +19,7 @@ org.openrewrite.staticanalysis.OnlyCatchDeclaredExceptions
 package org.apache.maven.scm.provider.git.jgit.command.checkout;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
@@ -205,7 +206,7 @@
             logger.debug("current branch: " + git.getRepository().getBranch());
 
             return new CheckOutScmResult("checkout via JGit", listedFiles);
-        } catch (Exception e) {
+        } catch (AmbiguousObjectException | CanceledException | CheckoutConflictException | CorruptObjectException | GitAPIException | IOException | IncorrectObjectTypeException | InvalidConfigurationException | InvalidRefNameException | InvalidRemoteException | MissingObjectException | NoHeadException | RefAlreadyExistsException | RefNotAdvertisedException | RefNotFoundException | RevisionSyntaxException | ScmException | TransportException | WrongRepositoryStateException e) {
             throw new ScmException("JGit checkout failure!", e);
         } finally {
             JGitUtils.closeRepo(git);
```

But given that we're not including this recipe by default anywhere (yet) I think that's fine for now; folks can run this when they want these results.

timtebeek (approved):
Thanks a lot for the initiative here @stefanodallapalma ! Hope you agree with the changes I've made to mostly keep the logic the same (not removing catch blocks) and opting to focus on declared exceptions.

stefanodallapalma:
> Thanks a lot for the initiative here @stefanodallapalma ! Hope you agree with the changes I've made to mostly keep the logic the same (not removing catch blocks) and opting to focus on declared exceptions.

Thanks for all the enhancements, Tim! I really appreciate them as they continue to teach me how to best use the OpenRewrite AST constructs.

PR #616, opened 2025-07-04
Place `Test` annotations first when reordering

timtebeek:
## What's your motivation?
We saw `@ParameterizedTest` moved down, which looks quite odd.

Are we ok to force this convention on others?

* Tolerate the change: apply as it was without this change
* Add an option with a sensible default that controls this behavior: we don't like a growing number of options here


src/test/java/org/openrewrite/staticanalysis/AbstractClassPublicConstructorTest.java
```diff
@@ -31,8 +31,8 @@ public void defaults(RecipeSpec spec) {
         spec.recipe(new AbstractClassPublicConstructor());
     }
 
-    @DocumentExample
     @Test
+    @DocumentExample
```
timtebeek, line 35: This is somewhat unfortunate, but a trade off compared to `@ParameterizedTest`.

Issue #627, opened 2025-07-08
Recipe to use varargs to Arrays.asList instead of new array construction


greg-at-moderne:
## What problem are you trying to solve?

Code readability in a very specific case of:
- calling `Arrays.asList` (or possibly similar collection constructions)
- with a newly array
- instead of using varargs

Behind the scenes the array is likely constructed anyway, but there's no need to make it explicit.

## Describe the situation before applying the recipe
```java
Arrays.asList(new String[]{"w", "k", "s"})
```

## Describe the situation after applying the recipe
```java
Arrays.asList("w", "k", "s")
```

## OSS repro
Examples:
- https://github.com/allaboutjst/airbnb/blob/40a6060d866fd0dd95a0138287b29b2770e1ab0d/src/main/java/travel_buddy/TravelBuddy.java#L83-L86
- https://github.com/EngineHub/CommandHelper/blob/c97a853c773199af2ae8a478b17c73694b6a0458/src/main/java/com/laytonsmith/tools/docgen/sitedeploy/SiteDeploy.java#L423-L425

PR #633, opened 2025-07-09
Simplify `Arrays.asList` with new array argument

timtebeek:
## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/627


timtebeek:
On Apache projects I saw `Arrays.asList(new Object[]{null})`, which needs special handling still.

Issue #628, opened 2025-07-08
Recipe to avoid concatenating with empty String to perform type conversion


greg-at-moderne:
## What problem are you trying to solve?

Code readability and possibly performance when using the `"" + ...` trick to convert something to a String.
`String.valueOf()` is a better alternative.

## Describe the situation before applying the recipe

```java
"" + 1
```

## Describe the situation after applying the recipe

```java
1
```

## OSS repro
- https://github.com/encryptedsystems/Clusion/blob/d6f89c1269f618ec42a9bfc42e940382fa14ae69/src/main/java/org/crypto/sse/DynRR.java#L47
- https://github.com/xebialabs/overcast/blob/8a494dc8f8572ae337ce91b06829eb9567c91f5d/src/main/java/com/xebialabs/overcast/host/CachedLibvirtHost.java#L57

Akanksha928:
Hi! I'd like to work on this issue. Let me know if it's still available. Thanks!

timtebeek:
Here's a quick first attempt:
- https://github.com/openrewrite/rewrite-static-analysis/pull/634/

Akanksha928:
Just to clarify, since the PR #634 is now open and seems to fully address this, should I consider this issue closed or still open for contribution?

PR #634, opened 2025-07-09
Replace empty String concatenation with `String.valueOf()`

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/628


timtebeek:
Seeing some perhaps undesirable effects:

PR #643, opened 2025-07-19
Add RemovePrivateFieldUnderscores recipe

### LINKED ISSUE #630: Recipe to remove underscores in private field names

Apply modern code standards.

I think it used to be a bit of a common practice to denote private class fields with names which either started with or ended with underscores. While technically allowed, I don't think this is mainstream in Java. Thus it makes sense to have a recipe to rename the fields not to have these `_` prefixes/suffixes.

```java
public class ParseLocation {
  private String _ruleName;

  public String getRuleName() {
    return _ruleName;
  }
  public void setRuleName(String ruleName) {
    _ruleName = ruleName;
  }
}
```

```java
public class ParseLocation {
  private String ruleName;

  public String getRuleName() {
    return ruleName;
  }
  public void setRuleName(String ruleName) {
    this.ruleName = ruleName;
  }
}
```

## Any additional context
- When making the changes, beware of name clashes. Even visible in the simple setter example above.
- For this reason, not marking this is a good-first-issue.

## OSS repro
- leading underscores: https://github.com/apache/pinot/blob/a02253d671d7781acc16b81628e51bd11a3005a0/pinot-spi/src/main/java/org/apache/pinot/spi/utils/builder/TableConfigBuilder.java#L90-L140
- trailing underscores: https://github.com/rstudio/rstudio/blob/0baa88b0d03238603c1adaa1fe96241cd54b1ba7/src/gwt/src/org/rstudio/studio/client/rmarkdown/events/RenderRmdEvent.java#L120-L129

jhl221123:
## What's changed?
This recipe removes prefix or suffix underscores from private class field names. This helps enforce modern Java coding standards.

For clarity and consistency, the recipe also adds a  qualifier to all updated field accesses. This prevents ambiguity with local variables and ensures a uniform style throughout the class.

## What's your motivation?
- Fixes #630

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemovePrivateFieldUnderscores.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemovePrivateFieldUnderscores.java
@@ -60,6 +60,10 @@ public class RemovePrivateFieldUnderscores extends Recipe {
                     String oldName = variable.getSimpleName();
                     if (oldName.startsWith("_") || oldName.endsWith("_")) {
                         String newName = oldName.startsWith("_") ? oldName.substring(1) : oldName.substring(0, oldName.length() - 1);
+                        if (newName.startsWith("_") || newName.endsWith("_") || newName.isEmpty()) {
+                            continue;
+                        }
+
                         doAfterVisit(new QualifyAmbiguousFieldAccess(variable, newName));
                         doAfterVisit(new RenameVariable(variable, newName));
                     }
--- a/src/test/java/org/openrewrite/staticanalysis/RemovePrivateFieldUnderscoresTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/RemovePrivateFieldUnderscoresTest.java
@@ -273,4 +273,19 @@ class RemovePrivateFieldUnderscoresTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void doesNotRenameToUnnamedVariable() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              class MyClass {
+                  private String __;
+                  private int ___;
+              }
+              """
+          )
+        );
+    }
 }


src/test/java/org/openrewrite/staticanalysis/RemovePrivateFieldUnderscoresTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+    @DocumentExample
+    @Test
+    void removesPrefixUnderscore() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              public class ParseLocation {
+                  private String _ruleName;
```
timtebeek, line 40: I think it would be good to add a test showing either `class_` or `_class` to show we do not end up renaming to restricted keywords.

jhl221123:
Thanks for the detailed feedback, @dsgrieve. I believe I've addressed everything now. Ready for another look when you have a moment.

dsgrieve (commented):
LGTM, but I defer to @timtebeek for approval.

src/main/java/org/openrewrite/staticanalysis/RemovePrivateFieldUnderscores.java
```diff
@@ -0,0 +1,153 @@
...
+                     null
+                   ),
+                   JLeftPadded.build(identifier.withPrefix(Space.EMPTY).withSimpleName(id.getSimpleName())),
+                   identifier.getType()
+                 );
+            }
+            return id;
+        }
+
+        private boolean isAmbiguous() {
```
greg-at-moderne, line 121: I am wondering if it wouldn't be better to re-use the existing logic of `org.openrewrite.java.VariableNameUtils#generateVariableName`. I'd call it with the `newName` as the `baseName` and refrain from any changes if the result is different.

PR #645, opened 2025-07-21
Unwrap else after return in preceding `if`

timtebeek:
Added a new recipe, polished it up, and ran it on this repository to flush out any potential issues. Seemed straightforward enough and nicely reduces needless indentation. Let me know if you agree and then we can run this more broadly, and enforce it going forward.

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/UnwrapElseAfterReturnTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/UnwrapElseAfterReturnTest.java
@@ -434,4 +434,165 @@ class UnwrapElseAfterReturnTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void simpleIfElseWithThrow() {
+        rewriteRun(
+          java(
+            """
+              class Test {
+                  void foo(boolean condition) {
+                      if (condition) {
+                          throw new IllegalArgumentException("Invalid condition");
+                      } else {
+                          System.out.println("Valid condition");
+                      }
+                  }
+              }
+              """,
+            """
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/staticanalysis/UnwrapElseAfterReturnTest.java
```diff
@@ -0,0 +1,374 @@
...
+                  int foo(String str) {
+                      if ("one".equals(str)) {
+                          return 1;
+                      } else if ("two".equals(str)) {
+                          return 2;
+                      } else if ("three".equals(str)) {
+                          return 3;
+                      } else {
+                          return Integer.MAX_VALUE;
+                      }
```
greg-at-moderne, line 351: Would be interesting to add a test case where only some of the `if`s have the `return` here. Like only the `two` branch. That's similar to `ifWithoutReturnNotChanged` test case, but with chaining.

greg-at-moderne (approved):
Another test case I'd add is with nested ifs in the primary branch. Something like:
```
if (someCondition) {
  // default logic
  if (somethingRare) {
     return "terminate all processing";
  }
} else {
  // non-standard logic
}
```
IMHO no changes should be made here.

greg-at-moderne:
Also, not sure if it should go into the same recipe, or a separate one. The same kind of analysis and changes could be made with `throw` statements.
```
if (creditCardNumber.length() < 16) {
   throw new IllegalArgumentException("No shopping for you");
} else {
   // here comes the real logic which could be unwrapped
}
```

Issue #623, opened 2025-07-07
Make setters return `this` to allow for fluent interfaces


greg-at-moderne:
## What problem are you trying to solve?

Nicer APIs.

## Describe the situation before applying the recipe
```java
public class Person {
    private int age;

    public void setAge(int age) {
        this.age = age;
    }
}
```

## Describe the situation after applying the recipe
```java
public class Person {
    private int age;

    public Person setAge(int age) {
       this.age = age;
    }
}
```

## Any additional context

- One could consider doing the same to other methods returning `void`, not only the setters.
- I haven't thought it through though, maybe some heuristics is needed to pick which methods it would be relevant for.

PR #646, opened 2025-07-21
Add FluentSetterRecipe

e5LA:
## What's changed?
Added new `FluentSetterRecipe` to convert void setter methods (and optionally other void methods) into fluent-style methods that return `this`.

## What's your motivation?
- Closes https://github.com/openrewrite/rewrite-static-analysis/issues/623

- Indentation and placement of the injected `return this;` statement.
- Return types use the simple name (e.g. `Inner` instead of `Outer.Inner`)

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
+++ b/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
@@ -29,6 +29,7 @@ import org.openrewrite.staticanalysis.java.MoveFieldAnnotationToType;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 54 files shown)


e5LA:
Thanks for the review @greg-at-moderne - that's good point, I will have a look.

e5LA:
Sorry for the delay, I didn't have much time lately.
While working on detecting whether a method overrides a superclass method, I realized there's a case we need to consider:

```
public class Parent {
    public void setName(String name) {
        this.name = name;
    }
}

public class Child extends Parent {
    private String name;

    @Override
    public void setName(String name) {
        this.name = name;
    }
}
```
In this situation, I see two options:

1. Either transform both `Parent#setName` and `Child#setName` to return `this` (+changing the return type) - that means we're explicitly transforming overridden methods.

1. Or skip transforming `Child#setName` because it’s an override - that seems straightforward.
However, the tricky part is with `Parent#setName` - it's not itself an override, but it is overridden. 
I wonder - can we reliably detect that it’s being overridden in a subclass?

timtebeek:
Thanks both! Wondering a bit indeed how we can make this safe; we don't immediately have a way to know if a method might be overridden, unless we use a ScanningRecipe, and even then we might miss cases where it's a library. Perhaps the safe thing to do would be to limit this to either the provided method pattern (opt in), or `final` classes when left out.

We can optionally and separately expand [FinalClass](https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/FinalClass.java#L28-L48) to make it a scanning recipe that adds that modifier to any classes not extended, that way the combination of these two recipes would achieve the safe results we're after.

PR #647, opened 2025-07-21
Recipe to inline returned value

timtebeek:
## What's your motivation?
- Extract part of https://github.com/openrewrite/rewrite-migrate-java/pull/795

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InlineReturnValue.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InlineReturnValue.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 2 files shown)


timtebeek:
Turns out we already have a recipe that does something similar, but with not exactly overlapping coverage:
https://github.com/openrewrite/rewrite-static-analysis/blob/a779afabaa0d12c77df16a384520e3a64bc762bd/src/main/java/org/openrewrite/staticanalysis/InlineVariable.java#L36-L46

src/test/java/org/openrewrite/staticanalysis/InlineVariableTest.java
```diff
@@ -252,4 +252,408 @@ int[] test() {
...
+                      result = "hello";
+                      return result;
+                  }
+              }
+              """,
+            """
+              class Test {
+                  String test() {
+                      String result;
+                      return "hello";
```
greg-at-moderne, line 274: Maybe we should trigger `RemoveUnusedLocalVariables` as an after visitor?

greg-at-moderne (approved):

PR #658, opened 2025-07-23
Add parentheses in `CompareEnumsWithEqualityOperator` as needed

### LINKED ISSUE #3: CompareEnumsWithEqualityOperator fails with negation
https://github.com/JabRef/jabref/blob/913b46cf0c31a4f7f73c84e4a8584c9adb8c701f/src/main/java/org/jabref/gui/journals/UndoableUnabbreviator.java#L61

```java
if ((!StandardField.JOURNAL.equals(field) && !StandardField.JOURNALTITLE.equals(field)) || !entry.hasField(AMSField.FJOURNAL)) {
```

gets rewritten to

```java
if ((!StandardField.JOURNAL != field && StandardField.JOURNALTITLE != field) || !entry.hasField(AMSField.FJOURNAL)) {
```

Which leads to a compile error

```
C:\git-repositories\jabref\src\main\java\org\jabref\gui\journals\UndoableUnabbreviator.java:61: Fehler: Ung├╝ltiger Operandentyp StandardField f├╝r un├ñren Operator "!"
        if ((!StandardField.JOURNAL != field && StandardField.JOURNALTITLE != field) || !entry.hasField(AMSField.FJOURNAL)) {
             ^
```

### LINKED ISSUE #657: CompareEnumsWithEqualityOperator needs parenthesis in some boolean expressions

* org.openrewrite:rewrite-core:8.58.0
* org.openrewrite.recipe:rewrite-static-analysis:2.12.0

In some boolean expressions, the output of CompareEnumsWithEqualityOperator needs to be wrapped in parentheses to ensure that the code produced is valid. https://github.com/openrewrite/rewrite-static-analysis/issues/3 is another example of this, and here another one found in our code base.

```java
  @Test
  public void enumEqual() {
    rewriteRun(
        spec -> spec.recipe(new CompareEnumsWithEqualityOperator()),
        java(
            """
            package com.helloworld;

            import java.time.DayOfWeek;

            public static class Foo {
              boolean foo = true == DayOfWeek.MONDAY.equals(DayOfWeek.TUESDAY);
            }
            """, """
            package com.helloworld;

            import java.time.DayOfWeek;

            public static class Foo {
              boolean foo = true == (DayOfWeek.MONDAY == DayOfWeek.TUESDAY);
            }
            """));
  }
```

```
org.opentest4j.AssertionFailedError: [Unexpected result in "com/helloworld/Foo.java":
diff --git a/com/helloworld/Foo.java b/com/helloworld/Foo.java
index daf2906..11a6884 100644
--- a/com/helloworld/Foo.java
+++ b/com/helloworld/Foo.java
@@ -3,5 +3,5 @@
 import java.time.DayOfWeek;

 public static class Foo {
-  boolean foo = true == (DayOfWeek.MONDAY == DayOfWeek.TUESDAY);
+  boolean foo = true == DayOfWeek.MONDAY == DayOfWeek.TUESDAY;
 }
\ No newline at end of file
]
expected:
  "package com.helloworld;

  import java.time.DayOfWeek;

  public static class Foo {
    boolean foo = true == (DayOfWeek.MONDAY == DayOfWeek.TUESDAY);
  }"
 but was:
  "package com.helloworld;

  import java.time.DayOfWeek;

  public static class Foo {
    boolean foo = true == DayOfWeek.MONDAY == DayOfWeek.TUESDAY;
  }"
	at org.openrewrite.test.RewriteTest.assertContentEquals(RewriteTest.java:622)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:511)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:130)
```

timtebeek:
- Fixes #657
- Closes #3

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperator.java
+++ b/src/main/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperator.java
@@ -58,14 +58,12 @@ public class CompareEnumsWithEqualityOperator extends Recipe {
                 if (enumEquals.matches(m) && m.getSelect() != null) {
                     Cursor parent = getCursor().dropParentUntil(is -> is instanceof J.Unary || is instanceof J.Block || is instanceof J.Binary || is instanceof J.Ternary || is instanceof J.Lambda);
                     boolean isNot = parent.getValue() instanceof J.Unary && ((J.Unary) parent.getValue()).getOperator() == J.Unary.Type.Not;
+                    boolean needsParentheses = false;
                     if (isNot) {
                         parent.putMessage("REMOVE_UNARY_NOT", parent.getValue());
-                    }
-
-                    // Check if the method invocation is part of a binary expression with equality/inequality operators
-                    // In such cases, we need parentheses to avoid chained equality operators
-                    boolean needsParentheses = false;
-                    if (!isNot && parent.getValue() instanceof J.Binary) {
+                    } else if (parent.getValue() instanceof J.Binary) {
+                        // Check if the method invocation is part of a binary expression with equality/inequality operators
+                        // In such cases, we need parentheses to avoid chained equality operators
                         J.Binary binary = parent.getValue();
                         J.Binary.Type op = binary.getOperator();
                         // Only add parentheses when the parent binary uses equality/inequality operators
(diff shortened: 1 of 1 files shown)


timtebeek:
We do, but in this case we only pass in the binary template, whereas the context that parentheses are necessary would only be visible if we were to use a context sensitive template right? Or should we still be able to work that out from the cursor position + coordinates?
https://github.com/openrewrite/rewrite-static-analysis/blob/9cbf2e88be2ce8fad72518d88e0084906edda4ea/src/main/java/org/openrewrite/staticanalysis/CompareEnumsWithEqualityOperator.java#L67-L73

knutwannheden:
I think it should be able to work it out using the cursor. At least the way I think I remember it, as I thought I added this for Refaster.

knutwannheden:
So possibly this is just a bug in the ParenthesizeVisitor.

PR #659, opened 2025-07-24
Recipe to simplify Boolean expressions using De Morgan's laws

greg-at-moderne:
## What's changed?

A new recipe called `SimplifyBooleanExpressionWithDeMorgan`.

## What's your motivation?

Add a new recipe, the trigger was I wanted to use it in the after-visit in one of the other recipes (#656).

I was wondering whether to plug it into existing `SimplifyBooleanExpression` as it would fit the basic description. I could see pros and cons, but decided not to. Please share your thoughts on this.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/SimplifyBooleanExpressionWithDeMorgan.java
+++ b/src/main/java/org/openrewrite/staticanalysis/SimplifyBooleanExpressionWithDeMorgan.java
@@ -87,18 +87,17 @@ public class SimplifyBooleanExpressionWithDeMorgan extends Recipe {
                         left = (Expression) this.visit(left, ctx);
                         right = (Expression) this.visit(right, ctx);
 
-                        if (newOperator != null) {
-                            Space prefix = unary.getPrefix();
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/SimplifyBooleanExpressionWithDeMorgan.java
```diff
@@ -0,0 +1,123 @@
...
+
+                        if (newOperator == null) {
+                            J.Binary visitedBinary = binary.withLeft(left).withRight(right);
+                            return unary.withExpression(parenthesesBinary.withTree(visitedBinary));
+                        }
+                        Space prefix = unary.getPrefix();
+                        List<Comment> comments = new ArrayList<>(prefix.getComments());
+                        comments.addAll(parenthesesBinary.getComments());
+                        comments.addAll(binary.getComments());
+                        prefix = prefix.withComments(comments);
```
timtebeek, line 97: `ListUtils.concat` might be somewhat preferred here, just to ensure that pattern is used where applicable as well.

PR #684, opened 2025-08-02
MissingOverrideAnnotation: removeFalsyOverrideAnnotation

Pankraz76:
## What's changed?

- https://docs.openrewrite.org/recipes/staticanalysis/missingoverrideannotation

It seems the rules does not cover false positives as well. Should this be included? 

## What's your motivation?

are those annotations really necessary

- https://github.com/apache/kafka/pull/20214
  - https://github.com/apache/kafka/pull/20214#discussion_r2249139705

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MissingOverrideAnnotation.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MissingOverrideAnnotation.java
@@ -17,6 +17,7 @@ package org.openrewrite.staticanalysis;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import lombok.var;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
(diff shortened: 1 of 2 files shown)


Pankraz76:
is this any good? @timtebeek

fell free to tune and complete @iddeepak

Pankraz76:
> aims to add missing override annotations

What about changing the scope to check for corrupt (missing/falsy) override annotations. 

We could create a new one like `CheckforFalsyOverrideAnnotation` but that would impose configuration burden on customer instead of applying convention. Once could argue that every recipe, could check for false negatives as well.

As if not, gonna have to copy potentially every recipe 1 times to check for false negative as well.

Pankraz76:
this was my intention:

https://github.com/apache/kafka/pull/20214/commits/2575f476fcb3576f05add3053c529db0a630b54b

Issue #665, opened 2025-07-28
Recipe to prefer numeric comparison `a == b` over `a -- b == 0`


greg-at-moderne:
## What problem are you trying to solve?

Code readability.

## Describe the situation before applying the recipe

```java
return a -- b == 0;
```

## Describe the situation after applying the recipe

```java
return a == b;
```

## Any additional context
`a -- b` might be in parentheses.

PR #686, opened 2025-08-03
Add PreferEqualityComparisonOverDifferenceCheck recipe

e5LA:
## What's changed?
Added a new recipe `PreferEqualityComparisonOverDifferenceCheck` that transforms expressions like `a - b == 0` into the more readable `a == b`.

## What's your motivation?
- Closes https://github.com/openrewrite/rewrite-static-analysis/issues/665 

- Name of the recipe :) 
- Logic for unwrapping parentheses around subtraction expression

Haven't really considered it - but alternative solution would be using `JavaTemplate`. I think using `withLeft` and `withRight` looks simpler though.

### Checklist


src/main/java/org/openrewrite/staticanalysis/PreferEqualityComparisonOverDifferenceCheck.java
```diff
@@ -0,0 +1,98 @@
...
+                    Expression unwrappedRight = unwrapParentheses(right);
+
+                    // Case 1: (a - b) == 0
+                    if (isSubtraction(unwrappedLeft) && isZero(unwrappedRight)) {
+                        J.Binary subtraction = (J.Binary) unwrappedLeft;
+                        return transformToEquality(binary, subtraction, ctx);
+                    }
+
+                    // Case 2: 0 == (a - b)
+                    if (isZero(unwrappedLeft) && isSubtraction(unwrappedRight)) {
```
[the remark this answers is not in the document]
e5LA: Thanks for the review @greg-at-moderne 

This line actually converts it to the "usual" syntax. It's just handling the reversed condition where someone writes `0 == a - b` instead of `a - b == 0`. 

I included this because the condition is mathematically equivalent and some developers may write it this way.
I can remove this if you think it's not worth supporting.

src/main/java/org/openrewrite/staticanalysis/PreferEqualityComparisonOverDifferenceCheck.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public JavaIsoVisitor<ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.Binary visitBinary(J.Binary binary, ExecutionContext ctx) {
+                binary = super.visitBinary(binary, ctx);
+
+                if (binary.getOperator() == J.Binary.Type.Equal) {
```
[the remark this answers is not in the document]
greg-at-moderne: Correct.

src/main/java/org/openrewrite/staticanalysis/PreferEqualityComparisonOverDifferenceCheck.java
```diff
@@ -0,0 +1,101 @@
...
+import org.openrewrite.Recipe;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+
+public class PreferEqualityComparisonOverDifferenceCheck extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Prefer direct comparison";
```
greg-at-moderne, line 28:
```suggestion
        return "Prefer direct comparison of numbers";
```

greg-at-moderne, line 28:
```suggestion
        return "Prefer direct comparison of numbers";
```

src/main/java/org/openrewrite/staticanalysis/PreferEqualityComparisonOverDifferenceCheck.java
```diff
@@ -0,0 +1,101 @@
...
+            public J.Binary visitBinary(J.Binary binary, ExecutionContext ctx) {
+                binary = super.visitBinary(binary, ctx);
+
+                J.Binary.Type operator = binary.getOperator();
+                if (isComparisonOperator(operator)) {
+                    Expression left = binary.getLeft();
+                    Expression right = binary.getRight();
+
+                    Expression unwrappedLeft = unwrapParentheses(left);
+                    Expression unwrappedRight = unwrapParentheses(right);
```
greg-at-moderne, line 51:
```suggestion
                    Expression unwrappedLeft = unwrapParentheses(binary.getLeft());
                    Expression unwrappedRight = unwrapParentheses(binary.getRight());
```

greg-at-moderne:
The CI failure seems unrelated. We'll wait and trigger the PR check again later.

greg-at-moderne:
Applied it to our main repo: https://github.com/openrewrite/rewrite/pull/5870

timtebeek:
Thanks both! Nice to see those fixes applied immediately. I'd briefly considered whether this would be a good Refaster style recipe, but I do see that become a bit of a combinatorial explosion with what we've covered here. Might still make for a nice compile time set of rules for @Stephan202 and [Picnicsupermarket/error-prone-support](https://github.com/Picnicsupermarket/error-prone-support/).

PR #689, opened 2025-08-05
Redeclare variables in new blocks when running MinimumSwitchCases

### LINKED ISSUE #687: `MinimumSwitchCases` does not redeclare variables in new blocks
Suppose we have a switch case:

```java
switch (someInt) {
    case 1:
            String data = getSomeString();
            doThingOneWith(data);
            break;
    case 2:
            data = getSomeOtherString();
            doThingTwoWith(data);
            break;
}
```

That will currently be rewritten by MinimumSwitchCases to

```java
if (someInt == 1) {
        String data = getSomeString();
        doThingOneWith(data);
} else if (someInt == 2) {
        data = getSomeOtherString();
        doThingTwoWith(data);
}
```

which does not compile due to `data` not being declared in the else block's scope.

Jenson3210:
## What's changed?
By storing (reducing) the variables declared in the first block, we can see that the usages in the second block not refer to them. By using `map.remove` we make sure that we only replace the first statement in the second block (if any)

## What's your motivation?
- Fixes #687 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
@@ -192,12 +192,8 @@ public class MinimumSwitchCases extends Recipe {
                         Map<String, J.VariableDeclarations> firstCaseDeclarations = new JavaIsoVisitor<Map<String, J.VariableDeclarations>>() {
                             @Override
                             public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable, Map<String, J.VariableDeclarations> variableDeclarations) {
-                                for (J.VariableDeclarations.NamedVariable var : multiVariable.getVariables()) {
-                                    if (multiVariable.getVariables().size() == 1) {
-                                        variableDeclarations.put(var.getSimpleName(), multiVariable);
-                                    } else {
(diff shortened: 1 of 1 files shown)


src/test/java/org/openrewrite/staticanalysis/MinimumSwitchCasesTest.java
```diff
@@ -887,4 +887,108 @@ void doSomethingElse() {}
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/687")
```
jevanlingen, line 891: TIL. Never thought this code would compile in the first place 🙈.

src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
```diff
@@ -192,6 +189,31 @@ public J visitSwitch(J.Switch switch_, ExecutionContext ctx) {
...
+                            }
+
+                            @Override
+                            public J.Block visitBlock(J.Block block, Map<String, J.VariableDeclarations> variableDeclarations) {
+                                for (Statement statement : block.getStatements()) {
+                                    if (statement instanceof J.VariableDeclarations) {
+                                        visitVariableDeclarations((J.VariableDeclarations) statement, variableDeclarations);
+                                    }
+                                }
+
```
jevanlingen, line 212:
```suggestion
```

Issue #683, opened 2025-07-31
Recipe to replace trivial if statements with ternary operator


greg-at-moderne:
## What problem are you trying to solve?

Code readability. Funnily, this is opposite of #50. Have a recipe to convert `if-else` statements to a ternary (Elvis) operator when applicable. 

In specific cases the ternary operator is more readable than a verbose if-else. I think we might want to apply some heuristics on the condition being simple and/or the values assigned/returned being simple.

## Describe the situation before applying the recipe
```java
        if (age >= 18) {
            category = "Adult";
        } else {
            category = "Minor";
        }
```

## Describe the situation after applying the recipe
```java
        category = (age >= 18) ? "Adult" : "Minor";
```

NatePombi:
Hi I'm a beginner in open source and I'd like to work on this issue as my first contribution. May I be assigned to it? Thanks!

PR #691, opened 2025-08-06
Add ReplaceIfElseWithTernary

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/683

### DIFF
--- a/src/main/resources/META-INF/rewrite/examples.yml
+++ b/src/main/resources/META-INF/rewrite/examples.yml
@@ -360,25 +360,6 @@ examples:
 - description: ''
   sources:
   - before: |
-      class A {
-          void foo() throws IllegalAccessException {
-              try {
-                  throw new IllegalAccessException();
-              } catch (Exception e) {
-                  throw e; // `e` is removed below
-              }
-          }
-      }
-    after: |
-      class A {
-          void foo() throws IllegalAccessException {
-              throw new IllegalAccessException();
-          }
-      }
-    language: java
-- description: ''
-  sources:
-  - before: |
       import java.io.FileReader;
       import java.io.IOException;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/staticanalysis/ReplaceIfElseWithTernary.java
```diff
@@ -0,0 +1,149 @@
...
+
+                // Don't convert if the then part contains another if statement (would create nested ternary)
+                if (thenStatement instanceof J.If) {
+                    return i;
+                }
+
+                // Check if both statements are assignments to the same variable
+                if (!(thenStatement instanceof J.Assignment) || !(elseStatement instanceof J.Assignment)) {
+                    return i;
+                }
```
greg-at-moderne, line 76: This could possibly also handle `return` statements.
Could be done in a separate PR though.

src/main/java/org/openrewrite/staticanalysis/ReplaceIfElseWithTernary.java
```diff
@@ -0,0 +1,149 @@
...
+            private boolean areAssignmentsToSameVariable(J.Assignment a1, J.Assignment a2) {
+                // Check if both assignments target the same variable
+                if (a1.getVariable() instanceof J.Identifier && a2.getVariable() instanceof J.Identifier) {
+                    J.Identifier id1 = (J.Identifier) a1.getVariable();
+                    J.Identifier id2 = (J.Identifier) a2.getVariable();
+                    return id1.getSimpleName().equals(id2.getSimpleName());
+                } else if (a1.getVariable() instanceof J.FieldAccess && a2.getVariable() instanceof J.FieldAccess) {
+                    J.FieldAccess fa1 = (J.FieldAccess) a1.getVariable();
+                    J.FieldAccess fa2 = (J.FieldAccess) a2.getVariable();
+                    return fa1.toString().equals(fa2.toString());
```
greg-at-moderne, line 125: Not sure if we should be comparing toStrings.

src/main/java/org/openrewrite/staticanalysis/ReplaceIfElseWithTernary.java
```diff
@@ -0,0 +1,149 @@
...
+                return false;
+            }
+
+            private boolean containsTernary(J expr) {
+                if (expr instanceof J.Ternary) {
+                    return true;
+                }
+
+                // Check if expression contains a ternary operator
+                boolean[] hasTernary = {false};
```
greg-at-moderne, line 136: The array seems like a hack. `AtomicBoolean` would be clearer.

PR #692, opened 2025-08-06
Fix RemoveExtraSemicolons to handle semicolons after imports

### LINKED ISSUE #669: RemoveExtraSemicolonsTest should remove extra semicolon after import too
## Expected
```
    @Test
    void afterImport() {
        rewriteRun(
          //language=java
          java(
            """
              import java.util.List;;
              interface A {
                  List<String> getList();
              }
            """,
            """
              import java.util.List;
              interface A {
                  List<String> getList();
              }
            """
          )
        );
    }
```

## Context
- the following test fails for two reasons now, one is that the semicolon is not removed, the other one is that extra `;` is parsed into whitespace as per https://github.com/openrewrite/rewrite/issues/5735#issuecomment-3095614920
- but the latter is a test-only issue in practice; so this issue can be solved without depending on the parsing issue

timtebeek:
## Summary
- Modified `RemoveExtraSemicolons` recipe to remove semicolons from all whitespace by overriding `visitSpace`
- Added test case to verify extra semicolons after imports are properly removed
- The solution handles the parsing issue where extra semicolons get placed into whitespace

## What's changed?
The recipe now removes any semicolons found in whitespace throughout the AST. This addresses the issue where extra semicolons after imports (like `import java.util.List;;`) are parsed into the whitespace rather than being parsed as separate tokens.

## Test plan
- [x] Added new test case `afterImport()` that verifies semicolons after imports are removed
- [x] All existing tests continue to pass
- [x] Test includes `TypeValidation.all().allowNonWhitespaceInWhitespace(true)` to handle the parser behavior

Fixes #669

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveExtraSemicolons.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveExtraSemicolons.java
@@ -21,6 +21,7 @@ import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.Comment;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JRightPadded;
 import org.openrewrite.java.tree.Space;
 import org.openrewrite.java.tree.Statement;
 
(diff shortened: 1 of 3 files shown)


timtebeek:
Already pushed to main in https://github.com/openrewrite/rewrite-static-analysis/commit/a802311b3916fbeb57c092ac81e03937677dff4b

PR #698, opened 2025-08-07
`MoveConditionsToWhile` recipe

### LINKED ISSUE #610: Recipe to amend `while true` with conditions found in conditional statements in the body
Simplify code and improve readability.

There are cases where `while (true) {}` loops make sense. But not all of them do.
In cases of `while` loop body starts with an `if` statement which decides to break the loop, the logic can be moved to the while loop condition.

## What precondition(s) should be checked before applying this recipe?
Probably to start with:
- the `if` is the first statement in the `while` loop
- the `if` body has only `break` and nothing else

I am not sure whether this recipe should kick in if there are multiple `break` statements in the loop. ~Probably not.~ EDIT: why not?

```java
        int counter = 0;

        while (true) {
            if (counter >= 5) {
                break;
            }

            System.out.println("Counter: " + counter);
            counter++;
        }
```

```java
        int counter = 0;

        while (!(counter >= 5)) {
            System.out.println("Counter: " + counter);
            counter++;
        }
```
Possibly the negation in the while condition could be simplified by delegating to other recipe.

## OSS repro
- https://github.com/AutoMQ/automq/blob/6143ec682c45edc6660e08833a5dd8239a29da98/generator/src/main/java/org/apache/kafka/message/SchemaGenerator.java#L133-L136
- https://github.com/ismartcoding/plain-app/blob/ad7cbf6bafd47596351cb3c9fc67437d047444f6/lib/src/main/java/com/ismartcoding/lib/androidsvg/utils/CSSFontFeatureSettings.java#L251-L253

greg-at-moderne:
## What's changed?

A new recipe `MoveConditionsToWhile`.

## What's your motivation?

- fixes #610

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MoveConditionsToWhile.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MoveConditionsToWhile.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2025 the original author or authors.
  * <p>
  * Licensed under the Moderne Source Available License (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


timtebeek:
Added a small bit of polish just now https://github.com/openrewrite/rewrite-static-analysis/commit/e775b10a5f2972e09168cba1eaf87c6d58149526

Issue #671, opened 2023-03-13
UseCollectionInterfaces changes type of variable passed as parameter to method in other compilation unit


arthurvl:
When the UseCollectionInterfaces recipe changes a variable's declared type, it does not take into account the variable's usage sites. If one of those usages happens to be defined in a different compilation unit (e.g. a maven dependency that is not in the source tree under analysis), and that usage's type requirement is not of an interface type but of a concrete collection type, the variable's declared type is still changed, leading to a subsequent compilation error on the usage site. Note that this does not happen for sources within a single analysis, as method declarations are also rewritten in that case.

That is

```java
public void passesConcreteType() {
    ArrayList<Integer> toPass = new ArrayList<>(Collections.singletonList(1));

    SomeExternal.methodRequiringArrayList(toPass);
}
```

will be rewritten to

```java
public void passesConcreteType() {
    List<Integer> toPass = new ArrayList<>(Collections.singletonList(1));

    SomeExternal.methodRequiringArrayList(toPass);
}
```

even when the externally imported SomeExternal class has a declaration like

```java
public class SomeExternal {
    public static void methodRequiringArrayList(ArrayList<Integer> formalParameter) { }
}
```

PR #709, opened 2025-08-14
fix: Enhanced UseCollectionInterfaces to prevent conversions when non interface methods are used

### LINKED ISSUE #688: `UseCollectionInterfaces` does not check for method calls not defined on the Interface type
On collection types that existed in the java standard library before the implementation of the Java Collections Framework (e.g. Vector and Hashtable), some methods exist that do not exist on the corresponding Java Collections interfaces (e.g. Vector::elements and Hashtable::keys). The UseCollectionInterfaces recipe does not check whether any of these methods are called when replacing a concrete declaration for one these classes with the corresponding collection interface.

Failing test:

```java
    @Test
    void usesMethodNotOnInterface() {
        rewriteRun(
          java(
            """
                import java.util.Enumeration;
                import java.util.Hashtable;

                public class A {
                    public Enumeration<Integer> usesMethodNotOnInterface() {
                        Hashtable<Integer,Object> table = new Hashtable<>();

                        return table.keys();
                    }
                }
              """
          )
        );
    }
```

e5LA:
## What's changed?
Enhanced `UseCollectionInterfaces` recipe to prevent type conversions when concrete-class-specific methods are being used. 
Also, added support for  `java.util.Vector`

## What's your motivation?
- Closes #688
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/671

- The `nonInterfaceMethods` to ensure we've covered all the important methods
- Two-phase approach, ie first detect and then transform. Whether this is the most efficient way

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UseCollectionInterfaces.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UseCollectionInterfaces.java
@@ -24,15 +24,10 @@ import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markers;
 
 import java.time.Duration;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 
-import static java.util.Collections.emptyList;
-import static java.util.Collections.singleton;
-import static java.util.Collections.unmodifiableSet;
+import static java.util.Collections.*;
 import static java.util.Objects.requireNonNull;
 import static org.openrewrite.Tree.randomId;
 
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks a lot for the clear fix here @e5LA ! I've simplified your approach a bit, but kept things functionally the same. Specifically you can pass in any second argument to your new visitor. An AtomicBoolean seemed the best fit. Also when you then do not call `super.visit(Tree,...)` there's no need for guards in your other visit methods, as those will not get called. Finally I found another disabled unit test that can be fixed with a small addition for a dual fix here. Great!

e5LA:
Thanks @timtebeek for the review and the improvements! Your changes are elegant and more efficient, the reducer pattern is much cleaner. Appreciate you taking the time to show the proper OpenRewrite patterns.

And great we could close two issues with this one PR 👍

PR #711, opened 2025-08-15
Improve support method invocations for `RemoveRedundantTypeCast`

jevanlingen:
## What's changed?
- If a cast is performed with the **exact** same type as the expression, remove the cast.
- if a cast is performed on a methodCall and the type is a subtype of the expression, remove the cast.

## What's your motivation?
- https://github.com/moderneinc/rewrite-hibernate/pull/28/files#r2266719230

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveRedundantTypeCast.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveRedundantTypeCast.java
@@ -75,7 +75,7 @@ public class RemoveRedundantTypeCast extends Recipe {
                 JavaType castType = visitedTypeCast.getType();
 
                 JavaType targetType = null;
-                // Null-check; because for Kotlin's Gradle files, we don't always have all type information available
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Looks good, thanks! Have you tried this out at scale against say Apache to ensure we've not missed necessary types past what the tests cover? If so then good to merge.

jevanlingen:
> Have you tried this out at scale against say Apache to ensure we've not missed necessary types past what the tests cover?

I did not. So I run it against Apache repositories!

Turned out a `@NonNull`field could still be null for Kotlin Gradle files. I fixed it + wrote a small test to show the recipe does do something for Kotlin as well: https://github.com/openrewrite/rewrite-static-analysis/pull/711/commits/4af1a1c69cbe1c54431ef8d01d9d92beb098b0a0. I am wondering though, do we really want to support Kotlin for this recipe or should we add an exclusion?

PR #714, opened 2025-08-19
Use classpath from resource for Refaster

jevanlingen:
## What's changed?
Refaster can use type tables.

Mirrors change in
- https://github.com/openrewrite/rewrite-apache/pull/97

## Any additional context
This repository defines Refaster recipes, but those recipes don't need extra jars at all. With other words, there was no `.javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))` call needed. So this PR just adds the option to use the `classpathFromResources`, but actually never uses it.

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,3 +47,7 @@ dependencies {
 tasks.withType<Test> {
     jvmArgs("-Xmx1g", "-Xms512m")
 }
+
+tasks.withType<JavaCompile> {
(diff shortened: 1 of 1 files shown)


timtebeek:
Strictly speaking we didn't need this here as there are no dependencies used, but it's good to be consistent until we flip the default to use `classpathFromResources` always, at which point we can remove this annotation processor argument again.

Issue #666, opened 2025-07-28
Recipe to simplify `else` branch if it only has a single `if`


greg-at-moderne:
## What problem are you trying to solve?
Code readability.

## Requirements
- `if` is the only statement in the `else` block

## Describe the situation before applying the recipe

```java
        if (password.length() < 6) {
            System.out.println("Password is too short.");
        } else {
            if (password.length() > 12) {
                System.out.println("Password is too long.");
            }
        }
```

## Describe the situation after applying the recipe

```java
        if (password.length() < 6) {
            System.out.println("Password is too short.");
        } else if (password.length() > 12) {
            System.out.println("Password is too long.");
        }
```

PR #726, opened 2025-08-28
Simplify `else` branch if it only has a single `if` (#666)

olwispe:
## What's changed?
Implemented #666

Maybe the way comments are handled between the `else` and the `if`.
They usually apply to the whole `else` branch.
With the simplification they could appear to apply only to the `then` part. 

@greg-at-moderne, @timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/SimplifyElseBranch.java
+++ b/src/main/java/org/openrewrite/staticanalysis/SimplifyElseBranch.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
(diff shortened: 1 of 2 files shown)


greg-at-moderne (approved):

greg-at-moderne:
Thanks for the contribution, @olwispe!
Looks nice! Approving. And if Tim doesn't have any other comments, we'll have it merged.

timtebeek (approved):
Looks good indeed, thanks a lot! My only thought is if we should make this part of a larger composite recipe, as otherwise folks would have to deliberately seek out and run this recipe. Perhaps best done after we've built up some experience with the recipe and seen it work well at scale.

PR #727, opened 2025-08-28
fix RemoveUnusedPrivateMethods: getHostDottedClassName

Pankraz76:
### Checklist


src/test/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethodsTest.java
```diff
@@ -103,6 +103,488 @@ private void dontRemove2() {
...
+                        }
+                    }
+                    return list;
+                }
+            }
+            """
+          )
+        );
+    }
+
```
timtebeek, line 587: Can we reduce this down to just the relevant bits? We surely don't need 400 lines just to replicate an issue. A lot of this will not be relevant, but still needs to be evaluated whenever we try to mentally parse this test.

src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java
```diff
@@ -75,53 +75,100 @@ private boolean unusedWarningsSuppressed(J classDeclaration) {
...
+                // If we're missing the enclosing class or CU, bail out.
+                J.ClassDeclaration classDeclaration = getCursor().firstEnclosing(J.ClassDeclaration.class);
+                if (classDeclaration == null) {
+                    return m;
+                }
+                JavaSourceFile cu = getCursor().firstEnclosingOrThrow(JavaSourceFile.class);
+
+                // If referenced anywhere in TypesInUse (e.g., via other classes), keep it.
+                for (JavaType.Method usedMethodType : cu.getTypesInUse().getUsedMethods()) {
+                    if (methodType.equals(usedMethodType)) {
```
timtebeek, line 108: Why are you changing unrelated lines here and above? This removal will lead to worse performance, based on
- https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L1563-L1572
- https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L576-L583

Pankraz76: thanks for the hint, yes will try to undo

src/test/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethodsTest.java
Pankraz76: yes first idea was to keep code like it is due to licence, will try to shrink as well.

Issue #537, opened 2025-04-30
Replace nested ifs with one if and conditions joined with &&


greg-at-moderne:
## What problem are you trying to solve?

Simplify code.

## Describe the situation before applying the recipe
```java
    class A {
        void foo(boolean a, boolean b) {
            if (a) {
                if (b) {
                }
            }
        }
    }
```

## Describe the situation after applying the recipe

```java
    class A {
        void foo(boolean a, boolean b) {
            if (a && b) {
            }
        }
    }
```

## Things to take into account

- parenthesis might or might not be needed
- only applicable if there's no other logic in the outer `if`

PR #734, opened 2025-09-08
Mergeable `if` statements should be combined - RSPEC-S1066

olwispe:
## What's changed?

This PR implements [RSPEC-S1066](https://sonarsource.github.io/rspec/#/rspec/S1066/java) which was requested in :
- openrewrite/rewrite-static-analysis#45
- openrewrite/rewrite-static-analysis#537

## What's your motivation?

Among others, help migrate to pattern matching for `instanceof` when combined with `org.openrewrite.staticanalysis.InstanceOfPatternMatch` :
```java
if (o instanceof String) {
  String s = (String) o;
  if (s.isEmpty()) {
     System.out.println("OK");
  }
}
```
```java
if (o instanceof String s && s.isEmpty()) {
  System.out.println("OK");
}
```

@timtebeek 

### Checklist


timtebeek (commented):
Great to see @olwispe ! Added some light polish and I've been looking at the changes at scale.

Looking at these two examples I wonder if perhaps we should move the inner if conditional to a new line, and keep the comments in more of their original position.

image

image

image

I've pushed a quick test update that shows a comment lost. What are your thoughts to the above proposal?

PR #736, opened 2025-09-09
721: Recipe to move fields to the top of class definition

### LINKED ISSUE #721: Recipe to move fields to the top of class definition

Increase readability. Follow conventions.

```java
class A {
    void foo(String bar) {
        int i = Integer.valueOf(bar);
    }

    String myField = "important";
}
```

```java
class A {
    String myField = "important";

    void foo(String bar) {
        int i = Integer.valueOf(bar);
    }
}
```

dsgrieve:
## What's changed?

Added MoveFieldsToTopOfClass recipe and tests. This recipe reorders class members so that all field declarations appear before any method declarations constructors, or other class members. This improves code organization and readability by  grouping field declarations together at the top of the class. Comments associated with fields are preserved during the reordering.

## What's your motivation?

Fixes #721 

## Any additional context

This PR adds the recipe to static-analysis.yml. I'm not sure if that should be done in a separate PR, or not done at all. 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MoveFieldsToTopOfClass.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MoveFieldsToTopOfClass.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.staticanalysis;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.Statement;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+
+@EqualsAndHashCode(callSuper = false)
+@Value
+public class MoveFieldsToTopOfClass extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Move fields to the top of class definition";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Reorders class members so that all field declarations appear before any method declarations, " +
+               "constructors, or other class members. This improves code organization and readability by " +
+               "grouping field declarations together at the top of the class. Comments associated with fields " +
+               "are preserved during the reordering.";
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(2);
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/staticanalysis/MoveFieldsToTopOfClass.java
```diff
@@ -0,0 +1,136 @@
...
+                for (Statement statement : statements) {
+                    if (statement instanceof J.VariableDeclarations) {
+                        fieldDeclarations.add(statement);
+                    } else {
+                        otherStatements.add(statement);
+                    }
+                }
+
+                // If all fields are already at the top, no changes needed
+                if (fieldDeclarations.isEmpty() || isAlreadyOrdered(statements, fieldDeclarations)) {
```
timtebeek, line 79: The same is checked in the method called.
```suggestion
                if (isAlreadyOrdered(statements, fieldDeclarations)) {
```

src/main/java/org/openrewrite/staticanalysis/MoveFieldsToTopOfClass.java
```diff
@@ -0,0 +1,136 @@
...
+                        Space newPrefix = firstStatementPrefix.withComments(originalPrefix.getComments());
+                        field = field.withPrefix(newPrefix);
+                    }
+                    reorderedStatements.add(field);
+                }
+
+                // Add other statements
+                reorderedStatements.addAll(otherStatements);
+
+                return autoFormat(cd.withBody(cd.getBody().withStatements(reorderedStatements)), ctx);
```
timtebeek, line 104: do we really need the autoformat here? I feel that might cause unexpected formatting on inconsistent projects.

src/main/java/org/openrewrite/staticanalysis/MoveFieldsToTopOfClass.java
```diff
@@ -0,0 +1,174 @@
...
+                    statements.set(0, firstStatement);
+                }
+
+                // ensure proper spacing before the first non-field statement
+                IntStream.range(1, statements.size())
+                    .filter(i -> !(statements.get(i) instanceof J.VariableDeclarations))
+                    .findFirst()
+                    .ifPresent(i -> statements.set(i, ensurePrecedingNewline(statements.get(i))));
+
+                return cd.withBody(cd.getBody().withStatements(statements));
```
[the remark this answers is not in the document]
dsgrieve: This recipe is no longer trying to ensure a new line. It just reorders. If a user want to fix line spacing, they can run the "BlankLines" recipe.

dsgrieve:
The current issue is keeping a static intializer with the variable it's initializing. For instance, this simple test case will fail. 
```
    @Test
    void staticInitializerStaysWithDeclaration() {
        rewriteRun(
          //language=java
          java(
            """
              class WithStaticInitializer {
                  public void method() {
                      System.out.println("method");
                  }

                  private static final String field1;
                  static {
                      field1 = "value1";
                  }

                  public WithStaticInitializer() {
                      // constructor
                  }
              }
              """,
            """
              class WithStaticInitializer {

                  private static final String field1;
                  static {
                      field1 = "value1";
                  }
                  public void method() {
                      System.out.println("method");
                  }

                  public WithStaticInitializer() {
                      // constructor
                  }
              }
              """
          )
        );
    }
```

In addition, I would expect a static initializer block that initializes more than one variable to appear after the last variable declared (of those being initialized). I would also expect that other static blocks that aren't initializing a variable (doing something you want done when the class is first loaded) to come after all the variable declarations. None of these are being handled. 

Also, as @Jenson3210  points out, this could be a reorder statements recipe - which it is tending toward anyway. I'll add that how the statements are ordered should be a Style (even for the field decls, the ordering for which follows [Oracle code conventions](https://www.oracle.com/java/technologies/javase/codeconventions-fileorganization.html#:~:text=First%20the%20public%20class%20variables,and%20understanding%20the%20code%20easier.)

Lastly, this recipe in its current state messes up the formatting. But, as discussed with @sambsnyd, trying to maintain spacing seems overreach for this recipe. 

I feel, however, that without handling statement reordering and formatting this recipe in its current incantation just makes a mess of things. In short, moving the fields to the top is fairly trivial. Doing it correctly, however, is a bit tricky.

PR #738, opened 2025-09-19
AnnotateNullableMethods: annotate methods that return nullable method…

stefanodallapalma:
This change extends the `AnnotateNullableMethods` recipe to detect when a method returns the result of calling another method that is already annotated with `@Nullable`. In such cases, the calling method is now automatically annotated as nullable as well.

**Scope:** This enhancement is specifically limited to direct method invocations in return statements. It does not yet handle cases where the nullable method call is part of a more complex expression (e.g., ternary operations, assignments, or chained calls) - these may be addressed in future iterations.

**Example:**
```java
// Before
@Nullable String maybeNullString() { return null; }
String getString() { return maybeNullString(); } // Not annotated

// After  
@Nullable String maybeNullString() { return null; }
@Nullable String getString() { return maybeNullString(); } // Now annotated
```

This improvement ensures more comprehensive nullable annotation coverage by following the nullable contract through method call chains for direct return scenarios.

@timtebeek 

- Relates to https://github.com/openrewrite/rewrite-static-analysis/pull/569

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
+++ b/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableMethods.java
@@ -183,7 +183,7 @@ public class AnnotateNullableMethods extends Recipe {
                 return ((J.Literal) returnExpression).getValue() == null;
             }
             if (returnExpression instanceof J.MethodInvocation) {
-                return isNullable((J.MethodInvocation) returnExpression) || isKnownNullableMethod((J.MethodInvocation) returnExpression);
+                return isLocalNullableMethod((J.MethodInvocation) returnExpression) || isKnownNullableMethod((J.MethodInvocation) returnExpression);
             }
             if (returnExpression instanceof J.Ternary) {
                 J.Ternary ternary = (J.Ternary) returnExpression;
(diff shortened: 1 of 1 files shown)


timtebeek:
Neat addition, thanks! I've added some light polish & clarification. I had hoped we'd have that `Nullable` annotation on the method invocation return type, but alas; then indeed a local search as you've done could help uncover more cases. Could be worth exploring if we can make the method invocation return type contain the annotations as well, as would then also pick up nullable  methods from other types.

stefanodallapalma:
> I had hoped we'd have that `Nullable` annotation on the method invocation return type [...] Could be worth exploring if we can make the method invocation return type contain the annotations as well, as would then also pick up nullable methods from other types.

Thanks @timtebeek. Indeed that would be a great addition!

> 
> A quick run against some OpenRewrite repositories already uncovered this case that was missed previously:
> 
> * [openrewrite/rewrite-templating@02cc7b0](https://github.com/openrewrite/rewrite-templating/commit/02cc7b03b9ef8eea0c609866bf0788d08731deb6)

It's important to note (or perhaps clarify in the doc?) that devs would benefit from this change especially when running the recipe multiple times to allow incremental annotations

PR #739, opened 2025-09-21
Naming conflict in `InstanceOfPatternMatch`

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/619

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
+++ b/src/main/java/org/openrewrite/staticanalysis/InstanceOfPatternMatch.java
@@ -289,11 +289,11 @@ public class InstanceOfPatternMatch extends Recipe {
             return replacements.isEmpty() && variablesToDelete.isEmpty();
         }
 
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/rewrite/examples.yml
```diff
@@ -1212,6 +1213,61 @@ examples:
...
+      }
+      class RightNode {
+          int bar() {
+              return 1;
+          }
+      }
+
+      class Foo {
+          void bar(Object o1, Object o2) {
+              if (o1 instanceof LeftNode node2 && o2 instanceof RightNode node3) {
```
knutwannheden, line 1258: Surprising to see node2 and node3 before node and node1.

timtebeek: Yes that slightly annoyed me as well, but figured since we're already a few levels deep into edge cases (classes with the same suffix, used in binaries, with an else of overlapping instanceof expressions) I didn't want to explore how to switch those around and risk potentially breaking other logic that had assumed a particular processing order. From a brief look into it it's not as easy as moving down a super.visit call unfortunately.

PR #743, opened 2025-09-27
Add `Objects.requireNonNullElse/ElseGet` support to `AnnotateNullableParameters`

timtebeek:
## Summary
- Enhanced `AnnotateNullableParameters` recipe to recognize `Objects.requireNonNullElse()` and `Objects.requireNonNullElseGet()` calls as indicators of nullable parameters
- These methods provide default values when arguments are null, implying they accept nullable parameters

## Why this change?
- `Objects.requireNonNull()` throws an exception if the argument is null, enforcing non-null parameters
- `Objects.requireNonNullElse()` and `Objects.requireNonNullElseGet()` provide default values for null arguments, indicating the parameters can be null
- Parameters that use these fallback methods should be annotated with `@Nullable`

## Changes
- Added `Objects.requireNonNullElse(..)` and `Objects.requireNonNullElseGet(..)` to the list of null-checking method matchers
- Implemented `visitMethodInvocation` to handle standalone calls (not just those in if conditions)
- Added comprehensive test coverage for both methods

## Test plan
- [x] Added test for `Objects.requireNonNullElse` with default values
- [x] Added test for `Objects.requireNonNullElseGet` with supplier functions
- [x] Added test for multiple parameters with mixed usage of both methods
- [x] All existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableParameters.java
+++ b/src/main/java/org/openrewrite/staticanalysis/AnnotateNullableParameters.java
@@ -167,6 +167,7 @@ public class AnnotateNullableParameters extends Recipe {
      * <ul>
      *   <li>Direct null comparisons (param == null, param != null)</li>
      *   <li>Known null-checking method calls (Objects.isNull, StringUtils.isBlank, etc.)</li>
+     *   <li>Objects.requireNonNull calls that throw on null parameters</li>
      *   <li>Negated null-checking method calls (!Objects.isNull, !StringUtils.isBlank, etc.)</li>
      * </ul>
      */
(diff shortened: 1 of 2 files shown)


timtebeek:
Not sure I agree with Claude here; when using `requireNonNull` I'd expect the variable to _not_ be nullable, as we don't expect folks to rely on the NPE that gets thrown.

timtebeek:
@stefanodallapalma this seems right up your alley ; any suggestions or concerns with this addition?

stefanodallapalma:
> Not sure I agree with Claude here; when using `requireNonNull` I'd expect the variable to _not_ be nullable, as we don't expect folks to rely on the NPE that gets thrown.

@timtebeek I had a similar concern [here](https://github.com/openrewrite/rewrite-static-analysis/pull/578#issuecomment-2921781312). 

`Objects.requireNonNullElse()` and `requireNonNullElseGet ` make sense. Now I realized we could support `Optional.ofNullable()` too

timtebeek:
Thanks yes `requireNonNull` is quite the opposite, but `requireNonNullElse*` and `Optional.ofNullable` do indicate nullable parameters I think, or at the very least defensive coding. From your comment it's not yet clear whether you'd agree with the changes made here for those specific methods (not including `requireNonNull` from an earlier iteration). Is the current diff ok to merge you think?

stefanodallapalma:
Oh yeah, thanks. The current diff looks good to me. I only left a minor comment

PR #752, opened 2025-10-08
No longer remove String.valueof when it's called on a method invocation

Laurens-W:
## What's changed?

No longer remove `String.valueof` when it's called on a method invocation
- Also added a testcase here as follow-up to https://github.com/openrewrite/rewrite/pull/6111 (which means the build here will fail until that PR is merged)

## What's your motivation?

When the result of the method invocation is nullable it would result in `"null"` instead of `null` after the recipe made changes

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/staticanalysis/NoValueOfOnStringTypeTest.java
+++ b/src/test/java/org/openrewrite/staticanalysis/NoValueOfOnStringTypeTest.java
@@ -249,23 +249,21 @@ class NoValueOfOnStringTypeTest implements RewriteTest {
           java(
             """
               class Test {
-                  static void method1() {
+                  static void unnecessary() {
                       String a = String.valueOf(method2());
                   }
 
-                  static String method2() {
-                      return "";
-                  }
-              }
-              """,
-            """
-              class Test {
-                  static void method1() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/staticanalysis/NoValueOfOnStringType.java
```diff
@@ -70,7 +70,7 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                 J.MethodInvocation mi = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
                 if (VALUE_OF.matches(mi) && mi.getArguments().size() == 1) {
                     Expression argument = mi.getArguments().get(0);
-                    if (TypeUtils.isString(argument.getType()) || removeValueOfForStringConcatenation(argument)) {
+                    if ((TypeUtils.isString(argument.getType()) && !(argument instanceof J.MethodInvocation)) || removeValueOfForStringConcatenation(argument)) {
```
Laurens-W, line 73: I didn't look into it too much but if we could determine here whether the result of the method invocation is actually nullable that would allow us to make more transformations

PR #782, opened 2025-11-13
Annotate required parameters with `@NonNull` and remove throws

timtebeek:


timtebeek:
Still working through the details here, but @stefanodallapalma I think you'd maybe enjoy this one. Discovered the pattern on
- https://github.com/jobrunr/jobrunr/pull/1429

Had you seen something like this before on your explorations there?

PR #834, opened 2026-03-19
Add SillyEqualsCheck recipe (RSPEC-S2159)

timtebeek:
## Summary

Implements SonarQube rule S2159 "Silly equality checks should not be made", which detects `.equals()` calls that always return false.

The recipe transforms equals(null) checks to == null comparisons and replaces array.equals() with Arrays.equals(). It also flags comparisons between unrelated types and array-vs-non-array comparisons as search results.

## Changes

- New recipe: `SillyEqualsCheck` with comprehensive unit tests
- Handles negation patterns (!x.equals(null) → x != null)
- Auto-fixes: null checks and array equality comparisons
- Search results: unrelated types and mixed array/non-array comparisons


steve-aom-elliott (approved):
Makes sense to me. Might want to update the copyright years to 2026, but that's about it.

PR #836, opened 2026-03-19
Add S2209/S3252 recipe: Static members via class name

timtebeek:
## Summary
Implements OpenRewrite recipe to fix SonarQube rules S2209 (MAJOR) and S3252 (MINOR). Replaces static field/method access through instance references with access via the declaring class name (e.g., `instance.staticField` → `ClassName.staticField`, `instance.staticMethod()` → `ClassName.staticMethod()`).

Includes JavaIsoVisitor implementation handling field access and method invocations, with conservative side-effect preservation (skips method calls and `new` expressions as select targets). Full test coverage with 10 test cases.


src/test/java/org/openrewrite/staticanalysis/StaticAccessViaInstanceTest.java
```diff
@@ -0,0 +1,397 @@
...
+                MyClass field;
+                static MyClass getInstance() { return new MyClass(); }
+                void foo() {
+                    int x = getInstance().field.COUNT;
+                }
+            }
+            """
+          )
+        );
+    }
```
steve-aom-elliott, line 396: Isn't the starting state going to cause a null pointer exception? I suppose it doesn't necessarily matter for detection here, but we know we never set `field` to anything.

steve-aom-elliott (approved):
I think it makes sense. I suppose to further be able understand whether the method invocation chained into field access would have been safe to swap out with just a field access we would need more detailed analysis of the flow present.

src/test/java/org/openrewrite/staticanalysis/StaticAccessViaInstanceTest.java
timtebeek: Indeed it's a heavily reduced example, but the compiler allows it, and it's enough for detection. Thanks for the ask though!

PR #837, opened 2026-03-19
Add S1185 recipe: Remove methods that only call super

timtebeek:
Implements RSPEC-S1185 to detect and remove methods that override a parent method but only delegate to super with the same arguments. Methods are safe to remove only if they have no semantic annotations, no Javadoc, and don't widen visibility.

Includes 12 comprehensive test cases covering positive and negative scenarios.


src/test/java/org/openrewrite/staticanalysis/RemoveMethodsOnlyCallSuperTest.java
```diff
@@ -0,0 +1,395 @@
...
+                  }
+              }
+              """,
+            """
+              class Child extends Parent {
+              }
+              """
+          )
+        );
+    }
```
steve-aom-elliott, line 154: Is this test redundant or was it meant to show something else?

src/test/java/org/openrewrite/staticanalysis/RemoveMethodsOnlyCallSuperTest.java
```diff
@@ -0,0 +1,395 @@
...
+                  @Deprecated
+                  @Override
+                  void foo() {
+                      super.foo();
+                  }
+              }
+              """
+          )
+        );
+    }
```
steve-aom-elliott, line 183: Why wouldn't we want to remove deprecated overrides?

src/test/java/org/openrewrite/staticanalysis/RemoveMethodsOnlyCallSuperTest.java
timtebeek: Yes it was, thanks!

src/test/java/org/openrewrite/staticanalysis/RemoveMethodsOnlyCallSuperTest.java
timtebeek: now also allow deprecated, but retain methods annotated with transactional

steve-aom-elliott (approved):

PR #854, opened 2026-04-11
Add space after // in single-line comments

avimta:
This recipe ensures that single-line comments have a single space after `//` when missing.

### Behavior:
- Converts `//Hello` → `// Hello`
- Leaves already formatted comments unchanged
- Skips special cases like `//language=java`

This improves consistency and readability of comments.

All tests pass successfully.


src/main/java/org/openrewrite/staticanalysis/SingleLineCommentSpacing.java
```diff
@@ -0,0 +1,113 @@
...
+            @Override
+            public Space visitSpace(Space space, Space.Location loc, ExecutionContext ctx) {
+                Space s = super.visitSpace(space, loc, ctx);
+
+                List<Comment> comments = s.getComments();
+                if (comments.isEmpty()) {
+                    return s;
+                }
+
+                boolean changed = false;
```
timtebeek: Instead of a boolean changed, let's use ListUtils.map

avimta: Thanks for the suggestion, 
I've updated the implementation to use ListUtils.map.

timtebeek (approved):
Thanks a lot @AVIMTA ! I've polished up your work a bit, mostly to minimize what we have to maintain going forward, and to handle a few more edge cases. With that we're good to merge!

This should go out in the next release; expected at most two weeks from now; until then you can use [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions).

PR #858, opened 2026-04-12
Fix UnnecessaryExplicitTypeArguments within lambda return statements

### LINKED ISSUE #670: `UnnecessaryExplicitTypeArguments` should work within lambdas too
## Problem statement

See `org.openrewrite.staticanalysis.UnnecessaryExplicitTypeArgumentsTest#withinLambda`.

https://github.com/openrewrite/rewrite-static-analysis/blob/0a470e4e038a77c405fedc5e825331e1992bdc49/src/test/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArgumentsTest.java#L77-L107

timtebeek:
- Closes #670

## Summary
- Fix `UnnecessaryExplicitTypeArguments` to properly remove unnecessary explicit type arguments inside lambda return statements
- The previous code used the lambda's functional interface type directly instead of resolving the SAM method's return type against the parameterized type
- Now correctly finds the Single Abstract Method, resolves generic type variables to actual type parameters, and compares against the method invocation's return type

## Test plan
- [x] Existing `withinLambda()` test now passes (previously `@ExpectedToFail`)
- [x] All other `UnnecessaryExplicitTypeArgumentsTest` tests still pass

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
@@ -90,7 +90,38 @@ public class UnnecessaryExplicitTypeArguments extends Recipe {
                             inferredType = methodDeclaration.getReturnTypeExpression().getType();
                         }
                     } else if (e instanceof J.Lambda) {
-                        inferredType = ((J.Lambda) e).getType();
+                        JavaType.Parameterized parameterized = TypeUtils.asParameterized(((J.Lambda) e).getType());
+                        if (parameterized != null) {
+                            JavaType.Method sam = null;
+                            for (JavaType.Method candidate : parameterized.getMethods()) {
+                                if (candidate.hasFlags(Flag.Default) || candidate.hasFlags(Flag.Static)) {
+                                    continue;
+                                }
+                                if (sam != null) {
+                                    sam = null;
+                                    break;
+                                }
+                                sam = candidate;
+                            }
+                            if (sam != null) {
+                                JavaType samReturn = sam.getReturnType();
+                                if (samReturn instanceof JavaType.GenericTypeVariable) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
```diff
@@ -90,7 +90,38 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                            for (JavaType.Method candidate : parameterized.getMethods()) {
+                                if (candidate.hasFlags(Flag.Default) || candidate.hasFlags(Flag.Static)) {
+                                    continue;
+                                }
+                                if (sam != null) {
+                                    sam = null;
+                                    break;
+                                }
+                                sam = candidate;
+                            }
```
greg-at-moderne: I fail to understand the shape of the loop here.
Line 104 saves the `sam`, but doesn't return it from the loop, only to wait to some next iteration to end the loop with `sam == null`.

I am not saying this is wrong, but this is hard to understand.

timtebeek: The way I read it is that any second assignment implies ambiguity, which then nulls out the reference to avoid changes. So we only make changes if w find a single one. Does that reading help you there?

greg-at-moderne: OK. That helps. Then I suggest a refactor to extract a method called `findMethodIfUnambiguous` or similar.

timtebeek: done, thanks! feel free to merge if you now agree 🛫

greg-at-moderne (approved):

PR #898, opened 2026-05-20
Fix RemoveUnusedPrivateMethods deleting methods called with raw collection arguments

### LINKED ISSUE #877: RemoveUnusedPrivateMethods removes private method still called with raw collection arguments

- rewrite-static-analysis: 2.34.0
- rewrite-core: 8.81.1
- rewrite-java: 8.81.1

This is an edge case involving raw collection arguments. The raw `Map` is important here; a parameterized `Map<String, String>` does not reproduce the issue.

```java
import java.util.Collection;
import java.util.Map;
import java.util.Set;

class Main {
  Main(Map templates) {
    compileTemplates(templates.keySet(), templates.values());
  }

  private Set<String> compileTemplates(
      final Set<String> compiledParam, final Collection<String> toCheck) {
    return compiledParam;
  }
}
```

Run `org.openrewrite.staticanalysis.RemoveUnusedPrivateMethods`.

The private `compileTemplates(..)` method should be preserved because it is called from the constructor.

Although the call uses raw `Map#keySet()` and `Map#values()` values, the original code compiles with unchecked/rawtype warnings.

## What did you see instead?

The recipe removes the private method and its imports, but leaves the call site behind:

```java
import java.util.Map;

class Main {
  Main(Map templates) {
    compileTemplates(templates.keySet(), templates.values());
  }
}
```

The rewritten code no longer compiles because `compileTemplates(..)` no longer exists.

habiibullahm:
…ctions

Summary

  • Fixes a false positive in RemoveUnusedPrivateMethods where a private method was removed even though it was still invoked.
  • Adds relaxed method matching via referencesMethod() so call sites using raw types (e.g. raw Map → Set / Collection) still match
    parameterized private method signatures (Set<String>, Collection<String>).
  • Removes the fragile Generic{ string-based workaround in favor of consistent type-based matching.

  Problem

  When a private method is called with values from a raw Map (templates.keySet(), templates.values()), strict JavaType.Method#equals between
  the declaration and TypesInUse.getUsedMethods() fails. The recipe treated the method as unused, deleted it, and left a non-compiling call
  site.

  - Fixes openrewrite/rewrite-static-analysis#877 (https://github.com/openrewrite/rewrite-static-analysis/issues/877).

  Solution

  Introduce referencesMethod(declaration, used) that considers a method used when:

  • Names match
  • Declaring class matches
  • Parameter arity matches
  • Each parameter is compatible via bidirectional TypeUtils.isOfType (handles raw vs parameterized mismatches)

  Test plan

  ◼ RemoveUnusedPrivateMethodsTest.privateMethodCalledWithRawCollectionArguments (issue reproducer)
  ◼ RemoveUnusedPrivateMethodsTest.privateMethodCalledWithParameterizedMapArguments
  - ◼ Existing RemoveUnusedPrivateMethodsTest cases (unused removal, serialization, @MethodSource, @SuppressWarnings, bounded generics #1536)

### Checklist


timtebeek:
Thanks for the work here so far! I've marked it as draft for now as I first want to explore if there's an upstream fix we can make to TypesInUse instead, before making accommodations here downstream. Ideally that would remove the need for adjustments on this PR.

src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java
```diff
@@ -102,14 +102,6 @@ private boolean unusedWarningsSuppressed(J classDeclaration) {
                         }
                     }
 
-                    // Temporary stop-gap until we have data flow analysis.
-                    // Do not remove method declarations with generic types since the method invocation in `cu.getTypesInUse` will be bounded with a type.
-                    for (JavaType.Method usedMethodType : cu.getTypesInUse().getDeclaredMethods()) {
-                        if (methodType.getName().equals(usedMethodType.getName()) && methodType.equals(usedMethodType) && m.toString().contains("Generic{")) {
-                            return m;
-                        }
-                    }
```
timtebeek, line 111: There's performance implications to dropping the precomputed `TypesInUse` here, so I want to explore an upstream fix first if possible.

timtebeek (changes requested):
Nothing immediately expected from your side, but we'll explore the alternative before we circle back here.

timtebeek:
Thanks again for the kick start here & tests; I've added a method upstream to move the type checks there:
- https://github.com/openrewrite/rewrite/pull/8035

I've made the corresponding change here for a more minimal fix.

timtebeek (approved):

PR #918, opened 2026-06-12
`InstanceOfPatternMatch` not to remove double primitive cast

### LINKED ISSUE #916: Double-Cast-Unboxing bug for InstanceOfPatternMatch
does not matter I'll create a test showing the issue on the latest state

With Moderne

I'll make a PR with a reproducing test - see #917 , and check line 2120 in the test ;-) 

That the double-cast is preserved

## What did you see instead?
a compilation error due to missing cast

greg-at-moderne:
## What's changed?

An amendment to `InstanceOfPatternMatch` recipe.
Make it not remove the double cast from `long` to `int`.

## What's your motivation?

- Otherwise it produces invalid code.
- Closes #916 
- Closes #917


MBoegers (approved):
same path I was thinking of. At least until next LST https://openjdk.org/jeps/532 ;)

aamotharald:
@MBoegers @greg-at-moderne ,
so this is for all double-casts of primitives, not only the long to int downcast, right?
SO even for cases where the double cast could be simplified for primitives, it would also not hurt as well to keep the double-cast.

MBoegers:
Yes from the code it only checks for primitive, should also work for `(short) (int) int1`
Agree we can add a second test case for this to prove.

PR #923, opened 2026-06-24
Add UseStringCaseInsensitiveOrder

steve-aom-elliott:
## Summary

Adds a Refaster recipe that replaces case-insensitive string comparator lambdas and method references with the JDK constant `String.CASE_INSENSITIVE_ORDER`:

```java
// before
list.sort((a, b) -> a.compareToIgnoreCase(b));
list.sort(String::compareToIgnoreCase);

// after
list.sort(String.CASE_INSENSITIVE_ORDER);
list.sort(String.CASE_INSENSITIVE_ORDER);
```

Improves readability and avoids one closure allocation per call site.

## Scope

Both transformations are scoped to `Comparator<String>` contexts via Refaster's type inference. The same lambda shape assigned to e.g. `BiFunction<String, String, Integer>` is correctly left alone, as is the inner lambda inside wrapping factories like `Comparator.nullsFirst((a, b) -> a.compareToIgnoreCase(b))` — those get rewritten in place to `Comparator.nullsFirst(String.CASE_INSENSITIVE_ORDER)`.

The following shapes are deliberately NOT matched, because the rewrite would change semantics:

- Reverse-order: `(a, b) -> b.compareToIgnoreCase(a)` would silently invert the sort.
- Locale-aware: `(a, b) -> a.toLowerCase().compareTo(b.toLowerCase())` is not semantically equivalent for non-ASCII case folding (e.g. Turkish dotless-i).

## Test plan

- [x] `lambdaRewrites` — canonical case rewrites correctly (`@DocumentExample`).
- [x] `methodReferenceRewrites` — `String::compareToIgnoreCase` form also rewrites.
- [x] `reverseOrderUnchanged` — reversed arg order skipped.
- [x] `toLowerCaseCompareUnchanged` — locale-aware form skipped.
- [x] `nonComparatorContextUnchanged` — same lambda shape in a `BiFunction` context skipped.
- [x] `wrappedInNullsFirst` — inner lambda still matched inside `Comparator.nullsFirst(...)`.
- [x] `./gradlew test` passes (full suite).
- [x] `./gradlew recipeCsvValidate` passes.

The CSV regenerated as part of this PR — main was already out of sync with `recipeCsvGenerate`'s canonical output, so the diff includes ordering churn beyond the three new rows for this recipe.


timtebeek (approved):
Nice and minimal, thanks! Should the generated recipe be included in any larger composite?

PR #925, opened 2026-06-29
Fix `RemoveRedundantTypeCast` removing cast that gives a `var` lambda its target type

### LINKED ISSUE #924: RemoveRedundantTypeCast removes the cast that gives a lambda its target type when assigned to `var`
**What version of OpenRewrite are you using?**

- rewrite-static-analysis: 2.37.0
- rewrite-core: 8.85.7
- rewrite-java: 8.85.7

**What is the smallest, simplest way to reproduce the problem?**

```java
import static org.openrewrite.java.Assertions.javaVersion;

import org.openrewrite.staticanalysis.RemoveRedundantTypeCast;
import org.openrewrite.test.RewriteTest;

class RemoveRedundantTypeCastTest implements RewriteTest {

  @Test
  void keepsCastForLambdaAssignedToVar() {
    rewriteRun(
        spec -> spec.recipe(new RemoveRedundantTypeCast()),
        // language=java
        java(
            """
            import java.util.function.Function;

            class Example {
              void run() {
                var converter = (Function<Integer, String>) (i -> Integer.toString(i));
              }
            }
            """,
            src -> src.markers(javaVersion(25))));
  }
}
```

**What did you expect to see?**

No change. The cast is not redundant. Because the local variable is declared with `var`, the cast to `Function<Integer, String>` is the only thing that supplies a target type for the lambda; removing it leaves a stand-alone lambda whose type cannot be inferred, so the result does not compile.

**What did you see instead?**

The test fails — the recipe removes the cast, producing code that does not compile:

```
expected:
  var converter = (Function<Integer, String>) (i -> Integer.toString(i));
 but was:
  var converter = (i -> Integer.toString(i));
```

The downstream Java compiler then reports:

```
error: cannot infer type for local variable converter
    var converter = (i -> Integer.toString(i));
        ^
  (lambda expression needs an explicit target-type)
```

- This looks like the same class of defect as the existing "removes necessary cast" reports (#874, #899, #901, #914): `RemoveRedundantTypeCast` should treat the cast as required (skip the removal) when the enclosing variable is declared with `var` and the cast operand is a lambda expression. The recipe documentation already notes that it "does not currently check casts in lambdas".

Note: the analogous method-reference form, `var supplier = (Supplier<String>) Example::make;`, is **not** affected — the recipe correctly preserves that cast — so this is specific to lambda expressions.

timtebeek:
- Fixes #924

`RemoveRedundantTypeCast` already guards against removing a cast whose operand is a lambda or method reference, but the check only matched when the operand was the *direct* cast expression. When the lambda is wrapped in parentheses — `var converter = (Function<Integer, String>) (i -> Integer.toString(i));` — the operand is a `J.ControlParentheses`/`J.Parentheses` node, so the guard missed it and removed the cast.

For a `var` local the cast is the only thing supplying the lambda's target type, so removing it produces code that does not compile (`cannot infer type for local variable`).

This unwraps parentheses before the lambda / method-reference check. The analogous method-reference form was already preserved (its operand isn't parenthesized); a regression test is added for it too.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveRedundantTypeCast.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveRedundantTypeCast.java
@@ -160,8 +160,16 @@ public class RemoveRedundantTypeCast extends Recipe {
                 if ((targetType instanceof JavaType.Primitive || castType instanceof JavaType.Primitive) && castType != expressionType) {
                     return visitedTypeCast;
                 }
-                if (typeCast.getExpression() instanceof J.Lambda || typeCast.getExpression() instanceof J.MemberReference) {
+                J castExpression = typeCast.getExpression();
+                while (castExpression instanceof J.Parentheses || castExpression instanceof J.ControlParentheses) {
+                    castExpression = castExpression instanceof J.Parentheses ?
+                            ((J.Parentheses<?>) castExpression).getTree() :
+                            ((J.ControlParentheses<?>) castExpression).getTree();
+                }
+                if (castExpression instanceof J.Lambda || castExpression instanceof J.MemberReference) {
                     // Not currently supported, this will be more accurate with dataflow analysis.
+                    // The cast supplies the target type for the lambda or method reference; removing it
+                    // can break compilation, e.g. when assigned to a `var` local.
                     return visitedTypeCast;
                 }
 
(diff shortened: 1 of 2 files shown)


protocol7:
Thanks!

timtebeek:
Thank you for the report!

PR #926, opened 2026-06-29
Add RemoveUnreachableMultiCatchAlternative

steve-aom-elliott:
## Summary

Adds `RemoveUnreachableMultiCatchAlternative`, which removes `catch` alternatives that are already covered by an earlier `catch` clause in the same `try`:

- `catch (X) {} catch (X | Y) {}` → `catch (X) {} catch (Y) {}`
- `catch (X) {} catch (Y | X) {}` → `catch (X) {} catch (Y) {}`
- `catch (IOException) {} catch (FileNotFoundException | IllegalStateException) {}` → `catch (IOException) {} catch (IllegalStateException) {}` (subtype shadowed)
- `catch (IOException) {} catch (FileNotFoundException | EOFException) {}` → `catch (IOException) {}` (whole later catch dropped)

When a multi-catch reduces to a single alternative it collapses back to a regular `catch`; when all alternatives are shadowed the entire `catch` clause is removed.

## Motivation

The pattern shows up after type-substitution migrations that cause two `catch` clauses to overlap — for example, when a recipe renames an exception so its parent and child end up in adjacent catches. The result is a Java compile error (`exception X has already been caught`). This recipe cleans that up by dropping the unreachable alternative.

Intended to run **before** `CombineSemanticallyEqualCatchBlocks` so collapsed singletons become candidates for combination.

## Notes

- Gated on `NoMissingTypes` — won't fire when type attribution is incomplete.
- No RSPEC tag attached: I couldn't find a Sonar rule that specifically targets cross-clause shadowing (Sonar likely treats it as a javac error). Happy to add one if reviewers know of a match.

## Test plan

- [x] `dropAlternativeShadowedByEarlierCatch` — `catch (X) ... catch (X | Y)` (`@DocumentExample`)
- [x] `dropTrailingAlternativeShadowedByEarlierCatch` — `catch (X) ... catch (Y | X)`
- [x] `dropMiddleAlternativeShadowedByEarlierCatch` — `catch (X) ... catch (Y | X | Z)`
- [x] `dropSubtypeAlternativeShadowedByEarlierCatch` — alternative is a subtype
- [x] `dropEntireCatchWhenAllAlternativesAreShadowed`
- [x] `dropEntireSingleTypeCatchWhenShadowed`
- [x] `doNothingWhenOrderingMeansNothingIsShadowed`
- [x] `doNothingWhenMultiCatchAlternativesAreAllNeeded`


knutwannheden:
When the recipe drops an entire `catch` clause (the fully-shadowed multi-catch and the shadowed single-type catch paths), the exception types' imports can be left behind unused. Calling `maybeRemoveImport` for the removed types would tidy that up. Minor — not a blocker.

greg-at-moderne (approved):

timtebeek (approved):

PR #943, opened 2026-07-16
FixStringFormatExpressions: don't strip args with no matched specifiers

neil-mushell:
## What's changed?                                                                                                                                                                                         
                                                                                                                                                                                                             
  `FixStringFormatExpressions` now skips its arg-trimming step entirely                                                                                                                                      
  whenever zero `%` conversion specifiers matched in the format string —                                                                                                                                     
  not just when the string happens to contain `{}` tokens. Previously it                                                                                                                                     
  deleted every trailing argument in that case, treating them as dead                                                                                                                                        
  code. Also fixes a related counting bug: `%n` and `%%` don't consume                                                                                                                                       
  an argument in `java.util.Formatter`, but the existing loop counted                                                                                                                                        
  them as if they did.                                                                                                                                                                                       
                                                                                                                                                                                                             
  ## What's your motivation?                                                                                                                                                                                 
                                                                                                                                                                                                             
  Deleting arguments from a format string with zero real specifiers is                                                                                                                                       
  only safe if you're sure they're actually unused. That's not a safe                                                                                                                                        
  assumption when the string uses a different templating convention                                                                                                                                          
  entirely — most commonly SLF4J-style `{}` placeholders mistakenly used                                                                                                                                     
  with `String#format` instead of a logger:                                                                                                                                                                  
                                                                                                                                                                                                             
      String.format("expect {}, actual {}", original, shadow)                                                                                                                                                
      → String.format("expect {}, actual {}")   // args silently dropped                                                                                                                                     
                                                                                                                                                                                                             
  The calling code already had a bug (wrong placeholder syntax for                                                                                                                                           
  `String#format`), but this recipe made it worse: it doesn't fix the                                                                                                                                        
  real problem, and it deletes the evidence (the arguments) that would                                                                                                                                       
  help a reviewer notice something's off. Real-world case that surfaced                                                                                                                                      
  this: a shadow-traffic mismatch reporter whose diagnostic messages all                                                                                                                                     
  degraded to the literal string `"expect {}, actual {}"`, with no                                                                                                                                           
  values, after this recipe ran on it.                   
  
  Two things beyond the headline fix:                                                                                                                                                                        
                                                                                                                                                                                                             
  1. The `%n`/`%%` counting fix — without it, a format string mixing                                                                                                                                         
     `%n` (e.g. from this recipe's own newline replacement) with a                                                                                                                                           
     zero-specifier string would be miscounted as having a real                                                                                                                                              
     specifier on a *later* recipe cycle, silently defeating the main                                                                                                                                        
     fix. Caught by the `newlineStillReplacedWhenArgsAreNotStripped`                                                                                                                                         
     test.                                                                                                                                                                                                   
  2. As a side effect, this also resolves a separate latent bug in the                                                                                                                                       
     `.formatted()` path: the trim step's `i == 0` keep-clause assumed                                                                                                                                       
     index 0 of the argument list is always the format string, true for                                                                                                                                      
     `String.format(fmt, args...)` but false for                                                                                                                                                             
     `"fmt".formatted(args...)` (where index 0 is the first real                                                                                                                                             
     argument) — that previously forced one argument to survive even                                                                                                                                         
     when zero should remain. That branch is only reachable when zero                                                                                                                                        
     specifiers matched, which is now always skipped, so no separate                                                                                                                                         
     fix was needed. Covered by                                                                                                                                                                              
     `doNotStripArgsWithNoSpecifiersEvenWithoutPlaceholders`.                                                                                                                                                
                                                                                                                                                                                                             
  No specific request — open to whoever's around.               
  
  
  Yes, two:                                                                                                                                                                                                  
                                                                                                                                                                                                             
  - A narrower fix that only special-cased `{}`-style tokens specifically                                                                                                                                    
    (detect SLF4J placeholder syntax via a dedicated regex), rather than                                                                                                                                     
    bailing out on *any* zero-specifier case. Went with the broader                                                                                                                                          
    version instead: it's simpler (no new pattern needed), and it's                                                                                                                                          
    robust to other broken-templating conventions we haven't thought of                                                                                                                                      
    (e.g. `MessageFormat`-style `{0}`), not just SLF4J. The tradeoff is                                                                                                                                      
    it also stops trimming genuinely dead args when there's no                                                                                                                                               
    placeholder-like syntax at all (covered by                                                                                                                                                               
    `doNotStripArgsWithNoSpecifiersEvenWithoutPlaceholders`) — a                                                                                                                                             
    reasonable cost for not having to guess.                                                                                                                                                                 
  - Disabling the whole recipe as a workaround downstream, which is what                                                                                                                                     
    prompted finding this in the first place. Fixing it here is                                                                                                                                              
    obviously preferable.                                                                                                                                                                                    
                                                                                                                                                                                                             
  ## Any additional context                                                                                                                                                                                  
                                                                                                                                                                                                             
  Full module suite: 2145 tests, 0 regressions from this change (1                                                                                                                                           
  pre-existing failure in                                                                                                                                                                                    
  `AllBranchesIdenticalTest.collapseIdenticalBranchesPython`, a local                                                                                                                                        
  Python-toolchain environment issue unrelated to this change).                                                                                                                                              
                                                                                                                                                                                                             
  ### Checklist                                                                                                                                                                                              
  - [x] I've read and applied the recipe conventions and best practices                                                                                                                                      

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/FixStringFormatExpressions.java
+++ b/src/main/java/org/openrewrite/staticanalysis/FixStringFormatExpressions.java
@@ -39,6 +39,10 @@ public class FixStringFormatExpressions extends Recipe {
     private static final Pattern FS_PATTERN = Pattern.compile("%(\\d+\\$)?([-#+ 0,(<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])");
     private static final Pattern NEWLINE_PATTERN = Pattern.compile("(?<!\\\\)\\n");
     private static final Pattern ESCAPED_NEWLINE_PATTERN = Pattern.compile("(?<!\\\\)\\\\n");
+    // Matches SLF4J-style placeholder syntax (`{}`), as opposed to a `String#format` conversion
+    // specifier. Used to detect when a format string with zero real specifiers is more likely using
+    // the wrong templating convention than genuinely containing dead arguments -- see below.
+    private static final Pattern UNRESOLVED_PLACEHOLDER_PATTERN = Pattern.compile("\\{}");
 
     private static final MethodMatcher FORMAT_MATCHER = new MethodMatcher("java.lang.String format(..)");
     private static final MethodMatcher FORMATTED_MATCHER = new MethodMatcher("java.lang.String formatted(..)");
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks a lot for the help here! I've trimmed the contribution a bit, but other than that it looks good to me.

PR #988, opened 2026-08-11
Keep annotated and `strictfp` overrides that only call super

martinfrancois:
**Suggested review order:** 8 of 52 (Score: 8)
**Review first:** https://github.com/openrewrite/rewrite-migrate-java/pull/1205

## What's changed?

Updates `RemoveMethodsOnlyCallSuper` to preserve forwarding overrides when a parameter or any nested part of the method signature carries an annotation other than `@Override`.

The change also preserves a `strictfp` override when the invoked super method is not `strictfp`. It still removes the override when both methods are `strictfp`, matching the recipe's existing handling of `synchronized`.

The three former failing reproductions are now passing regression tests, and a positive control verifies the safe `strictfp` removal. All 20 tests in `RemoveMethodsOnlyCallSuperTest` pass.

## What's your motivation?

**Recipe:** [`org.openrewrite.staticanalysis.RemoveMethodsOnlyCallSuper`](https://docs.openrewrite.org/recipes/staticanalysis/removemethodsonlycallsuper).

#971 taught this recipe to keep overrides that add declaration annotations or the `synchronized` modifier. Its method-level annotation query does not include annotations on parameters or type-use annotations nested inside the return type, so the recipe still deletes overrides that carry those contracts.

#### Before

```java
class Child extends Parent {
    @Override
    void foo(@Nullable String s) {
        super.foo(s);
    }
}
```

#### Actual after the recipe

```java
class Child extends Parent {
}
```

#### Expected after the recipe

`(unchanged)`

Removing the method silently removes the parameter contract read by annotation processors and static-analysis tools. The same loss occurs for a nested type-use annotation such as `public String @Nullable [] foo()`.

#### Before

```java
class Child extends Parent {
    @Override
    strictfp void foo() {
        super.foo();
    }
}
```

#### Actual after the recipe

The override is deleted, leaving the same empty `Child` class shown in Case 1.

#### Expected after the recipe

`(unchanged)` when the super method is not `strictfp`.

The previous transformation drops a reflection-visible modifier. Before Java 17, it also changes floating-point semantics. When the super method is already `strictfp`, removing this forwarding override remains safe and the recipe still removes it.

### Confirmed real-world execution

- Project: [Hibernate Validator](https://github.com/hibernate/hibernate-validator) (1,267 stars on 2026-08-16).
- Location: [`MethodOverridingTests.java` at `e1f271a7`](https://github.com/hibernate/hibernate-validator/blob/e1f271a7ad429690fd18b0f4cbb7ea40dd8af575/annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/overriding/MethodOverridingTests.java) and its [integration test](https://github.com/hibernate/hibernate-validator/blob/e1f271a7ad429690fd18b0f4cbb7ea40dd8af575/annotation-processor/src/test/java/org/hibernate/validator/ap/ConstraintValidationProcessorIT.java).
- Recipe release: `org.openrewrite.recipe:rewrite-static-analysis:2.41.0`.

The released recipe deletes a delegating override whose parameter annotations intentionally create the invalid constraint declaration tested by Hibernate Validator. The deletion removes the diagnostic condition and weakens the test.

### Affected code in real projects

- [`didi/DoKit` `CpuMainPageFragment.java`](https://github.com/didi/DoKit/blob/626827cddb2feb2f3aee87a52a064b4e5ca2bed4/Android/dokit/src/main/java/com/didichuxing/doraemonkit/kit/parameter/cpu/CpuMainPageFragment.java#L24-L28): an `onViewCreated` override that only forwards to super but carries `@NonNull` and `@Nullable` parameter annotations. The recipe from main deletes the whole override, silently dropping the parameter nullability contracts it declares.
- [`Justson/AgentWeb` `AgentActionFragment.java`](https://github.com/Justson/AgentWeb/blob/e683a5c31acdd17a86c3e92b4b3790770ec016c3/agentweb-core/src/main/java/com/just/agentweb/AgentActionFragment.java#L88-L91): the same shape in a library fragment, `super.onViewCreated(view, savedInstanceState)` with `@NonNull`/`@Nullable` annotated parameters. The recipe from main removes the method together with those annotations.

Please review the conservative signature traversal, which intentionally stops before the method body, and the decision to mirror the existing `synchronized` parent-method comparison for `strictfp`.

One alternative is to enumerate specific signature locations. Traversing the complete signature is safer because an override carrying annotation metadata is not redundant, regardless of whether the annotation is attached to a parameter, type parameter, return type, or nested type use.

For `strictfp`, preserving every marked override would also retain behavior. Comparing the super method retains the recipe's cleanup when the modifier is redundant.

## Any additional context

Pre-existing tests changed: None.

Related: #971 added the `synchronized` guard and the declaration-level annotation guard. This change extends the same behavior to the remaining signature locations and to `strictfp`.

This change was prepared with AI assistance. I reviewed the implementation, regression tests, focused test results, and contribution text.

### Checklist

- [x] I've run `./gradlew build` locally, and committed any resulting changes to `recipes.csv`
- [x] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/RemoveMethodsOnlyCallSuper.java
+++ b/src/main/java/org/openrewrite/staticanalysis/RemoveMethodsOnlyCallSuper.java
@@ -23,6 +23,7 @@ import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.AnnotationMatcher;
+import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.staticanalysis.kotlin.KotlinFileChecker;
@@ -135,10 +136,9 @@ public class RemoveMethodsOnlyCallSuper extends Recipe {
             }
 
             private boolean hasSemanticAnnotation(J.MethodDeclaration method) {
-                AtomicBoolean found = new AtomicBoolean();
-                new JavaVisitor<AtomicBoolean>() {
+                return new JavaIsoVisitor<AtomicBoolean>() {
                     @Override
-                    public J.Annotation visitAnnotation(J.Annotation annotation, AtomicBoolean ignored) {
+                    public J.Annotation visitAnnotation(J.Annotation annotation, AtomicBoolean found) {
                         if (!OVERRIDE.matches(annotation)) {
                             found.set(true);
                         }
@@ -146,11 +146,10 @@ public class RemoveMethodsOnlyCallSuper extends Recipe {
                     }
 
                     @Override
-                    public J.Block visitBlock(J.Block block, AtomicBoolean ignored) {
+                    public J.Block visitBlock(J.Block block, AtomicBoolean found) {
                         return block;
                     }
-                }.visit(method, found);
-                return found.get();
+                }.reduce(method, new AtomicBoolean()).get();
             }
 
             private boolean argumentsMatchParameters(List<Statement> parameters, List<Expression> arguments) {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the double fixes here!

PR #997, opened 2026-08-12
Fix crashes on Go `for` loops and `select` statements

knutwannheden:
## Motivation

Three recipes crash outright when they visit LSTs produced by the Go parser. Both `for` loop recipes index into the loop control's init and update lists, which Go leaves genuinely empty — Java's parser always puts a `J.Empty` placeholder there, so `get(0)` looked safe. `MinimumSwitchCases` casts every clause of a switch to `J.Case`, which holds for Java but not for Go's `select` statement, whose clauses are `Go.CommClause`.

These showed up as `SourcesFileErrors` on real repositories:

```
java.lang.IndexOutOfBoundsException: Index: 0
  org.openrewrite.staticanalysis.ForLoopIncrementInUpdate$1.visitForLoop(ForLoopIncrementInUpdate.java:59)

java.lang.IndexOutOfBoundsException: Index: 0
  org.openrewrite.staticanalysis.WhileInsteadOfFor$1.visitForLoop(WhileInsteadOfFor.java:51)

java.lang.ClassCastException: class org.openrewrite.golang.tree.Go$CommClause cannot be cast to class org.openrewrite.java.tree.J$Case
  org.openrewrite.staticanalysis.MinimumSwitchCases$1.lambda$doRewrite$2(MinimumSwitchCases.java:235)
```

In each case skipping is the right outcome, not just the safe one: a Go `for cond { }` is already a `while` loop, and Go has no `while` keyword to convert it to.

## Summary

- `ForLoopIncrementInUpdate` checks that the init list is non-empty before reading its first element.
- `WhileInsteadOfFor` checks that the init and update lists are non-empty before reading their first elements, so a loop that has no init/update elements at all is left alone.
- `MinimumSwitchCases.doRewrite` returns `false` when the switch holds a clause that is not a `J.Case`, which also protects the `J.Case[]` collection further down `visitSwitch`.

## Test plan

- [x] Added a regression test per recipe. No parser on the test classpath produces these LST shapes, so each test parses Java and then reshapes the result — emptying the `for` control's init/update lists, or replacing the switch clauses with statements that are not `J.Case` — and asserts the recipe leaves the tree untouched instead of throwing.
- [x] Confirmed each new test fails with exactly the production exception above before its fix, and passes after.
- [x] `./gradlew test` — the only failures are the JavaScript/Python RPC tests (`RPC process shut down early with exit code 1`), which fail identically on an untouched tree in this environment and are unrelated to these recipes.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ForLoopIncrementInUpdate.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ForLoopIncrementInUpdate.java
@@ -56,9 +56,10 @@ public class ForLoopIncrementInUpdate extends Recipe {
         return new JavaVisitor<ExecutionContext>() {
             @Override
             public J visitForLoop(J.ForLoop forLoop, ExecutionContext ctx) {
-                Statement init = forLoop.getControl().getInit().get(0);
-                if (init instanceof J.VariableDeclarations) {
-                    J.VariableDeclarations initVars = (J.VariableDeclarations) init;
+                // Go for loops have no init element at all, where Java has a J.Empty placeholder
+                List<Statement> init = forLoop.getControl().getInit();
+                if (!init.isEmpty() && init.get(0) instanceof J.VariableDeclarations) {
+                    J.VariableDeclarations initVars = (J.VariableDeclarations) init.get(0);
 
                     Statement body = forLoop.getBody();
                     Statement lastStatement;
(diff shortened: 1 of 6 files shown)


timtebeek:
Closed for the same reason as https://github.com/openrewrite/rewrite-static-analysis/pull/986 : we want to address this in the parser instead.

PR #998, opened 2026-08-12
UnnecessaryExplicitTypeArguments: retain witness when enclosing method has dependent type parameters

### LINKED ISSUE #783: `UnnecessaryExplicitTypeArguments` incorrectly removes type parameter needed for dependent type parameters.
The `UnnecessaryExplicitTypeArguments` recipe incorrectly removed an explicit type parameter in a location where it was necessary (specifically, a method with type parameters where one type parameter extends another).

- This is similar but distinct from the previously fixed issue #164 "Unnecessary explicit type arguments removes necessary type arguments." Here type arguments needed for methods with dependent type parameters (e.g., `<T, S extends T>`).

## Minimal Reproduction

```java
import java.util.Comparator;
import java.util.List;

public class MinimalReproOpenRewriteBug {

    /**
     * A method with stricter type bounds than standard Java APIs.
     * The type parameter S extends T, which requires explicit type specification
     * when T cannot be inferred from the Comparator argument alone.
     */
    public static <T, S extends T> Comparator<Iterable<S>> lexicographical(Comparator<T> comparator) {
        return (list1, list2) -> {
            var it1 = list1.iterator();
            var it2 = list2.iterator();
            while (it1.hasNext() && it2.hasNext()) {
                int cmp = comparator.compare(it1.next(), it2.next());
                if (cmp != 0) {
                    return cmp;
                }
            }
            return Boolean.compare(it1.hasNext(), it2.hasNext());
        };
    }

    // This works - explicit type parameter
    private static final Comparator<Iterable<Integer>> WORKS =
        lexicographical(Comparator.<Integer>naturalOrder());

    // This fails to compile - type parameter removed by OpenRewrite
    private static final Comparator<Iterable<Integer>> FAILS =
        lexicographical(Comparator.naturalOrder());

    record Example(List<Integer> numbers) implements Comparable<Example> {
        private static final Comparator<Example> COMPARATOR =
            Comparator.comparing(Example::numbers, WORKS);

        @Override
        public int compareTo(Example other) {
            return COMPARATOR.compare(this, other);
        }
    }
}
```

OpenRewrite should preserve the explicit type parameters here. What actually happened it changed:

```java
Comparator.<Integer>naturalOrder()
```
to:
```java
Comparator.naturalOrder()
```

This causes the compilation error:

```
MinimalReproOpenRewriteBug.java:50: error: method lexicographical in class MinimalReproOpenRewriteBug cannot be applied to given types;
        lexicographical(Comparator.naturalOrder());
        ^
  - required: Comparator<T#1>
  - found:    Comparator<T#2>
  - reason: inference variable S has incompatible upper bounds T#3,Comparable<? super T#3>,Object,T#1
  - where T#1,S,T#2,T#3 are type-variables:
    - T#1 extends Object declared in method <T#1,S>lexicographical(Comparator<T#1>)
    - S extends T#1 declared in method <T#1,S>lexicographical(Comparator<T#1>)
    - T#2 extends Comparable<? super T#2>
    - T#3 extends Comparable<? super T#3> declared in method <T#3>naturalOrder()
1 error
```

Niloyyy:
- Fixes #783

## What I found

- The reproduction as filed in #783 no longer reproduces on `main`. `Comparator.naturalOrder()` takes no arguments, so it is already caught by the no-argument check in `canInferTypeArgumentsFromArguments`, and the witness is retained. That case appears to have been fixed incidentally rather than deliberately, and there was no regression test pinning it, which is likely why the issue stayed open.

The case @timtebeek described in the issue thread is still open though:

> Now we need to figure out how to recognize that case and avoid removing the explicit type arguments when used as an argument to such a method.

Nothing in the recipe looks at any method's declared type-parameter *bounds*. The existing checks only compare which type variables appear in the inner call's own return type against those in its own parameter types. So once the inner call takes arguments, the no-argument check no longer fires and the witness is removed even though the enclosing method's dependent type parameters still need it:

```java
static <T, S extends T> Comparator<Iterable<S>> lexicographical(Comparator<T> comparator) { ... }
static <U> Comparator<U> reverse(Comparator<U> comparator) { ... }

// witness removed, no longer compiles
static final Comparator<Iterable<Integer>> COMPARATOR =
    lexicographical(Test.<Integer>reverse(Comparator.naturalOrder()));
```

Verified against `javac` rather than assumed — with the witness it compiles, without it:

```
error: method reverse in class Test cannot be applied to given types;
  - required: Comparator<U#1>
  - found:    Comparator<T#1>
  reason: inferred type does not conform to equality constraint(s)
    - inferred: T#2
    - equality constraints(s): U#2
```

## The change

In the argument-position branch, retain the witness when the enclosing method's declared signature has a type parameter whose bound references a **different** type parameter of that same method. Inference resolves those jointly against the target type, so the witness can be load-bearing regardless of what the inner call's own arguments pin down.

Self-referential F-bounds such as `<E extends Enum<E>>` and `<T extends Comparable<T>>` are deliberately excluded. They express no dependency on another type parameter and are common enough that counting them would cause widespread over-retention.

`collectGenericTypeVariableNames` is generalised to `collectGenericTypeVariables`, collecting into a map so the bounds are reachable, rather than adding a second parallel walker. `findDeclaredSignature` is reused as-is.

## Tests

- Two tests added, both tagged to #783: the original reproduction as a regression guard, and the variant that does still reproduce.

Full suite run locally on JDK 21: 2242 tests, 0 failures, 0 errors. In particular `removesWitnessOnInstanceMethodArgumentWhenInferableFromOwnArguments` still passes, so witnesses that genuinely are redundant are still removed, and `retainsWitnessOnInstanceMethodPassedAsArgumentToGenericMethod` covers the F-bound exclusion.

Happy to adjust the naming or fold the check in differently if you'd prefer it structured another way.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
+++ b/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java
@@ -22,8 +22,10 @@ import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.staticanalysis.java.JavaFileChecker;
 
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 public class UnnecessaryExplicitTypeArguments extends Recipe {
@@ -70,6 +72,13 @@ public class UnnecessaryExplicitTypeArguments extends Recipe {
                         }
                         // Cannot remove type parameters if it would introduce ambiguity about which method should be called
                         if (enclosingMethod.getMethodType() == null) {
+                            return m;
+                        }
+                        // The enclosing method's type parameters may be interdependent (e.g. `<T, S extends T>`).
+                        // Inference then resolves them jointly against the target type, so this argument's witness
+                        // can be load-bearing even though this call's own type variables are inferable from its
+                        // arguments. Retain it rather than reason about the enclosing method's inference.
+                        if (hasInterdependentTypeParameters(enclosingMethod.getMethodType())) {
                             return m;
                         }
                         if (!(enclosingMethod.getMethodType().getDeclaringType() instanceof JavaType.Class)) {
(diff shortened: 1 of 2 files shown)


sambsnyd (approved):

sambsnyd:
Thanks @Niloyyy !

motlin:
Thank you very much!

PR #999, opened 2026-08-13
WhileInsteadOfFor not to run for Go code

greg-at-moderne:
## What's changed?

Making the `WhileInsteadOfFor` recipe not run against Go code.

## What's your motivation?

There's no `while` loop in Go. So the recipe would produce invalid code.


timtebeek:
Confirming this from the flagship runs, with a note on the timing in case anyone else chases it.

The crash this guards against was still landing on `dev` as of the 2026-08-20 03:28 UTC run of *Common static analysis issues* against org "Netflix + Spring" (changeset `20260820032852-e8B2R`), on `Netflix/bettertls`:

```
java.lang.NullPointerException: Cannot invoke "org.openrewrite.java.tree.JRightPadded.getElement()" because "this.condition" is null
  org.openrewrite.java.tree.J$ForLoop$Control.getCondition(J.java:2486)
  org.openrewrite.staticanalysis.WhileInsteadOfFor$1.visitForLoop(WhileInsteadOfFor.java:53)
  org.openrewrite.staticanalysis.WhileInsteadOfFor_1_GolangVisitor.visitForLoop(WhileInsteadOfFor_1_GolangVisitor.zig:100)
```

That is *not* a gap in this PR — it is deployment lag. The bundle carrying it, `org.openrewrite.recipe:rewrite-static-analysis:2.42.0-20260819.173457-20`, only finished installing on that tenant at 2026-08-20 07:44:01 UTC, four hours after the run started. (Worth flagging for anyone reading deployment state off the version string: the `20260819.173457` is the upstream jar build time, not when the platform picked it up.) Source file for the record: `test-suites/impltests/envoy.go`.

For completeness, the underlying LST shape is unchanged and outlives this guard: `J.ForLoop.Control#condition` is left nil for Go's bare `for { }`, where the Java parser inserts a `J.Empty`. openrewrite/rewrite#8479 covered `init` and `update` but not `condition`. I've added that, plus a `Go.CommClause` / `J.Case` cast that still hits `MinimumSwitchCases`, to openrewrite/rewrite#8461.

PR #1000, opened 2026-08-13
Add `RemoveDuplicateAnnotations`

timtebeek:
- Extracted from openrewrite/rewrite-migrate-java#1200, where this started life as a JSpecify-specific recipe before we realised it applies far more widely.

When several distinct annotations migrate to a single new annotation, each `ChangeType` rewrites its own annotation and the element ends up annotated twice:

```java
import javax.annotation.CheckForNull;
import javax.annotation.Nullable;

class Foo {
    public void bar(@Nullable @CheckForNull String baz) {
    }
}
```

- becomes `@Nullable @Nullable String baz` — sometimes with one of the two fully qualified, when the simple name can no longer be shortened. That is a compile error for non-`@Repeatable` annotations. See openrewrite/rewrite-migrate-java#1199 for a report in the wild.

This recipe removes annotations repeated on the same element, keeping the first occurrence. It covers leading annotations on classes, methods and variables, annotations attached to modifiers, and type-use annotations on array types, nested types and type arguments.

Two guards keep it conservative:
- only annotations *semantically equal* to an earlier one are removed, so `@Nullable @NonNull` and `@Tag("a") @Tag("b")` are untouched;
- `@Repeatable` annotations are skipped, since repeating those is meaningful.

`annotationType` is optional and defaults to any annotation, so the recipe is useful standalone; rewrite-migrate-java will call it with `org.jspecify.annotations.*` from its JSpecify migrations.


timtebeek:
CI is red on ten TypeScript tests (`JSON-RPC peer closed the stream` / `RPC process shut down early with exit code 1`). Those are pre-existing on `main` — the same ten fail in run 31728116299 on `main` at 17:55, before this branch, and they fail locally on a clean checkout too. `RemoveDuplicateAnnotationsTest` passes.

Issue #229, opened 2023-12-18
`FallThrough` should not add break when `switch` is last statement in a method requiring a `return`


knutwannheden:
In the following case
```java
  public class A {
      public int m(int i) {
          switch (i) {
              case 0:
                  if (true) return i;
              default:
                  throw new IllegalStateException();
          }
      }
  }
```
the `FallThrough` recipe should not add a `break` statement after the `if`. The resulting code doesn't compile. Instead it would be better to add a `// fall through` comment here.

knutwannheden:
@jetbrains-junie Check if this issue is still applicable or if it can be closed.

PR #1008, opened 2026-08-16
FallThrough: don't add a break after a guarded early exit

### LINKED ISSUE #460: breaking switch logic: `org.openrewrite.staticanalysis.CodeCleanup`: JavadocToAsciidocTransformerConfigItemTest.escape:315
run: https://docs.openrewrite.org/recipes/staticanalysis/codecleanup#usage

in: https://github.com/quarkusio/quarkus/tree/main/core

the added `break` is causing the test so fail:

Image

`JavadocToAsciidocTransformerConfigItemTest.java:315`

Image

@timtebeek

Niloyyy:
- Fixes #460

Per your call in the issue — the recipe should leave this shape alone.

## What was happening

`FallThrough` treated a case ending in an `if` without an `else` as an unintentional fall-through and appended a `break`, which changed behavior:

```java
case ']':
    if (inlineMacroMode) {
        - sb.append("&#93;");
        break;
    }
    break;   // <-- added; `]` is no longer escaped when !inlineMacroMode
case '#':
case '*':
    if (!escaping) {
        sb.append("++");
        escaping = true;
    }
    sb.append(ch);
    break;
```

`breaks(Statement)` returned `false` for an `if` with no `else`, no relief comment was present, so a `break` was appended.

## The change

An `if` without an `else` whose then part completes abruptly is a guarded early exit — the statements after it are reachable only when the guard is false, so a fall-through past it is deliberate rather than an omission.

I added this as a separate `guardsWithEarlyExit` check rather than relaxing `breaks(..)`, because the two mean different things. `breaks(..)` answers "does this always complete abruptly", and it is used recursively for `try`, `switch` and blocks. Loosening it there regressed `nestedBlocks`, where

```java
try {
    if (true) {
        return 1;
    }
} catch (Exception e) {
    ...
}
```

can still complete normally and so genuinely does need the `break`. `guardsWithEarlyExit` therefore unwraps only blocks and labels, and deliberately does not descend into `try` or `switch`.

The trade-off is that the same shape can be a real omission, so some detection is lost. That seemed to be the intent of "not making any change here makes most sense", but happy to narrow it further if you'd rather.

## Tests

- Three added, all tagged to #460: the reduced quarkus case, the block-wrapped variant that reaches the check through the block recursion, and a negative case confirming a `break` is still added when unguarded code follows the guard.

Full suite locally on JDK 21: 2277 tests, 0 failures.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/FallThroughVisitor.java
+++ b/src/main/java/org/openrewrite/staticanalysis/FallThroughVisitor.java
@@ -156,9 +156,32 @@ public class FallThroughVisitor<P> extends JavaIsoVisitor<P> {
                 return trees.stream()
                         .reduce((s1, s2) -> s2) // last statement
                         .map(s -> breaks(s) || // https://github.com/openrewrite/rewrite-static-analysis/issues/173
+                                guardsWithEarlyExit(s) || // https://github.com/openrewrite/rewrite-static-analysis/issues/460
                                 s.getComments().stream().anyMatch(HAS_RELIEF_PATTERN_COMMENT) ||
                                 s instanceof J.Block && ((J.Block) s).getEnd().getComments().stream().anyMatch(HAS_RELIEF_PATTERN_COMMENT)
                         ).orElse(false);
+            }
+
+            /**
+             * An {@code if} without an {@code else} whose then part completes abruptly guards the statements
+             * after it, so a fall-through past it is deliberate rather than an omission and adding a
+             * {@code break} would change behavior. Only blocks and labels are unwrapped; constructs such as
+             * {@code try} and {@code switch} are deliberately not, because they can complete normally even
+             * when every {@code if} they contain completes abruptly.
+             */
+            private static boolean guardsWithEarlyExit(Statement s) {
+                if (s instanceof J.Block) {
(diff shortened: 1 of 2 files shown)


Niloyyy:
While looking at #229 I noticed this change also covers it — it's the same shape, an `if` without an `else` whose then part completes abruptly.

On `main`:

```java
case 0:
    if (true) return i;
    break;      // added
default:
    throw new IllegalStateException();
```

which fails to compile with `error: missing return statement`. On this branch the case is left unchanged and it compiles.

It doesn't do the rest of what #229 asks for, though. The `@ExpectedToFail` `switchAsLastStatement` test (added in 1a65c15) expects a `// fall through` comment to be inserted instead:

```java
case 0:
    if (true) return i;
    // fall through
default:
```

This branch adds no comment, so that test still fails and stays `@ExpectedToFail`.

Would you like the comment insertion as well, either here or as a follow-up? Happy to do it — I just didn't want to assume the recipe should start writing comments without checking first.

Niloyyy:
@timtebeek this is the fix PR from #460.

@knutwannheden the question above is really yours — for #229, would you still want the `// fall through` comment inserted, or is leaving the case unchanged enough?

Issue #87, opened 2023-05-04
ReplaceStackWithDeque sometimes fails with an exception


koppor:
> org.openrewrite.internal.RecipeRunException: Exception while visiting project file 'src\main\java\org\jabref\model\openoffice\ootext\OOTextI
ntoOO.java (in OOTextIntoOO.write)', caused by: org.openrewrite.java.controlflow.ControlFlowIllegalStateException: Unable to create new J.Vari
ableDeclarations with name `pairIterator`

https://github.com/JabRef/jabref/blob/main/src/main/java/org/jabref/model/openoffice/ootext/OOTextIntoOO.java

<details>
<pre>

> org.openrewrite.internal.RecipeRunException: Exception while visiting project file 'src\main\java\org\jabref\model\openoffice\ootext\OOTextI
ntoOO.java (in OOTextIntoOO.write)', caused by: org.openrewrite.java.controlflow.ControlFlowIllegalStateException: Unable to create new J.Vari
ableDeclarations with name `pairIterator`
        AST: J.ForEachLoop(padding=org.openrewrite.java.tree.J$ForEachLoop$Padding@1cdd3384, id=300fe438-9af5-44c7-b3e5-f2efec566c14, prefix=S
pace(comments=<0 comments>, whitespace='\n┬ÀÔéü┬ÀÔéé┬ÀÔéâ┬ÀÔéä┬ÀÔéà┬ÀÔéå┬ÀÔéç┬ÀÔéê┬ÀÔéë┬ÀÔéè┬ÀÔéü┬ÀÔéé┬ÀÔéâ┬ÀÔéä┬ÀÔéà┬ÀÔéå┬ÀÔéç┬ÀÔéê┬ÀÔéë┬ÀÔéè
'), markers=Markers(id=7f8ffa86-0646-4c43-b758-2f899d90e725, markers=[]), control=OOPair<String, String> pair  attributes, body=J.Block(paddin
g=org.openrewrite.java.tree.J$Block$Padding@2034cf68, id=aba06d4c-25e9-45df-884f-f0ed5bb8ca89, prefix=Space(comments=<0 comments>, whitespace=
'┬ÀÔéü'), markers=Markers(id=7f8ffa86-0646-4c43-b758-2f899d90e725, markers=[]), statik=JRightPadded(element=false, after=Space(comments=<0 com
[...]

PR #1030, opened 2026-08-26
Only run data flow analysis on `Stack` variables in `ReplaceStackWithDeque`

timtebeek:
## What's changed

`ReplaceStackWithDeque.visitVariable` ran `FindLocalFlowPaths` for **every** initialized variable in any file that mentions `java.util.Stack`, and only then applied `ChangeType` to the initializer — a no-op for variables that aren't a `Stack`. Swapping the order (apply `ChangeType` first, bail out when it leaves the initializer untouched) is equivalent and confines the analysis to the variables the recipe can actually rewrite.

## Why

This is how the crash surfaced. `org.openrewrite.staticanalysis.ReplaceStackWithDeque` has been failing with `java.lang.StackOverflowError` on `spring-projects/spring-data-mongodb`, file `spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoExampleMapper.java`, recurring since at least 2026-08-17 in the "Java best practices" and "Common static analysis issues" flagship runs on `dev` (`ALL`, `Open Source`, `Netflix + Spring + Apache`, `Netflix + Spring`). Confirmed in run `20260825041417-SSuzc`; absent on 2026-08-26, so it is input/attribution dependent and a clean run isn't evidence of a fix.

The `SourcesFileErrors` data table for that run gives ~1000 frames alternating between exactly two call sites:

```
java.lang.StackOverflowError
  org.openrewrite.Cursor.getParentTreeCursor(Cursor.java:253)
  org.openrewrite.analysis.trait.expr.VarAccessBase.viewOf(VarAccess.java:221)
  ...
  org.openrewrite.analysis.dataflow.DataFlowNode.of(DataFlowNode.java:63)
  org.openrewrite.analysis.dataflow.analysis.ForwardFlow.computeVariableAssignment(ForwardFlow.java:458)
  org.openrewrite.analysis.dataflow.analysis.ForwardFlow.computeVariableAssignment(ForwardFlow.java:495)
  org.openrewrite.analysis.dataflow.analysis.ForwardFlow.computeVariableAssignment(ForwardFlow.java:466)
  org.openrewrite.analysis.dataflow.analysis.ForwardFlow.computeVariableAssignment(ForwardFlow.java:495)
  org.openrewrite.analysis.dataflow.analysis.ForwardFlow.computeVariableAssignment(ForwardFlow.java:466)
  ...
```

- `ForwardFlow.computeVariableAssignment` recurses from an argument into the call's select (`:466`) and from the select back into an argument (`:495`) with no visited set, so when `isFlowStep(..)` holds in both directions it never terminates. Root cause filed as openrewrite/rewrite-analysis#113 with the analysis of which model groups produce the two directions.

The added test reproduces the production stack trace exactly (502 frames at `:466`, 501 at `:495`) on `main`, and passes here. The `Stack` in it only satisfies the recipe's `UsesType` precondition — the overflow comes from the unrelated `String a`, which is also what happens in `MongoExampleMapper`.

## Scope

- This narrows the blast radius; it is not the engine fix. A `Stack` whose flow reaches such an expression can still overflow, and any other recipe using `FindLocalFlowPaths` remains exposed until openrewrite/rewrite-analysis#113 is fixed. A caller can't work around it themselves — `DataFlowSpec.isFlowStep` is `final`, so a spec cannot suppress the external models that create the cycle.

- openrewrite/rewrite-static-analysis#87 names this recipe but is a different defect (`ControlFlowIllegalStateException` out of `ControlFlow.createFakeIteratorVariableDeclarations`, not `ForwardFlow`), so it is not the tracking issue for this and is not closed by this PR.

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/ReplaceStackWithDeque.java
+++ b/src/main/java/org/openrewrite/staticanalysis/ReplaceStackWithDeque.java
@@ -55,6 +55,16 @@ public class ReplaceStackWithDeque extends Recipe {
             @Override
             public J.VariableDeclarations.NamedVariable visitVariable(J.VariableDeclarations.NamedVariable variable, ExecutionContext ctx) {
                 J.VariableDeclarations.NamedVariable v = super.visitVariable(variable, ctx);
+                if (v.getInitializer() == null) {
+                    return v;
+                }
+
+                Expression initializer = (Expression) new ChangeType("java.util.Stack", "java.util.ArrayDeque", false)
+                        .getVisitor().visitNonNull(v.getInitializer(), ctx, getCursor().getParentOrThrow());
+                if (initializer == v.getInitializer()) {
+                    // Not a `Stack`, so skip the data flow analysis below, which is costly on every variable in the file
+                    return v;
+                }
 
                 DataFlowSpec returned = new DataFlowSpec() {
                     @Override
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
Why is this a problem specifically for stacks? This seems like a more general problem, or like we are missing more general recursion detection that would break a recursive check

JLLeitschuh:
This doesn't seem like the right way of solving this problem. This seems like a more general recursion problem where we aren't checking the graph being built correctly.

Can we reconsider how we're fixing this and build something more resilient?

PR #1031, opened 2026-08-27
Don't rewrite MinimumSwitchCases when the selector's type is unresolved

neil-mushell:
## What's changed?                                                                                                                                                                            
                                                                                                                                                                                             
  MinimumSwitchCases no longer rewrites a switch statement into an ==/equals() chain when the selector's own type can't be resolved (e.g. an incomplete classpath). Previously it only bailed  out when a case label's own type was unresolved; now it also bails out when the selector's type is unresolved, since in that situation the case labels' resolved types can't be trusted      either.                                                                                                                                                                                    
                                                                                                                                                                                             
 ## What's your motivation?                                                                                                                                                                    
                                                                                                                                                                                             
  Enum switch case labels are resolved specially, only among the switch's own enum constants, ignoring anything else in scope (JLS 14.11.1). That special lookup no longer applies once the  label is spliced into a plain == expression outside the switch. When the selector's type is unresolved, a case label identifier that otherwise looks resolved may really be a same-named   symbol from elsewhere in scope, such as a statically imported constant, rather than the switch's true value. That produced type-incorrect code like x.getFoo() == UNRELATED_STRING_CONSTANT   instead of the recipe correctly skipping the transform.                                                                                                                                    
                                                                                                                                                                                             
  Worth a close look at whether bailing out on any unresolved selector type is too broad, versus only bailing when there's an actual name collision with something else in scope. I went with  the broader, more conservative check since the narrower one would need reliable visibility into what else is in scope, which is harder to establish soundly from a single-file recipe  pass.                                                                                                                                                                                      
                                                                                                                                                                                             
  Considered only bailing when a case label identifier's name matches a statically imported or otherwise in-scope symbol, but that requires more scope analysis than the recipe currently      does and seemed more likely to have its own gaps.                                                                                                                                          
                                                                                                                                                                                             
 ## Any additional context                                                                                                                                                                     
                                                                                                                                                                                             
  Added MinimumSwitchCasesTest#doNotRewriteWhenSelectorTypeIsUnresolved (no rewrite when selector type is unresolved and case labels shadow an unrelated static import) and                  
  #rewritesWhenSelectorTypeIsResolvedDespiteShadowingStaticImport (same shadowing shape, but with a resolvable selector type, confirming the rewrite still happens correctly) to cover both    the bug and the non-regression case.                                                                                                                                                       
                                                                                                                                                                                             
 ## Checklist                                                                                                                                                                                  
                                                                                                                                                                                             
  - [X] I've read and applied the recipe conventions and best practices                                                                                                                      
  - [X] I've run ./gradlew build locally, and committed any resulting changes to recipes.csv                                                                                                 
  - [x] I've formatted the lines I changed, without reformatting code I didn't touch

### DIFF
--- a/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
+++ b/src/main/java/org/openrewrite/staticanalysis/MinimumSwitchCases.java
@@ -236,6 +236,8 @@ public class MinimumSwitchCases extends Recipe {
             }
 
             private boolean hasUnresolvableIdentifierCasePattern(J.Switch switch_) {
+                // Enum case labels resolve specially (JLS 14.11.1); that lookup doesn't apply once spliced outside the switch, so an unresolved selector type makes any label untrustworthy.
+                boolean selectorTypeUnresolved = isUnresolved(switch_.getSelector().getTree().getType());
                 for (Statement statement : switch_.getCases().getStatements()) {
                     if (statement instanceof J.Case) {
                         J.Case aCase = (J.Case) statement;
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the help here @neil-mushell ! Found a small related issue reviewing, and I've narrowed the type validation just to ensure we do so for the right reasons, which is always tricky with tests that don't fully compile.

timtebeek (approved):

PR #1046, opened 2026-09-01
Add UnnecessaryFinalInTryWithResources

sullis:
Adds an `UnnecessaryFinalInTryWithResources` recipe that removes the redundant `final` modifier from resources declared in a try-with-resources statement. Such resources are implicitly final ([JLS 14.20.3](https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-14.20.3)), so the modifier adds no meaning.

```java
// before
try (final var input = new ByteArrayInputStream(new byte[0])) {
    int x = input.read();
}

// after
try (var input = new ByteArrayInputStream(new byte[0])) {
    int x = input.read();
}
```

### Relationship to `NoFinalizedLocalVariables`

Per @greg-at-moderne's feedback, the two overlap but are not the same thing:

- `NoFinalizedLocalVariables` is a stylistic preference — it strips `final` from *every* local variable (and optionally parameters), where the modifier is meaningful. It is opt-in and belongs to no composite recipe.
- This recipe only touches the one position where `final` is provably redundant by the language spec, so it changes nothing about what the code means.

Also per that feedback, the recipe was renamed from the original `CleanupTryWithResources` — "cleanup" didn't say what it did — to `UnnecessaryFinalInTryWithResources`, matching the neighbouring `UnnecessaryCloseInTryWithResources`.

### Implementation

- Java-only: the visitor is gated behind `Preconditions.check(new JavaFileChecker<>(), ...)`.
- Only `J.Try.Resource` declarations are visited, so `final` on ordinary local variables is untouched.
- A `final` modifier carrying its own annotations is left alone, since dropping it would lose them.
- Formatting is preserved deliberately, since `final` owns whitespace and can carry comments:
  - comments in the removed modifier's prefix are concatenated onto the following token (type expression, or the variable name when the type is implicit) in source order;
  - a declaration that was wrapped across lines around `final` is collapsed back onto one line, but only the gap before each variable name — the initializer keeps its own layout, so a multi-line `new ...(...)` is not reflowed;
  - when `final` leads the declaration its whitespace is dropped entirely; behind a leading annotation the gap belongs to the annotation and survives.
- `estimatedEffortPerOccurrence` is 1 minute. No RSPEC tag — this is not backed by a SonarQube rule.
- Registered in `META-INF/rewrite/recipes.csv`; intentionally **not** added to `common-static-analysis.yml`, per the repo convention that the file holds proven recipes only.

### Tests

`UnnecessaryFinalInTryWithResourcesTest`, 12 cases:

| Case | Covers |
| --- | --- |
| `removeFinalModifier` (`@DocumentExample`) | single resource with `var` |
| `removeFinalWithExplicitType` | explicit type expression instead of `var` |
| `cleanupMultiline` | declaration wrapped across lines is collapsed |
| `removeFinalFromEachOfMultipleResources` | several resources, all `final` |
| `removeFinalOnlyFromTheResourceThatHasIt` | mixed list, only one `final` |
| `retainLeadingAnnotationWhenRemovingFinal` | leading annotation keeps its prefix |
| `retainCommentBetweenFinalAndType` | comment between `final` and the type |
| `retainCommentBetweenAnnotationAndFinal` | comment between an annotation and `final` |
| `retainLineCommentAndItsLineBreak` | line comment keeps its trailing newline |
| `doNotReflowInitializerAcrossLines` | multi-line initializer left as written |
| `doNotChangeResourceWithoutFinal` | no-op |
| `doNotChangeResourceReferencingExistingVariable` | no-op on `try (existingVar)` |


sullis:
Ready for review / feedback 
@greg-at-moderne

sullis:
Ready for review / feedback: CleanupTryWithResources recipe

greg-at-moderne:
BTW, we already have `NoFinalizedLocalVariables` which does something similar. Or the same thing, but with broader scope. I am not sure yet, how to proceed. Probably just have the new recipe as you suggest.

The name is too generic though. "Cleanup" doesn't necessarily tell the user what it does.

sullis:
> BTW, we already have `NoFinalizedLocalVariables` which does something similar. Or the same thing, but with broader scope. I am not sure yet, how to proceed. Probably just have the new recipe as you suggest.
> 
> The name is too generic though. "Cleanup" doesn't necessarily tell the user what it does.

I pushed an update.  Recipe has been renamed to UnnecessaryFinalInTryWithResources

greg-at-moderne (approved):

sullis:
Rebased

src/test/java/org/openrewrite/staticanalysis/UnnecessaryFinalInTryWithResourcesTest.java
```diff
@@ -0,0 +1,410 @@
+/*
+ * Copyright 2024 the original author or authors.
```
greg-at-moderne:
```suggestion
 * Copyright 2026 the original author or authors.
```

src/main/java/org/openrewrite/staticanalysis/UnnecessaryFinalInTryWithResources.java
```diff
@@ -0,0 +1,131 @@
...
+        comments.set(last, comments.get(last).withSuffix(
+                tighten(comments.get(last).getSuffix() + second.getWhitespace())));
+        comments.addAll(second.getComments());
+        return Space.build(first.getWhitespace(), comments);
+    }
+
+    /**
+     * Drops the whitespace that `final` used to occupy while leaving any comments in place.
+     */
+    private static Space reflow(Space space, boolean leadsDeclaration) {
```
greg-at-moderne, line 115:
```suggestion

    private static Space reflow(Space space, boolean leadsDeclaration) {
```

greg-at-moderne (approved):

PR #1056, opened 2026-09-10
Fix serialVersionUID for generic Serializable classes

### LINKED ISSUE #8848: AddSerialVersionUidToSerializable skips generic classes implementing Serializable
**What version of OpenRewrite are you using?**

- rewrite-maven-plugin 6.46.1
- rewrite-static-analysis 2.41.1 (latest release available on Maven Central at the time of writing)
- rewrite-core 8.90.2

**How are you running OpenRewrite?**

Maven plugin, single-module run:

```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.46.1</version>
    <configuration>
        <activeRecipes>
            <recipe>org.openrewrite.staticanalysis.AddSerialVersionUidToSerializable</recipe>
        </activeRecipes>
    </configuration>
</plugin>
```

**What is the smallest, simplest way to reproduce the problem?**

```java
import java.io.Serializable;

public class Foo<T> implements Serializable {
}
```

Running `AddSerialVersionUidToSerializable` over this file produces no change. A non-generic class with the same shape (`public class Bar implements Serializable {}`) gets `serialVersionUID` added correctly.

This is how we noticed it: one run on our codebase added `serialVersionUID` to 99 non-generic classes and silently skipped 6 generic ones (e.g. `public class CommonResult<T> implements Serializable`). Generic classes that are serializable only through a generic superclass are skipped as well (e.g. `class WrapperX<T> extends Wrapper<T>` where the parent hierarchy implements `Serializable`).

**What did you expect to see?**

```java
import java.io.Serializable;

public class Foo<T> implements Serializable {
    private static final long serialVersionUID = 1L;
}
```

Per the recipe's display name, "Add serialVersionUID to a Serializable class when missing", `Foo<T>` directly implements `Serializable` and should get the field.

**What did you see instead?**

No change — the class is silently skipped; `rewrite.patch` contains no entry for it.

**Additional context — probable cause**

In `AddSerialVersionUidToSerializable#requiresSerialVersionField(JavaType)` (2.41.1):

```java
if (type instanceof JavaType.Parameterized) {
    if (parameterized.isAssignableTo("java.util.Collection")
            || parameterized.isAssignableTo("java.util.Map")) {
        // checks whether the type parameters are serializable
    }
    //All other parameterized types fall through
} else if (type instanceof JavaType.FullyQualified) {
    // Interface / Throwable / isAssignableTo("java.io.Serializable") checks happen here
}
return false;
```

For a generic class declaration, `J.ClassDeclaration#getType()` is a `JavaType.Parameterized` (type parameters are `GenericTypeVariable`s). A class declaration is almost never a `Collection`/`Map`, so the `else if` branch — the only place where `isAssignableTo("java.io.Serializable")` is evaluated — is unreachable for generic classes, and the method returns `false`.

Also worth noting: the recipe is tagged `RSPEC-S2057`, but Sonar's implementation of that rule (`SerialVersionUidCheck` in sonar-java, `type.isSubtypeOf("java.io.Serializable")`) reports generic classes such as `class Foo<T> implements Serializable` — there is no generic special case in the rule's semantics, so this skip looks like an implementation gap rather than intended fidelity to S2057.

A possible fix direction: when the parameterized type is not a `Collection`/`Map`, fall through to the same `FullyQualified` checks applied to the underlying raw type instead of returning `false`.

**Are you interested in contributing a fix?**

Possibly — happy to prepare a PR with a failing test first if the direction above sounds right.

LouisDeconinck:
## What changed and why?

Fixes OpenRewrite/rewrite#8848. `AddSerialVersionUidToSerializable` skipped generic class declarations because their `JavaType.Parameterized` type never reached the Serializable check used for fully qualified types. Non-collection/map parameterized types now fall through to that check.

Added a focused regression test covering a generic Serializable class and a generic non-Serializable class.

## Testing

- `./gradlew test --tests org.openrewrite.staticanalysis.AddSerialVersionUidToSerializableTest` *(blocked locally: the current build plugin resolves unreleased OpenRewrite 8.91.6 artifacts from the authenticated CodeGnome repository; those credentials are not available in this environment)*
- `git diff --check`

## Checklist

- [x] Added a regression test for the reported issue
- [x] Covered a similar case that should remain unchanged
- [x] Kept the implementation and test diff focused


greg-at-moderne:
Thanks for getting involved.
The description text is broken somehow, @LouisDeconinck. Could you please have a look at it and try to clean it up? Otherwise it's hard to reason about your suggestion.

LouisDeconinck:
Cleaned up — the description should render properly now. Thanks for flagging it.

PR #1062, opened 2026-09-16
ForLoopIncrementInUpdate: do not move increment skipped by continue

### LINKED ISSUE #1061: ForLoopIncrementInUpdate changes behavior when the loop body contains continue

I am using

- Gradle plugin `org.openrewrite.rewrite` 7.39.0
- `rewrite-recipe-bom` 3.37.0
- `rewrite-static-analysis` 2.41.0 and 2.41.1 (both affected; the recipe source on `main` is unchanged)

Gradle plugin, single-module project, only this recipe active:

```kotlin
plugins {
    java
    id("org.openrewrite.rewrite") version "7.39.0"
}
repositories { mavenCentral() }
dependencies {
    rewrite(platform("org.openrewrite.recipe:rewrite-recipe-bom:3.37.0"))
    rewrite("org.openrewrite.recipe:rewrite-static-analysis:2.41.1")
}
rewrite { activeRecipe("org.openrewrite.staticanalysis.ForLoopIncrementInUpdate") }
```

We found this in JabRef, where the recipe runs in CI and the change broke a unit test.

```java
class A {
    int count(String s) {
        int n = 0;
        for (int i = 0; i < s.length(); i++) {
            if (s.charAt(i) != '#') {
                continue;
            }
            n++;
            i++;
        }
        return n;
    }
}
```

No change. The trailing `i++` runs only on iterations that do not hit `continue`. Moving it into the update clause makes it run on every iteration, which changes behavior.

```java
class A {
    int count(String s) {
        int n = 0;
        for (int i = 0; i < s.length(); i++) {
            if (s.charAt(i) != '#') {
                continue;
            }
            n++;
            i++;
        }
        return n;
    }
}
```

## What did you see instead?

```java
class A {
    int count(String s) {
        int n = 0;
        for (int i = 0; i < s.length(); i++, i++) {
            if (s.charAt(i) != '#') {
                continue;
            }
            n++;
        }
        return n;
    }
}
```

The loop now skips every other character: `count("a#")` returns 1 before and 0 after.

Suggested fix: `ForLoopIncrementInUpdate` should skip the loop when its body contains a `continue` that targets this loop (an unlabeled `continue` not nested in an inner loop, or a labeled `continue` with this loop's label). `break`/`return` are fine, since the loop variable is out of scope afterwards.

None.

Yes, I will open a PR with a fix proposal.

koppor:
## What's changed?

`ForLoopIncrementInUpdate` no longer moves the trailing increment into the update clause when the loop body contains a `continue` that targets this loop (unlabeled and not inside a nested loop, or labeled with this loop's label).

## What's your motivation?

- Fixes #1061. A `continue` skips the trailing increment, so moving it into the update clause makes it run on every iteration and changes behavior.

`break` and `return` are intentionally not handled: the loop variable is out of scope after the loop exits.

## Any additional context

I could not run the project's build locally (plugin artifacts come from the authenticated codegenome repository), so I ran `ForLoopIncrementInUpdateTest` against `rewrite-recipe-bom` 3.37.0: all four tests pass with the change, and the two new "do not move" tests fail without it.


greg-at-moderne (approved):
Thanks for the contribution!

PR #1064, opened 2026-09-17
UnwrapRepeatableAnnotations: only unwrap container annotations

Niloyyy:
- Fixes #389

Thanks @greg-at-moderne for the isolated reproduction; this PR adds it as a test, with the annotation members corrected so the sources compile.

## The problem

`unwrap` searched an annotation's entire subtree for any repeatable annotation, and replaced the annotation with whatever it found. Any annotation that merely takes a repeatable annotation as one of its elements was therefore replaced by that element, and its remaining elements were discarded:

```diff
- @CollectionTable(name = "OBJECT_ENV", joinColumns = @JoinColumn(name = "id"))
+ @JoinColumn(name = "id")
```

`@CollectionTable` and its `name` are gone, which is what produces the Hibernate failure in the issue.

## The change

Unwrap only when the annotation is actually a container: it has a single argument, given either positionally or as `value`, holding a repeatable annotation or an array of them. Anything else is left alone.

The issue suggests a more precise alternative, keying on the container named by the inner annotation's `@Repeatable`. That is not reachable from the type model: `JavaType.FullyQualified.getAnnotations()` returns only the annotation types, so `@Repeatable`'s value is not available to compare against. The structural check is what can be implemented today, and it rejects the reported case for the right reason, since a container annotation by definition has nothing else in it.

## Tests

Two added:

- the reported case, tagged to #389, which fails without this change
- `@Annotations(value = {@Annotation, @Annotation})`, covering the explicit `value` form that the new check has to keep unwrapping

The existing `unwrapRepeatable` test is untouched and still passes.

Full suite locally on JDK 21: 2338 tests, the only failures being 7 pre-existing `Failed to start RPC process: rewrite-go-rpc` errors that occur identically on unmodified `main` in this environment.

Unrelated note: `./gradlew` currently fails to configure for me because the `latest.release` build plugins resolve to `2.23.4`, which needs `org.openrewrite:rewrite-java:8.91.4`, and that version is not on the buildscript repositories. I pinned the two plugins to `2.23.1` locally to run the tests; that pin is not part of this PR.

PR #1068, opened 2026-09-18
ForLoopIncrementInUpdate: don't move the increment when a continue can skip it

### LINKED ISSUE #1061: ForLoopIncrementInUpdate changes behavior when the loop body contains continue

Niloyyy:
- Fixes #1061

## The problem

The recipe moves a trailing `i++` from the loop body into the update clause without checking whether anything in the body can jump past it. A `continue` skips the rest of the body but not the update clause, so once the increment is moved it runs on iterations that previously skipped it:

```java
for (int i = 0; i < s.length(); i++) {
    if (s.charAt(i) != '#') {
        continue;
    }
    n++;
    i++;
}
```

became

```java
for (int i = 0; i < s.length(); i++, i++) {
    if (s.charAt(i) != '#') {
        continue;
    }
    n++;
}
```

Running both over `"a#b#"` gives `2` before and `0` after. (Some inputs happen to agree, which is presumably why this went unnoticed until it broke a JabRef test.)

## The change

Before moving the increment, check whether the body contains a `continue` that targets this loop. That is either an unlabeled `continue` with no other loop between it and this one, or a labeled `continue` naming this loop's label. If so, leave the loop alone.

The check deliberately ignores a `continue` that belongs to a nested loop, since that one never bypasses the outer increment, and there is a test asserting the outer increment is still moved in that case. `break`, `return` and `throw` need no handling: they leave the loop entirely, so neither the trailing increment nor the update clause runs.

## Tests

- Three added, tagged to #1061: the reported unlabeled case, a labeled `continue outer` from inside a nested loop, and a positive case where the `continue` belongs to the nested loop and the outer increment is still moved. The first two fail without this change; the existing test is untouched and still passes.

Full suite locally on JDK 21: 2413 tests, the only failures being 8 pre-existing `Failed to start RPC process: rewrite-go-rpc` errors from Go tests that cannot start the RPC binary in this environment, unrelated to this change.

- Note: as with #1064, `./gradlew` currently fails to configure for me because the `latest.release` build plugins resolve to `2.24.2`, needing `org.openrewrite:rewrite-java:8.92.5`, which is not on the buildscript repositories. I pinned the plugins to `2.23.1` locally to run the tests; the pin is not part of this PR.


greg-at-moderne:
Thanks, @Niloyyy, but I think the very same issue with a similar PR has just been submitted in #1062 and already merged.
Thanks for getting involved anyway!

Niloyyy:
Yup I have seen this after creating my PR. Anyway willing to contribute further.

